PR #176, opened 2022-01-10
Configure build with GitHub Actions

ccjernigan:
This PR configure CI and deployment.

Firebase Test Lab can be used for automated testing, which is set up with a Gradle plugin.  This allows both CI and local development the ability to use the same testing infrastructure if desired.

Secrets need to be defined for testing and deployment to occur.  If those secrets are missing, those build steps will be automatically skipped.  This makes it possible for forks to at least partially benefit from the CI setup (compile and static analysis) even if they don't configure the secrets needed for Firebase Test Lab.

Some things that might not be totally obvious without prior Android experience:
 - The signing key for Google Play will be the upload key.  If that is ever leaked, we can rotate it.  The real signing key will be escrowed with Google (and it is possible to retain our own copy separate from the deployment process).
 - Google Play allows for different access control for different users, so we can ensure that a deploy to production from CI is not possible from the service user account


pacu:
Assigning a reviewer that actually knows about CI 😅

@mdr0id feel free to assign whomever else from DevSecOps :)

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,44 @@
+name: 'Setup Java and Dependency Cache'
+description: "Configures the build environment and caches Gradle and dependencies."
+runs:
+  using: "composite"
+  steps:
+    - name: Set Env
+      shell: bash
+      run: |
+        echo "home=${HOME}" >> $GITHUB_ENV
```
defuse, line 9:
```suggestion
        echo "home=${HOME}" >> "$GITHUB_ENV"
```

(In case `$GITHUB_ENV` contains a space)

docs/CI.md
```diff
@@ -0,0 +1,44 @@
...
+* `SIGNING_KEY_ALIAS_PASSWORD` — Password for key alias.
+
+To obtain the values for the Google Play deployment, you'll need to
+
+* Create a service account with access to your Google Play account.  Recommended permissions are to "edit and delete draft apps" and "release apps to testing tracks".
+* Configure [workload identity federation](https://github.com/google-github-actions/auth#setting-up-workload-identity-federation)
+
+Note that security of release deployments is enhanced via two mechanisms:
+ - CI signs the app with the upload keystore and not the final release keystore.  If the upload keystore is ever leaked, it can be rotated without impacting end user security.
+ - Deployment to Google Play can only be made to testing tracks.  Release to production requires manual human login under a different account with greater permissions.
```
defuse, line 44: Excellent documentation!

defuse (commented):
Partial review comments. I still need to:

- [ ] Review `deploy.yml`
- [ ] Understand how the OpenID Connect thing works for Firebase (and separately, audit that configuration)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+# SIGNING_KEY_ALIAS - The key alias inside SIGNING_KEYSTORE_BASE_64
+# SIGNING_KEY_ALIAS_PASSWORD - The password for the key alias
+
+name: Deploy
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
defuse, line 16: (Reminder to myself that we need to configure branch protection rules prior to merging this!)

defuse: Specifically, we need at least one review approval from someone other than the PR author, otherwise a dev could steal the secrets by approving their own malicious PR.

defuse: I set up the branch protection rule

defuse (approved):
utACK

We need to do the firebase setup carefully. Untrusted PRs can effectively get a token which allows them to access that API and run tests which cost money after a certain threshold. Someone could make PRs to run up our bill, so we need to make sure we're monitoring the billing and that the service account doesn't have any access beyond the bare minimum. We can probably set up a whitelist so that this only runs for PRs created by us using the `--attribute-condition` argument to `gcloud iam workload-identity-pools providers create-oidc`.

Issue #218, opened 2022-02-16
Prevent Illegal Input in ZEC Send/Request Form


ccjernigan:
### Describe the issue
The ZEC request form prevents non-numeric input but it does not enforce correctly formatted amounts

#### If so, please list the steps to reproduce below:
1. Visit the ZEC request screen
2. Type `1.2.3.4`

Note: this also reproduces on the ZEC send screen

### Expected behaviour
This should be prevented, such that the user can only type 1.234.  A more complex issue is the thousands separator, which also needs to be correctly enforced.

### Actual behaviour + errors
Users are allowed to type 1.2.3.4 which will have unexpected behavior when trying to parse as ZEC into Zatoshi

HonzaR:
Hi @ccjernigan,

I've checked possible solutions here:

- Use method `toDoubleOrNull()`, which helps us with checking the correct use of `.`, but unfortunately does not help us with USA thousand `,` separator. 
- Then the proper way to check number would be using of regex. I've already created one for our purpose `"^[0-9]+([,][0-9]+)?([.][0-9]+)?$"`, which works pretty well, but is built only for USA number formatting patterns. There is also a problem with other locale like Arabic, Chinese, etc.
- Use some of existing solid solution, e.g. `Commons.NumberUtils` (https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/math/NumberUtils.html#isNumber-java.lang.String-)
- There are more solutions available, which don't seem to work with our use case (e.g. https://www.techiedelight.com/determine-string-is-number-kotlin/)

Ideally, we'd like to use some Kotlin build-in validation feature to bypass problems with different number formats in different countries as well as for different alphabets. Do you know some? What would be the way we should go? Thank you.

Here is a quite good source of information, which we need to keep in mind: https://docs.microsoft.com/en-us/globalization/locale/number-formatting

ccjernigan:
In general, we avoid floating point values because of imprecision.  (If absolutely necessary, we might use BigDecimal in a few places).

Since we’re defining our own currency, we have some additional control over the formatting rules.  At the moment, I think we’re mostly just trying to be consistent with the user’s locale for text direction and thousands/decimal separators.

We know what the separators are, extracting them out from the currency localization.  We likewise can get the text direction.

I think a regex solution is probably the most viable.  We might turn the regex into a string template to swap out the position of the comma and period.  That would cover most languages.  Then I suppose we need a different template for RTL languages.

[...]

Issue #248, opened 2022-02-26
RTL Localization Issues


ccjernigan:
## Is your feature request related to a problem? Please describe.
The application does not currently support any RTL languages.

## Additional context
Once we have our first RTL language, some things will need to be updated. This issue provides the ability to add TODO comments in the codebase which makes it easier to track RTL issues that we need to follow up on.

Known issues include:
 - String formatters for abbreviated addresses.  See `WalletExtTest`

PR #295, opened 2022-03-25
[#218] Prevent illegal number on ZEC amount

HonzaR:
The aim while solving the issue https://github.com/zcash/secant-android-wallet/issues/218 was to not let users type invalid numbers into the ZEC amount textfield, which concerns screens `Request ZEC` and `Send ZEC`.  As an appropriate solution here was chosen solution with regular expression:

`^^$|[0-9]+[{separators.grouping}]$|[{separators.grouping}][0-9]+)?([{separators.decimal}]$|[{separators.decimal}][0-9]+)?$`

This regex is taken from `strings.xml` for future support of any other localizations. Also the regex contains monetary separators (grouping and decimal), as these separators may differ in countries:

_In the United States, this character is a comma (,). In Germany, it is a period (.). Thus one thousand and twenty-five is displayed as 1,025 in the United States and 1.025 in Germany. In Sweden, the thousands separator is a space, etc._

The solution to this issue consists of these steps:

1. New `sdk-ext-ui` module for UI related SDK helper components.
2. Regex for continuous validation of the entered ZEC amount.
3. Added a new unit tests for validation of the regex.
4. Using the regex on ZEC Request and ZEC Send screens.
5. Updated existing and created a new UI tests for validating entered ZEC amount values on Request and Send screens and its overall cooperation with other screen UI elements.
6. Architecture documentation update with the newly added `sdk-ext-ui` module.

Implementing the issue there where found next unwanted behaviours or bugs - https://github.com/zcash/secant-android-wallet/issues/288, https://github.com/zcash/secant-android-wallet/issues/289, https://github.com/zcash/secant-android-wallet/issues/292, https://github.com/zcash/secant-android-wallet/issues/294.

Closing https://github.com/zcash/secant-android-wallet/issues/218.

# Author

- [ ] Code coverage: Did you check the code coverage report for the automated tests?  _While we are not looking for perfect coverage, the tool can point out potential cases that have been missed._ Code Coverage can be run with: `./gradlew connectedCheck -PIS_COVERAGE_ENABLED=true`
- [x] Documentation: Did you update documentation as appropriate? (e.g [README.md](../blob/main/README.md), etc.)
- [x] Run the app: Did you run the app and try the changes?
- [x] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? _Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit._

# Reviewer

- [x] How is Code Coverage affected by this PR? _We encourage you to compare coverage before and after changes and when possible, leave it in a better place._


ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/home/model/WalletSnapshot.kt
```diff
@@ -6,6 +6,7 @@ import cash.z.ecc.android.sdk.ext.ZcashSdk
 import cash.z.ecc.android.sdk.type.WalletBalance
 import cash.z.ecc.sdk.model.Zatoshi
 
+// TODO [#292]: Should be moved to SDK-EXT-UI module.
```
ccjernigan, line 9: Awesome, thanks for adding this context in the code.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/request/view/RequestView.kt
```diff
@@ -90,22 +92,27 @@ private fun RequestTopAppBar(onBack: () -> Unit) {
 
 // TODO [#215]: Need to add some UI to explain to the user if a request is invalid
 // TODO [#217]: Need to handle changing of Locale after user input, but before submitting the button.
+// TODO [#288]: TextField component can't do long-press backspace.
```
ccjernigan, line 95: Likewise, great to see this added here.

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        regexAmountChecker.also {
+            // valid inputs
+            assertTrue(it.matches(""))
+            assertTrue(it.matches("123"))
+            assertTrue(it.matches("123${separators.grouping}"))
+            assertTrue(it.matches("123${separators.grouping}456"))
+            assertTrue(it.matches("123${separators.decimal}"))
+            assertTrue(it.matches("123${separators.decimal}456"))
+            assertTrue(it.matches("123${separators.grouping}456${separators.decimal}"))
+            assertTrue(it.matches("123${separators.grouping}456${separators.decimal}789"))
```
ccjernigan, line 45: I'd like to see an additional case with 2 groupings.  E.g. 1,234,567.00

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        regexAmountChecker.also {
+            assertNotNull(regexAmountChecker)
+            assertTrue(regexAmountChecker.pattern.isNotEmpty())
+        }
+
+        regexAmountChecker.also {
+            // valid inputs
+            assertTrue(it.matches(""))
+            assertTrue(it.matches("123"))
+            assertTrue(it.matches("123${separators.grouping}"))
```
ccjernigan, line 40: This is good for the regex which is responsible for validation.

However this raises a question about how the ultimate conversion in `Zatoshi.Companion.fromZecString` will work.  If the user types in `123,` and hits send is that converted to `123` or `123,000`?   @geffenz do you have thoughts on what we do here?

sdk-ext-ui/src/main/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegex.kt
```diff
@@ -0,0 +1,32 @@
...
+     *
+     * Valid amounts: "" 123 123, 123. 123,456 123.456 123,456.789 etc.
+     * Invalid amounts: 123,, 123,. 123.. .123 ,123 123,456,789 123.456.789 etc.
+     *
+     * @param context used for loading localized pattern from strings.xml
+     * @param separators which consist of localized monetary separators
+     *
+     * @return regex with pattern for entered ZEC amount
+     */
+    fun getZecAmountContinuousFilter(context: Context, separators: MonetarySeparators): Regex {
```
ccjernigan, line 23: I like the name "continuous filter" as that does a great job describing what this method does.

Thinking about this as a "public API" (since it is in sdk-ext-ui and we hope to make it reusable), we could probably simplify the return type.  I don't think clients need to know this is implemented as a regex, and a simple boolean return type is probably good.  Following Kotlin conventions for collections.filter, a true return value means keep the item (e.g. it is good)

So what do you think of a signature like this?

`fun ZecString.filterContinuous(context: Context, separators: MonetarySeparators, zecString: String)`

ccjernigan (commented):
This is a great enhancement.  In addition to fixing the illegal amount validation, I appreciate the other continuous improvement enhancements here.

Some comments inline, curious to hear your feedback.

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import cash.z.ecc.sdk.model.MonetarySeparators
+import org.junit.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+
+class ZecRegexTest {
+
+    @Test
+    @MediumTest
```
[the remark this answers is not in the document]
HonzaR: Thank you for clarification, I wanted to ask for these annotations using. I'll change it.

sdk-ext-ui/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources xmlns:tools="http://schemas.android.com/tools" xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
+    <string name="zec_amount_regex_filter" formatted="true" tools:ignore="TypographyDashes">^^$|[0-9]+([<xliff:g id="group" example="12,345">%1$s</xliff:g>]$|[<xliff:g id="group" example="12,345">%2$s</xliff:g>][0-9]+)?([<xliff:g id="dec" example="1.2345">%3$s</xliff:g>]$|[<xliff:g id="dec" example="1.2345">%4$s</xliff:g>][0-9]+)?$</string>
```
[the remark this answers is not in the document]
HonzaR: Yes, good idea. I'll improve it like this.

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        assertNotNull(regexString)
+
+        val regexAmountChecker = regexString.toRegex()
+
+        regexAmountChecker.also {
+            assertNotNull(regexAmountChecker)
+            assertTrue(regexAmountChecker.pattern.isNotEmpty())
+        }
+
+        regexAmountChecker.also {
```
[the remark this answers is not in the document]
HonzaR: I like this idea. I will divide it as proposed.

ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/screen/request/view/RequestViewTest.kt
```diff
@@ -81,49 +73,91 @@ class RequestViewTest {
...
+        // e.g. 123,456.789
+        composeTestRule.setAmount("123${separators.grouping}456${separators.decimal}789")
+        composeTestRule.clickCreateAndSend()
+        assertEquals(5, testSetup.getOnCreateCount())
+    }
 
-        composeTestRule.onNodeWithText(getStringResource(R.string.request_create)).also {
-            it.performClick()
-        }
+    @Test
```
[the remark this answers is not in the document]
HonzaR: Ok, understand. I check it and we'll see.

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
HonzaR: That is something which needs additional regex enhancement, thank you for pointing it out.

sdk-ext-ui/src/main/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegex.kt
HonzaR: Yes, this can work. To be clear here, you propose to keep it in the `ZecRegex.kt` and make an extension function of 
`ZecString.kt` from it?

ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/screen/request/view/RequestViewTest.kt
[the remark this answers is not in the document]
HonzaR: @ccjernigan Now I've enhanced the regex the way it allows us to input decimals `.123` and then the subsequent formatting with `Zatoshi.Companion.fromZecString` results like this `0.12300000`. So that seems correct to me.

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
HonzaR: @ccjernigan @geffenz this particular case of `123,` ends up with `123.00000000` after processed by `Zatoshi.Companion.fromZecString`.

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
```diff
@@ -0,0 +1,86 @@
...
+            arrayOf(
+                separators.grouping,
+                separators.decimal
+            )
+        ).toRegex()
+
+        regex.also {
+            assertTrue(it.matches(""))
+            assertTrue(it.matches("123"))
+            assertTrue(it.matches("${separators.decimal}"))
```
ccjernigan, line 50: I think we are going to have trouble with some, but not all, of these test cases in the future.  The behavior of the test case depends on the locale of the device it is running on (and the availability of localized resources for RTL).  To make this easier on our future selves, a simple solution might be:

1. Add annotation to require API 17 for this test case.  I think it is `@SdkSuppress(minSdkVersion = Build.VERSION_CODES.N)`  We haven't settled on min API yet, but 17 is reasonable and required for the next suggestion.  (We're currently targeting 16, hence the need for the annotation)

2. Use an if statement checking for [LTR](https://developer.android.com/reference/android/text/TextUtils.html#getLayoutDirectionFromLocale(java.util.Locale) ) for assertions that would matter on direction.  (That means primarily the ones that have both separators and numbers)

3. Don't worry about adding RTL test cases in the else block.  Instead add a TODO pointing to https://github.com/zcash/secant-android-wallet/issues/248

This comment would apply to both here and in the Compose UI test.

ccjernigan (commented):
This is looking great!  I had one small localization concern in our tests.

We could file another issue for followup, but I think the fix is simple enough to implement before merging.  What do you think?

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
HonzaR: Thank you @ccjernigan for the proposed solutions. I think that the solution **2 - tests for both LTR and RTL** is the way we should go. I've just tested how the OS treats our test with **Force RTL** developer option as well as with switch device language to **Arabic** on both real and emulator devices, and I believe that we don't nee to make any changes in our regex patter in strings.xml nor in implementation of our tests. Even if the method `TextUtils.getLayoutDirectionFromLocale(locale)` returns us RTL, the regex and tests for LTR succeed. It seems that RTL vs LTR only affects UI displaying and our tests are still valid.

What do you think about it?

HonzaR: I found another problem in Arabic locale and that it related to `ZecString.fromZecString` method. 
`IllegalArgumentException: Unquoted special character '#' in pattern "#٬##0٫0#"`
So I've added TODO pointing to https://github.com/zcash/secant-android-wallet/issues/248. You probably also noticed as you have it marked in the pointing issue.

ccjernigan: I want to make sure we're on the same page:

The test cases won't fail right now because we're missing the RTL localization in strings.xml.  But once it is added, the all the test cases are broken.  E.g. Testing against a hard-coded string of "1,234.00" breaks once the RTL strings.xml is added since the order should be reversed.  I am not suggesting we add a RTL localization or fix those test cases now.

I'm only suggesting adding the if statement in the tests so that when we come back to this both A. the tests won't fail immediately and B. the hook for adding additional test cases is present.  I am suggesting this because code has a habit of being copy-pasted, and I don't want hard-coded localization issues in our tests to propagate to other screens.  This makes others who review the code aware that we know RTL is not fully supported here.

ccjernigan: That's a great catch in fromZecString.

Does it crash when the system language is set to Arabic and then we try to send ZEC?  That sounds like a bug, because someone with an RTL language should basically see the app work as it does in English until we add the right RTL resources.  If that is true, then it might need to be a separate bug report because #248 is about adding a new feature (supporting RTL in the UI).

ccjernigan (approved):
Looks good to squash and merge, pending the CI build completing.

sdk-ext-ui/src/androidTest/java/cash/z/ecc/sdk/ext/ui/regex/ZecRegexTest.kt
HonzaR: Discussed more in details out of this thread. For now we suppose that the tests won't be affected once we add RTL language to the app in the future.

About the crash on the `ZEC Request` screen after typing a single character to amount field - filed a new issue https://github.com/zcash/secant-android-wallet/issues/321 and linked form implementation too.

Issue #71, opened 2021-11-06
Add re-creation test for onboarding


ccjernigan:
## Is your feature request related to a problem? Please describe.
The automated test coverage for the onboarding flow does not currently verify state restoration after a configuration change.

## Describe the solution you'd like
Add an automated test that moves the screen forward from the first step, triggers a configuration change, and verifies that the state is retained (e.g. the position in the onboarding flow is preserved).

PR #361, opened 2022-04-10
[#71] Add re-creation test for onboarding

HonzaR:
- Added test function `stage_restoration()` to test correct state of onboarding after activity recreation.
- Fix typo.

Closes https://github.com/zcash/secant-android-wallet/issues/71.

# Author

- [ ] Code coverage: Did you check the code coverage report for the automated tests?  _While we are not looking for perfect coverage, the tool can point out potential cases that have been missed._ Code Coverage can be run with: `./gradlew connectedCheck -PIS_COVERAGE_ENABLED=true`
- [x] Run the app: Did you run the app and try the changes?
- [x] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? _Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit._

# Reviewer

- [ ] Checklist review: Did you go through the code with the [Code Review Guidelines](../blob/main/docs/CODE_REVIEW_GUIDELINES.md) checklist?
- [ ] Ad hoc review: Did you perform an ad hoc review?  _In addition to a first pass using the code review guidelines, do a second pass using your best judgement and experience which may identify additional questions or comments. Research shows that code review is most effective when done in multiple passes, where reviewers look for different things through each pass._
- [ ] Automated tests: Did you review the automated tests?
- [ ] How is Code Coverage affected by this PR? _We encourage you to compare coverage before and after changes and when possible, leave it in a better place._


ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/screen/onboarding/view/OnboardingViewTest.kt
```diff
@@ -239,10 +242,37 @@ class OnboardingViewTest {
...
+    fun stage_restoration() {
+        val testSetup = newTestSetup(OnboardingStage.ShieldedByDefault)
+
+        assertEquals(OnboardingStage.ShieldedByDefault, testSetup.getOnboardingStage())
+
+        composeTestRule.onNodeWithText(getStringResource(R.string.onboarding_next)).also {
+            it.performClick()
+        }
+
+        ActivityScenario.launch(MainActivity::class.java).also { scenario ->
```
ccjernigan, line 256: I'm pretty sure this ends up doing something weird.

`ComposeTestRule` (and the compose test framework) bundle a separate ComposeTestActivity (I don't know the exact name but we can find it in AOSP or by looking at the merged manifest) into the test APK.  So all of our tests are run in that special activity, which just sets the view to the Composable we configure.

`ActivityScenario.launch(MainActivity::class.java)` ends up launching a second Activity on top.

The fact that the onboarding stage carries forward to MainActivity after clicking next in ComposeTestRule doesn't immediately make sense to me.

ccjernigan: Separate from the comment above, I have a concern that this ends up being an integration test that depends on the application having no state on disk (if another test saves a mnemonic, then this test case will fail).

ccjernigan (commented):
Thanks for taking the first stab at this!  Although it seems small, figuring this out will be really important to our overall testing patterns for the UI.  So we'll probably have more discussion here compared to some issues.

As a general comment, I think we're hoping to achieve a few big things:
1. A test for this particular screen to ensure state restoration is correct.  I think we basically have that, but there might be a bug as commented inline.
2. A generic approach we can use for all Compose screens with state that should save/restore e.g. that have user input (UI unit tests).
3. A generic approach we can use to test ViewModels
4. A generic approach we can use for Activity lifecycle (UI integration tests).  Hopefully we have very few of these.

I don't exactly know how we'll achieve this yet, although with your help we'll get there.  Some of the testing patterns with Compose are different, so this is new for me too.  I'm guessing we might leverage a combination of these APIs at some point:
 - For Compose views only: https://developer.android.com/reference/kotlin/androidx/compose/ui/test/junit4/StateRestorationTester
- For Activity/ViewModel integration testing
    1. Use UiAutomator `UiDevice.setOrientationRight()`
    1. Use Espresso `activity.setRequestedOrientation(orientation)`

A few trends might happen:
- Encourage refactoring of UI state retention into rememberSaveable fields, e.g. the onboarding stage in order to allow for more compose-only unit testing, therefore reducing the integration testing.
- Encourage unit testing of ViewModels in isolation, instead of integrated into MainActivity with all the views.
- Separate integration testing into different files or even Gradle modules.
    - Separate files: E.g. new file `OnboardingIntegrationTest.kt`.  I think the separation might be necessary, especially dealing with problems between `ComposeTestRule` versus `ActivityTestRule` at the same time.
    - Separate modules: Some integration tests will end up setting persisted state, which needs test orchestrator to clear.  I don't want to use Test Orchestrator in the ui-lib tests, because it makes the tests really slow during development.  We are using Orchestrator in the app module, or we could do a new ui-lib-integration-test module.  I don't think we're at the point of needing this yet, just highlighting where we might end up and why.

With all that background, how would we write a test that reproduces this bug? https://github.com/zcash/secant-android-wallet/issues/246

Answering that question while implementing this will probably help us arrive at a good solution.

HonzaR:
> 

Thank you for pointing it out. Seems that we have a lot of work here. I'll go through it and come up with some updates soon.

ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/screen/onboarding/view/OnboardingViewTest.kt
HonzaR: Yes, agree, I had also another solution previously, which was probably better. I'll change it.

ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/screen/onboarding/view/OnboardingIntegrationTest.kt
```diff
@@ -0,0 +1,70 @@
...
+     * restoration, and then make the assertion on it again.
+     */
+    @Test
+    @MediumTest
+    fun current_stage_restoration() {
+        val restorationTester = StateRestorationTester(composeTestRule)
+        var testSetup: OnboardingTestSetup? = null
+
+        restorationTester.setContent {
+            testSetup = newTestSetup(OnboardingStage.UnifiedAddresses)
```
ccjernigan: As-is, I think this line means that the assertions below can never fail because we re-set the state here.  I'm pushing a commit now that I think will resolve it.

ccjernigan (approved):
Assuming my additional commit makes sense to you, I think this is good to go!

ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/screen/onboarding/view/OnboardingIntegrationTest.kt
HonzaR: Oh yes, my bad. Thank you for the review. Now it's good to go.

PR #446, opened 2022-05-26
Reproduce crash in lint

ccjernigan:
Do not merge.  This is for reporting a reproducible bug to Google in Android lint.  I want the CI job to run which shows the failure occurring.


ccjernigan:
Closing PR; it will still be accessible for Google to implement the lint fix after being closed.

PR #627, opened 2022-10-18
throttling wallet snapshot emits

gasstan:

# Author

- [ ] Code coverage: Did you check the code coverage report for the automated tests?  _While we are not looking for perfect coverage, the tool can point out potential cases that have been missed._ Code coverage can be generated with: `./gradlew check` for Kotlin modules and `./gradlew connectedCheck -PIS_ANDROID_INSTRUMENTATION_TEST_COVERAGE_ENABLED=true` for Android modules.
- [ ] Run the app: Did you run the app and try the changes?
- [ ] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? _Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit._

# Reviewer

- [ ] How is code coverage affected? _We encourage you to compare coverage before and after changes and when possible, leaving it in a better place._

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
@@ -1,0 +1,58 @@
+package co.electriccoin.zcash.ui.common
+
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
+import kotlin.time.Duration
+import kotlin.time.ExperimentalTime
+import kotlin.time.TimeSource
+
+@OptIn(ExperimentalTime::class)
+fun <T> Flow<T>.throttle(
+    duration: Duration,
+    timeSource: TimeSource = TimeSource.Monotonic
+): Flow<T> = flow {
+    coroutineScope {
+        val context = coroutineContext
+        val mutex = Mutex()
+
+        var timeMark = timeSource.markNow()
+        var delayEmit: Deferred<Unit>? = null
+        var firstValue = true
+        var valueToEmit: T
+        collect { value ->
+            if (firstValue) {
+                firstValue = false
+                emit(value)
+                timeMark = timeSource.markNow()
+                return@collect
+            }
+            delayEmit?.cancel()
+            valueToEmit = value
+
+            if (timeMark.elapsedNow() >= duration) {
+                mutex.withLock {
+                    emit(valueToEmit)
+                    timeMark = timeSource.markNow()
+                }
+            } else {
+                delayEmit = async(Dispatchers.Default) {
+                    mutex.withLock {
+                        delay(duration)
+                        withContext(context) {
+                            emit(valueToEmit)
+                        }
+                        timeMark = timeSource.markNow()
+                    }
+                }
+            }
+        }
+    }
+}
--- a/ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/common/FlowExtTest.kt
+++ b/ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/common/FlowExtTest.kt
@@ -1,0 +1,67 @@
+package co.electriccoin.zcash.ui.common
+
+import androidx.test.filters.FlakyTest
+import androidx.test.filters.SmallTest
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+import kotlin.test.assertTrue
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.ExperimentalTime
+import kotlin.time.TimeMark
+import kotlin.time.TimeSource
+
+class FlowExtTest {
+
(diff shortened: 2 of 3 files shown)


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
```diff
@@ -0,0 +1,52 @@
...
+            }
+
+            if (timeMark.elapsedNow() >= duration) {
+                emit(value)
+                lastEmittedValue = value
+                delayEmit?.cancel()
+                timeMark = timeSource.markNow()
+            } else {
+                delayEmit?.cancel()
+                delayEmit = async(Dispatchers.IO) {
```
ccjernigan, line 41: I have a concern that we could potentially end up emitting an old value due to a race condition.

If the timing is right, I think a newly collected value in the outer collect could emit and then the async emit could occur with the prior value.

I think it would be helpful to see if we can write a test case that triggers this condition.  It would be a non-deterministic test case (e.g. doesn't fail every time), but if we run the test in a loop it should occur frequently enough that it becomes a robust verification.  Honza did something similar for the SDK's DatabaseCoordinatorTest.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
```diff
@@ -0,0 +1,52 @@
...
+                emit(value)
+                lastEmittedValue = value
+                timeMark = timeSource.markNow()
+                return@collect
+            }
+
+            if (timeMark.elapsedNow() >= duration) {
+                emit(value)
+                lastEmittedValue = value
+                delayEmit?.cancel()
```
ccjernigan, line 37: Related to the race condition concern below, I think the cancel should be moved before the emit.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
ccjernigan, line 41: Is there a way to do this without Dispatchers.IO?

If we need an async context, would Dispatchers.Default be better?  (Default is for CPU work while IO is for threads that are blocked on disk or network writes).

ccjernigan (commented):
Thanks for the effort that went into this!  There's a lot of good implementation here.

I have a potential concern about whether a race condition exists, and I think we can verify with some testing to see if we can trigger a race or not.  (It might be something along the lines of taking a flow of ordered values and ensuring the output is always in increasing order but with some items dropped)

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
```diff
@@ -0,0 +1,52 @@
...
+
+            if (timeMark.elapsedNow() >= duration) {
+                delayEmit?.cancel()
+                emit(value)
+                Twig.warn { " Normal Emmit $value" }
+                timeMark = timeSource.markNow()
+            } else {
+                delayEmit?.cancel()
+                delayEmit = async(Dispatchers.Default) {
+                    delay(duration /2)
```
gasstan, line 42: @ccjernigan after some testing, i found out that reducing delay to half the duration dramatically reduces the race condition. Without it the test would crash with throttle duration set to 0.01.seconds but with delay = duration / 2 i was able to go as low as 0.0001. It is probably not ideal but in theory "delayEmit" should be invoked only for the last item. What do you think?

ccjernigan: It sounds like you now have a test case that reproduces a race condition, which is good news!

Now that we can reproduce it, our goal isn't to reduce the race condition but to eliminate it entirely.  Code like this should be as deterministic as possible (knowing that we'll still have some timing variations just due to CPU clocks, etc).  So I don't recommend cutting the interval in half.

I'm not 100% certain but a Mutex seems like part of a possible solution.  E.g. The async block acquires the mutex to ensure exclusion, as well as the other branch of the if statement.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
```diff
@@ -0,0 +1,57 @@
...
+                    emit(value)
+                    timeMark = timeSource.markNow()
+                }
+            } else {
+                delayEmit?.cancel()
+                delayEmit = async(Dispatchers.Default) {
+                    mutex.withLock {
+                        delay(duration)
+                        withContext(context) {
+                            emit(value)
```
ccjernigan, line 49: I think we still have a bug to address: this emits potentially stale values.  The value is captured at the time of `collect` above but is then delayed.  We actually want to emit the most recently received value.  I think this implies having a `var` outside of the `collect` holding a reference to the most recently received value.

So the async isn't about delayed emission of a specific value but actually delayed emission of the latest value.

ccjernigan (commented):
We're getting there! A few more comments. Although the issue description was simple, this is a pretty tricky thing to implement so I think we need one more small iteration.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/FlowExt.kt
```diff
@@ -0,0 +1,59 @@
...
+    timeSource: TimeSource = TimeSource.Monotonic
+): Flow<T> = flow {
+    coroutineScope {
+        val context = coroutineContext
+        val mutex = Mutex()
+
+        var timeMark = timeSource.markNow()
+        var delayEmit: Deferred<Unit>? = null
+        var firstValue = true
+        var valueToEmit: T
```
gasstan, line 29: @ccjernigan something like this?

ccjernigan: Yes. But I think you still have a race condition.  valueToEmit gets assigned and then you call `delayEmit?.cancel()`.  Cancellation needs to occur first, otherwise it can emit the value twice in a row.

ccjernigan (changes requested):

ccjernigan (approved):

PR #641, opened 2022-10-20
implementation of push and hold to send zec

gasstan:

# Author

- [ ] Self-review: Did you review your own code in GitHub's web interface? _Code often looks different when reviewing the diff in a browser, making it easier to spot potential bugs._
- [ ] Automated tests: Did you add appropriate automated tests for any code changes?
- [ ] Code coverage: Did you check the code coverage report for the automated tests?  _While we are not looking for perfect coverage, the tool can point out potential cases that have been missed._ Code coverage can be generated with: `./gradlew check` for Kotlin modules and `./gradlew connectedCheck -PIS_ANDROID_INSTRUMENTATION_TEST_COVERAGE_ENABLED=true` for Android modules.
- [ ] Run the app: Did you run the app and try the changes?
- [ ] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? _Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit._

# Reviewer

- [ ] How is code coverage affected? _We encourage you to compare coverage before and after changes and when possible, leaving it in a better place._
- [x] Run the app: Did you run the app and try the changes? _While the CI server runs the app to look for build failures or crashes, humans running the app are more likely to notice unexpected log messages, UI inconsistencies, or bad output data. Perform this step last, after verifying the code changes are safe to run locally._

### DIFF
--- a/ui-design-lib/build.gradle.kts
+++ b/ui-design-lib/build.gradle.kts
@@ -35,7 +35,6 @@ dependencies {
     implementation(libs.kotlinx.coroutines.android)
     implementation(libs.kotlinx.coroutines.core)
     implementation(projects.spackleAndroidLib)
-    implementation(projects.sdkExtLib)
 
     androidTestImplementation(libs.bundles.androidx.test)
     androidTestImplementation(libs.androidx.compose.test.junit)
--- a/ui-design-lib/build.gradle.kts
+++ b/ui-design-lib/build.gradle.kts
@@ -35,6 +35,7 @@ dependencies {
     implementation(libs.kotlinx.coroutines.android)
     implementation(libs.kotlinx.coroutines.core)
     implementation(projects.spackleAndroidLib)
+    implementation(projects.sdkExtLib)
 
     androidTestImplementation(libs.bundles.androidx.test)
     androidTestImplementation(libs.androidx.compose.test.junit)
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -6,8 +6,6 @@
           <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
         </value>
       </option>
-      <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
-      <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="2147483647" />
     </JetCodeStyleSettings>
     <codeStyleSettings language="XML">
       <option name="FORCE_REARRANGE_MODE" value="1" />
--- a/ui-design-lib/src/androidTest/java/co/electriccoin/zcash/ui/design/component/ButtonTest.kt
+++ b/ui-design-lib/src/androidTest/java/co/electriccoin/zcash/ui/design/component/ButtonTest.kt
@@ -9,7 +9,6 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
 import androidx.test.filters.MediumTest
-import cash.z.ecc.android.sdk.ext.collectWith
 import co.electriccoin.zcash.ui.design.theme.ZcashTheme
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
@@ -48,8 +47,8 @@ class ButtonTest {
         }
 
         launch {
-            testSetup.mutableActionExecuted.collectWith(this) {
-                if (!it) return@collectWith
+            testSetup.mutableActionExecuted.collect {
+                if (!it) return@collect
 
                 assertTrue { mark.elapsedNow() >= 2.seconds }
                 this.cancel()
--- a/ui-design-lib/src/androidTest/java/co/electriccoin/zcash/ui/design/component/ButtonTest.kt
+++ b/ui-design-lib/src/androidTest/java/co/electriccoin/zcash/ui/design/component/ButtonTest.kt
@@ -1,0 +1,94 @@
+package co.electriccoin.zcash.ui.design.component
+
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.interaction.PressInteraction
+import androidx.compose.material3.Text
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.test.junit4.ComposeContentTestRule
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.test.filters.MediumTest
+import cash.z.ecc.android.sdk.ext.collectWith
+import co.electriccoin.zcash.ui.design.theme.ZcashTheme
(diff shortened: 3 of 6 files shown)


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/Button.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+@Composable
+fun TimedButton(duration: Duration = 5.seconds, onClick: () -> Unit, content: @Composable RowScope.() -> Unit) {
```
ccjernigan, line 19: This implementation looks reasonable.

Can you add a test for it?

ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/common/ButtonTest.kt
```diff
@@ -0,0 +1,84 @@
...
+class ButtonTest {
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @After
+    fun tearDown() {
+        Dispatchers.resetMain()
+    }
+
```
ccjernigan, line 42: Can you add one more test case that tries doing a simple button press, verifying that a simple normal button click does not trigger the button?

ui-lib/src/androidTest/java/co/electriccoin/zcash/ui/common/ButtonTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    fun buttonClickTest() = runTest {
+        val testDispatcher = StandardTestDispatcher(testScheduler)
+        val testSetup = newTestSetup(testDispatcher, 2.seconds)
+
+        composeTestRule.onNodeWithText("button").also {
+            it.performClick()
+        }
+        advanceTimeBy(3.seconds.inWholeMilliseconds)
+
+        launch {
```
ccjernigan, line 78: Can this be simplified down to just reading the value?

gasstan: without the launch block i can't get it to work. I'm always get After waiting for 60000 ms, the test coroutine is not completing. But when i try to cancel it like so : 
``` 
testSetup.mutableActionExecuted.collect {
    if (!it) return@collect
    assertTrue { mark.elapsedNow() >= 2.seconds }
    this.cancel()
} 
```
then i get TestScopeImpl was cancelled; job=TestScope[test ended]
so i thought using the launch was right approach

ccjernigan: Why can you not make this very very simple? `assertFalse(testSetup.mutableActionExecuted.value)`

ccjernigan (dismissed):

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/Button.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+@Suppress("LongParameterList")
+@Composable
+fun TimedButton(
+    modifier: Modifier = Modifier,
+    duration: Duration = 5.seconds,
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    coroutineDispatcher: CoroutineDispatcher = Dispatchers.Default,
```
ccjernigan, line 26: While I like default arguments for things that are test-only, I do have concerns when we are propagating them across many different layers like is happening in the recent changes.

It makes it much more likely that someone will pass in a different value at the top level but accidentally an intermediate layer forgets to pass that value through all the way to the end.

Can you provide some context why we're injecting a dispatcher and whether other options have been considered?

gasstan: 1. Main reason is that I wasn't able to test TimedButton without passing the test dispatcher, because it launches a new coroutine to introduce the delay. I'm not sure if there's more elegant way to do this, but if you have something in mind please let me know.
2. As for default argument, i agree that it's not ideal to have default param on every layer, but are you saying that we should avoid using default params ?

ccjernigan:
@HonzaR I added you as a reviewer because I'd like to get your perspective on the coroutine dispatcher.  While it is helpful for testing and having it at the layer of the button is OK, I dislike that the interaction source and the dispatcher are now part of the public API of the top level `Send` composable.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/Button.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+@Suppress("LongParameterList")
+@Composable
+fun TimedButton(
```
HonzaR, line 22: May we move this new component into the ui-design-lib module to its Button class?

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/Button.kt
HonzaR: Did you try to run the delay above `Dispatchers.Default` and return it to `Dispatchers.Main` (as you already do)? We'd be able to get rid of the one from parameter.

HonzaR: We discussed this question today and we came up with a few ideas. Basically, we have these ideas:
- Have the duration as a parameter and input there the real value (5s) for production calls and some test (e.g. 0.5s) while calling this api from test context
- We could also go with something similar to Android Views approach - measure a difference between press_down and press_up events in onTouchEvent callback and compare it with our constant value (5s)
- Or provide both callbacks in compose code. One for normal click and second long click and in the test SendViewTest we'd be ok with just click event, and in the separate TimedButtonTest we'd test the long press. It'd be useful, because I believe that we'll still need to catch also normal click to e.g. toast user that the long click is necessary or for showing some countdown to user, so he/she understands what time is needed.

gasstan:
filed folowup issue https://github.com/zcash/secant-android-wallet/issues/676

ui-design-lib/build.gradle.kts
```diff
@@ -35,6 +35,7 @@ dependencies {
     implementation(libs.kotlinx.coroutines.android)
     implementation(libs.kotlinx.coroutines.core)
     implementation(projects.spackleAndroidLib)
+    implementation(projects.sdkExtLib)
```
ccjernigan, line 38: The design library probably shouldn't be dependent on the SDK or SDK extensions

ccjernigan (approved):

Issue #1249, opened 2024-02-14
Add documentation and tests on background syncing


HonzaR:
## Is your feature request related to a problem? Please describe.
We'd like to add:
- a new `.md` documentation file describing the new Background block synchronization feature
- a manual test guide to induce the bg sync feature running
- a test that would verify the proper time setting on the bg sync trigger

PR #1251, opened 2024-02-14
[#1151] Background syncing

### LINKED ISSUE #634: Wallet synchronisation continues even when app is in background
### Describe the issue
Walet synchronisation is not paused when app is brought to background

#### If so, please list the steps to reproduce below:
1. install app
2. onboard wallet
3. when app is syncing leave the app
4. logcat continues to show sync updates

- App Version: 0.1-debug
- Android Version: 12
- Device: (if applies)

### LINKED ISSUE #1151: Background syncing
The goal is to be able to sync in the background. There are several possible scenarios where each one needs different handling:
- syncing at the time people sleep - the power + wifi are required
- being able to finish some request when going to the background - e.g. when send is initiated and a user doesn't wait until the request is done

All information and investigation outcomes will be gathered in this ticket.

Some form of background syncing was supposedly added back in #191 but I was unaware of it existing. We should check whether it is indeed working in the same way as https://github.com/Electric-Coin-Company/zashi-ios/issues/982 will work, and if not we should bring Zashi Android up to parity (as much as is possible on the Android platform).

HonzaR:
- A periodic background block synchronization has been added. When the device is connected to the internet using an unmetered connection and is plugged into the power, the background task will start to synchronize blocks randomly between 3 and 4 a.m.
- The background worker was in place but not fully working, plus was set to trigger randomly in 24 hours
- Changelog update
- Closes #1151
- Closes #634
- Its follow-up #1249

> **Note**

# Author

> **Note**

# Reviewer

- [x] Check the code with the [Code Review Guidelines](../blob/main/docs/CODE_REVIEW_GUIDELINES.md) **checklist**

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
@@ -84,6 +84,8 @@ class SyncWorker(context: Context, workerParameters: WorkerParameters) : Corouti
                     .setRequiresCharging(true)
                     .build()
 
+            // TODO [#1258]: Consider using flexInterval in BG sync trigger planning
+            // TODO [#1258]: https://github.com/Electric-Coin-Company/zashi-android/issues/1258
             return PeriodicWorkRequestBuilder<SyncWorker>(SYNC_PERIOD.toJavaDuration())
                 .setConstraints(constraints)
                 .setInitialDelay(targetTimeDiff.toJavaDuration())
--- a/ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
@@ -12,55 +12,115 @@ import cash.z.ecc.android.sdk.Synchronizer
 import cash.z.ecc.android.sdk.WalletCoordinator
 import cash.z.ecc.android.sdk.model.PercentDecimal
 import co.electriccoin.zcash.global.getInstance
+import co.electriccoin.zcash.spackle.Twig
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.takeWhile
+import kotlinx.datetime.Clock
+import kotlinx.datetime.DateTimeUnit
+import kotlinx.datetime.Instant
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.atTime
+import kotlinx.datetime.toInstant
+import kotlinx.datetime.toLocalDateTime
+import kotlinx.datetime.until
+import kotlin.random.Random
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.days
 import kotlin.time.Duration.Companion.hours
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
 import kotlin.time.toJavaDuration
+
+// TODO [#1249]: Add documentation and tests on background syncing
+// TODO [#1249]: https://github.com/Electric-Coin-Company/zashi-android/issues/1249
 
 @Keep
 class SyncWorker(context: Context, workerParameters: WorkerParameters) : CoroutineWorker(context, workerParameters) {
     @OptIn(ExperimentalCoroutinesApi::class)
     override suspend fun doWork(): Result {
-        // Enhancements to this implementation would be:
-        // - Quit early if the synchronizer is null after a timeout period
-        // - Return better status information
+        Twig.debug { "BG Sync: starting..." }
 
         WalletCoordinator.getInstance(applicationContext).synchronizer
             .flatMapLatest {
+                Twig.debug { "BG Sync: synchronizer: $it" }
+
                 it?.status?.combine(it.progress) { status, progress ->
-                    StatusAndProgress(status, progress)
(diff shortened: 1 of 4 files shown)


LukasKorba (approved):

ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
```diff
@@ -12,55 +12,115 @@ import cash.z.ecc.android.sdk.Synchronizer
...
         fun newWorkRequest(): PeriodicWorkRequest {
+            val targetTimeDiff = calculateTargetTimeDifference()
+
+            Twig.debug { "BG Sync: necessary trigger delay time: $targetTimeDiff" }
+
             val constraints =
                 Constraints.Builder()
                     .setRequiresStorageNotLow(true)
-                    .setRequiredNetworkType(NetworkType.CONNECTED)
+                    .setRequiredNetworkType(NetworkType.UNMETERED)
```
nuttycom, line 83: One question; since this is a change to an existing method, doesn't this mean that this now makes it that sync can _only_ happen on wifi and while charging?

nuttycom (changes requested):
Can you confirm that after this change, that it's still possible to sync normally, i.e. if I'm in the coffee shop and open my wallet, it will complete the sync to the chain tip from the last time my wallet synced in the background?

Marking as changes requested pending the answer to this question.

ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
HonzaR: The existing solution of background syncing wasn't working as expected (see #634) and had a problem with its triggering because of the `it.progress.isLessThanHundredPercent()` condition, although it was enabled for users. These issues are now fixed.

HonzaR:
> Can you confirm that after this change, that it's still possible to sync normally, i.e. if I'm in the coffee shop and open my wallet, it will complete the sync to the chain tip from the last time my wallet synced in the background?

Yes, this is definitely a different case from the regular foreground syncing. This one is triggered from the background via `SyncWorker` that calls `WalletCoordinator` inside it. The foreground sync path is calling the same `WalletCoordinator` component but from the ViewModel, which is called from UI. So those are separated and not related sync paths.

nuttycom (approved):
utACK b1c457bc113edc9d58fb4ad1425bfeec773dcb6f

ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
```diff
@@ -12,55 +12,115 @@ import cash.z.ecc.android.sdk.Synchronizer
...
+                    .setRequiresCharging(true)
                     .build()
 
-            return PeriodicWorkRequestBuilder<SyncWorker>(DEFAULT_SYNC_PERIOD.toJavaDuration())
+            return PeriodicWorkRequestBuilder<SyncWorker>(SYNC_PERIOD.toJavaDuration())
```
str4d, line 87: The documentation for this constructor says:
> The [`PeriodicWorkRequest`](https://developer.android.com/reference/androidx/work/PeriodicWorkRequest) is guaranteed to run exactly one time during this interval (subject to OS battery optimizations, such as doze mode). [..] It may run immediately, at the end of the period, or any time in between so long as the other conditions are satisfied at the time.

AFAICT this means that the interval will start at some time `X` between 3am and 4am (due to `setInitialDelay`), but that `SyncWorker` can be triggered at any time between `X` and `X + SYNC_PERIOD` (i.e. it could sync at 2pm, if `constraints` happen to be met at that time). Is this the intended behaviour? I don't know how likely it is that jobs are always run at the beginning of their interval.

If we actually want to require that the background sync only be triggered to start between 3am and 4am, then the alternative "flex period" constructor would be more suitable: we'd set the interval to start at 4am, with a 1 hour `flexInterval` so that the flex perid is 3-4am (`repeatInterval - flexInterval` to the end of the interval, per the docs).

str4d (approved):
utACK b1c457bc113edc9d58fb4ad1425bfeec773dcb6f, with a question about how the background work is scheduled.

ui-lib/src/main/java/co/electriccoin/zcash/work/SyncWorker.kt
str4d: A side-effect of using the flex period would be that the first background sync would be on the second night after first installing the app (as we can only shift the start point of the interval with `setInitialDelay`, and the flex period is the last `flexInterval` of the interval). But maybe there's a way around that by also scheduling a one-off task for the first night.

HonzaR: Thanks for pointing this out. Yes, I was aware of the `flexInterval` when implementing this. It is one option to consider for our use case. The "second night" drawback was something that I didn't know how to by-pass, and from my experience and testing, the trigger usually happened at the beginning of the `repeatInterval` as we have limited it with the **unmetered connection** and **is plugged into the power** constraints under which is the system in a perfect position to start any planned worker. 

My second thought is that we might have only some of the users having their devices plugged in and on wifi at night, so the trigger for those will start once these constraints are met, and that could be at any time during the day. Thus, the 3-4 a.m. requirement would not be met, but the main goal of a regular sync will be met.

We can still file a follow-up and revisit the sync planning again. What do you think?

str4d: As long as we anticipate user feedback of the form "my wallet isn't syncing" because the sync process is getting triggered when their phone is plugged in for a short time during the day, instead of for a long time at night, then I think this is fine to merge as-is and we can revisit this later if needed.

HonzaR: Filed #1258.

Issue #1257, opened 2024-02-16
Send.Form TextFields not persisted on a configuration change


HonzaR:
## Is your feature request related to a problem? Please describe.
The Send.Form screen TextFields keep their content in case of any configuration change when the underlying ViewPager is set to the Send screen or adjacent pages (Account, Receive), but not when this happens on the Balances page, as it's out of the ViewPager's preloaded screens.

Issue #1144, opened 2023-12-29
Send screen


HonzaR:
## Is your feature request related to a problem? Please describe.
Add the proper UI and logic for the Send screen according to the Figma design.

Issue #1047, opened 2023-11-16
Consistent representation of Zatoshi amounts in UI


HonzaR:
## Is your feature request related to a problem? Please describe.
We'd like to update the wallet ZEC representation according to the rules below.

## Describe the solution you'd like
```
Expanded amounts look like this: “1.12345678”

Or like this: “1.12345”

(I.e., omit trailing least-significant 0 digits.)

Abbreviated amounts look like this:

If the most significant digits are non-zero then “1.123”, regardless of whether the least-significant digits are zero or not.
If the most significant digits are zero but the least significant ones are non-zero then “0.000…”
If all the digits are zero then “0.000”.

Interactable amounts (such as the transaction amounts shown in the transaction row on the home screen) are shown abbreviated.

Non-interactable amounts are shown expanded.

For fees, specifically, they always say “(Fee < 0.001)”. (Fees are always < 0.001 ZEC.)
```

The rules are already implemented on the iOS side: https://github.com/Electric-Coin-Company/zashi-ios/pull/921/files#diff-b9a6370acdd507de92ab4346561978d98be6b09200034bd66faf1f39c5908988

PR #1265, opened 2024-02-26
[#1144] Send.Form screen rework

### LINKED ISSUE #340: Design: Display Errors When Sending Zec
The send ZEC screen should provide good feedback to the user about errors on the form input.  Some examples of errors:

- [x] Amount too small (cannot cover fee)
- [x] Amount too large (exceeds current balance)
- [x] Invalid amount
- [x] Bad address
- [x] Insufficient funds
- [x] Invalid address

### LINKED ISSUE #810: Disable Memo UI field in case of Transparent address
We'd like to improve our Send screen UI with mechanisms of hiding/disabling Memo UI field in case of Transparent address detected as receiver. 

Keep the Memo field in place and disable/enable it while receiver address validated and recognised its type. Show the Memo field in the red error state and add there an error message below it.  Also don't let user to hit the Send button (disable it) unless all characters from the Memo field removed.

Validation of the receiver address could be done with simply `t` starting character checking for now.

## Alternatives you've considered
Alternatively we could consider usage of `isValid...` methods from the SDK. 

These methods maybe not need to be suspend
Maybe we could move these addresses validations from `Synchronizer` to Wallet Address object, and we need then to keep the suspend https://github.com/zcash/zcash-android-wallet-sdk/issues/342. We have an issue to make the wallet address validation more type safe.  Ultimately, we probably need one more companion function on WalletAddress in `sdk-incubator-lib` that takes in a string (for address and returns the correct sealed class type).

### LINKED ISSUE #826: SendArgumentsWrapper object properties validation
We'd like to have the `SendArgumentsWrapper` class properties validated, and thus create a type-safe model class.

We have a concern that the `SendArgumentsWrapper` is not type-safe, so the fields contained within it may not necessarily be valid. This would allow bypassing the live validation we have as the user types, meaning incorrect values could get pre-filled into the Send.Form screen text fields.

We're not sure about the most suitable solution here, but it could be e.g.
- Adding type safety to the `SendArgumentsWrapper` so that we skip scanning bad QR codes
- Adding validation to Send.Form screen before we use the values from `SendArgumentsWrapper`
- Adding validation before we insert values to the `SendArgumentsWrapper` model data class

We realize that the arguments we pass through the Navigation library are all strings, so that complicates type safety a bit. We could validate before serializing to strings, then assume they're safe upon deserializing since it's an internal API.

### LINKED ISSUE #1157: Send Form memo length counter

### LINKED ISSUE #1158: Send Form memo field availability

### LINKED ISSUE #1253: Change fee label to say "typical fee < 0.001"
We will be enabling ZIP 317 fees for 1.0; this ensures that if we don't have time to expose the actual fees, users are still being told something true.

### LINKED ISSUE #1254: Ensure that all errors related to transaction creation are shown to users on the send screen
#340 is about doing real time validation of inputs as users fill in the Send screen. This is a simpler issue that could be solved by removing all real time validation and instead relying on the existing error message reporting of send-time errors (though once #1159 is done, those errors would show on the send confirmation screen instead, which is less great but still better than the user not knowing why they can't click the Send button).

HonzaR:
- This follows the new Figma design. It also adds the TextFields values validation and proper UI reactions.
- Closes #340
- Closes #810
- Closes #1157
- Closes #1158
- Closes #1253
- Closes #1254
- Closes #826
- Follow-ups: #1047, #1257

**AFTER CHANGES:** (Note the screen rotation only demonstrates the TextFields proper values persisting rather than a real-world situation, as Zashi has rotation currently locked in the portrait mode)

https://github.com/Electric-Coin-Company/zashi-android/assets/9017390/e15194fc-7ce8-4be4-84ed-6142204677e8

**BEFORE CHANGES:**

> **Note**

# Author

> **Note**

# Reviewer


daira:
I noticed from the video that the memo was not cleared when you pasted in the t-address. I think this is different from Zashi iOS.

ui-design-lib/src/main/java/co/electriccoin/zcash/ui/design/component/TextField.kt
```diff
@@ -31,40 +32,53 @@ fun FormTextField(
...
     textStyle: TextStyle = ZcashTheme.extendedTypography.textFieldValue,
-    label: @Composable (() -> Unit)? = null,
-    leadingIcon: @Composable (() -> Unit)? = null,
-    trailingIcon: @Composable (() -> Unit)? = null,
+    placeholder:
+        @Composable()
+        (() -> Unit)? = null,
+    leadingIcon:
+        @Composable()
+        (() -> Unit)? = null,
```
LukasKorba, line 43: What leadingIcon does? Seems like it's something from the previous design we had months or even years ago. Do we want to keep it or clean up the code?

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/compose/BalanceWidget.kt
```diff
@@ -40,19 +40,21 @@ private fun BalanceWidgetPreview() {
...
+                                WalletBalance(
+                                    Zatoshi(1234567891234567),
+                                    Zatoshi(123456789),
+                                    Zatoshi(123)
+                                )
+                        ),
+                    isReferenceToBalances = true,
+                    onReferenceClick = {},
+                    modifier = Modifier
+                )
```
LukasKorba, line 57: This is more about my curiosity, I checked the raw file and the only change here is that exactly the same code has been wrapped into parentheses (). Is it by accident or it has some meaning?

HonzaR:
> I noticed from the video that the memo was not cleared when you pasted in the t-address. I think this is different from Zashi iOS.

Yes, it's not cleared from the memo field, but the field is disabled, and the message from the memo is not used when the Preview button hits. The pro of this solution is that the message value is not lost when the address is changed to a unified or sapling one. @LukasKorba @true-jared @adjyrichard What do you think?

ui-design-lib/src/main/java/co/electriccoin/zcash/ui/design/component/TextField.kt
HonzaR: The TextField is a reusable component accessible from our `ui-design-lib` module, which makes it an API with a different use case across the app's UI. Having the leading icon parameter defined as nullable is fine even if not used. The `trailingIcon` parameter is used on the Send.Form for the recipient address TextField, so logically, the API could also provide the `loadingIcon.`

LukasKorba (approved):
utACK, 2 non-blockers

LukasKorba:
> > I noticed from the video that the memo was not cleared when you pasted in the t-address. I think this is different from Zashi iOS.
> 
> Yes, it's not cleared from the memo field, but the field is disabled, and the message from the memo is not used when the Preview button hits. The pro of this solution is that the message value is not lost when the address is changed to a unified or sapling one. @LukasKorba @true-jared @adjyrichard What do you think?

iOS is clearing it as it was required. If Android is behaving differently I think it's better if it's discussed prior PR.

LukasKorba:
I noticed in the video that memo field's height is expanding, it was supposed to be fixed size.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/compose/BalanceWidget.kt
HonzaR: Having it in parentheses is IMHO not necessary and was caused by Ktlint auto-formtting.

true-jared:
@HonzaR We have discussed the memo behaviour, and made a decision to clear the message. Regarding the fixed height/expandable height of the input fields, I understand your UX perspective but I 100% agree that making any changes to the requirements/designs while working on an issue should be raised prior to making a PR and definitely before merging any changes you have made. I don't want to hold you back and make you rework anything right now, but please respect the process going forward. We can discuss/track any improvements you would like to add, but to meet the release deadline, we need to focus on the hard functional requirements first, and not spend time on nice-to-haves.

true-jared:
Filed here: #1269 @HonzaR .

PR #1462, opened 2024-06-04
ISSUE-1412 Receive Page split into horizontal pager

Milan-Nerevar:
Snímka obrazovky 2024-06-04 o 17 56 36

The design is not 100% pixel perfect because didnt have time, but I guess its quite enough for a demo. Also scroll is not broken because i disabled the horizontal pager one so u can still swipe between bottom pages.


ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
```diff
@@ -170,73 +162,139 @@ private fun ReceiveContents(
...
+private fun AddressTabs(
+    modifier: Modifier,
+    pagerState: PagerState,
+    tabs: List<WalletAddress>,
+    coroutineScope: CoroutineScope = rememberCoroutineScope(),
+    onTabSelected: (index: Int) -> Unit = {},
+) {
+    TabRow(
+        modifier = modifier
+            .padding(horizontal = 64.dp)
```
HonzaR, line 216: We could use one of the predefined padding sizes from `dimens` to keep paddings as close as possible across different screens. This comment applies to more occurrences below.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
```diff
@@ -170,73 +162,139 @@ private fun ReceiveContents(
     versionInfo: VersionInfo,
     modifier: Modifier = Modifier,
 ) {
-    Column(
-        modifier =
-            Modifier
-                .fillMaxHeight()
-                .verticalScroll(rememberScrollState())
```
HonzaR, line 177: We still need to use these modifiers to keep the below content vertically scrollable.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
```diff
@@ -170,73 +162,139 @@ private fun ReceiveContents(
...
-        Spacer(modifier = Modifier.height(ZcashTheme.dimens.spacingHuge))
-
-        Address(
-            walletAddress = walletAddress.transparent,
-            onAddressCopyToClipboard = onAddressCopyToClipboard,
-            onQrImageShare = onQrImageShare,
-        )
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+private fun AddressTabs(
```
HonzaR, line 207: I'm thinking about moving this UI component to the design-lib module so that it is available for possible other usage, the same as we do with the buttons, for example. What do you think?

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
```diff
@@ -170,73 +162,139 @@ private fun ReceiveContents(
...
+    walletAddress: WalletAddress,
+    selected: Boolean,
+    onClick: () -> Unit,
+) {
+    Tab(
+        modifier = Modifier
+            .fillMaxWidth(),
+        selected = selected,
+        onClick = onClick,
+        selectedContentColor = Color.Black,
```
HonzaR, line 248: We could rather use the predefined colors from `ZcashTheme.colors` to make it easier for us once we finish the dark mode integration.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
```diff
@@ -170,73 +162,139 @@ private fun ReceiveContents(
...
+    if (screenBrightnessState == ScreenBrightnessState.FULL) {
+        BrightenScreen()
+        DisableScreenTimeout()
+    }
+
+    val state by remember {
+        derivedStateOf {
+            listOfNotNull(
+                walletAddress.unified,
+                walletAddress.sapling.takeIf { versionInfo.isTestnet },
```
HonzaR, line 174: Should we bring another UI solution instead of making the new tabs toggle three-buttons? As those surely won't fit. I propose moving the Sapling part (which is visible only in the Testnet version of the app) below the Unified one with its own title.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
```diff
@@ -170,73 +162,139 @@ private fun ReceiveContents(
...
+        Box(
+            modifier = Modifier
+                .fillMaxSize()
+                .background(
+                    if (selected) Color.Black else Color.Transparent
+                )
+                .padding(vertical = 12.dp, horizontal = 2.dp),
+            contentAlignment = Alignment.Center,
+        ) {
+            Text(
```
HonzaR, line 260: We could improve this a bit by making the text max two lines and ellipsis it if a longer text comes to be covered on very small devices and long locales.

HonzaR (commented):
Thank you for the changes. It's very appreciated! I left some comments inline and a few notes here below:
1. You seem to have different AS editor settings that caused several unnecessary changes in indentation.
2. Although it's not perfectly visible at first look, I think we need to revert the UI logic so that the selected tab has a white button state and the second one has only black surrounding it.
3. I've tried the proposed changes on a split-screen device to emulate a very small screen, and the address tabs do not seem to be vertically scrollable anymore. You can also test by rotating the device to landscape, which is supported in the debug build type.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
Milan-Nerevar: Yes, totally doable. Judging from figma there are several occurrences of these tabs so yes they should be reused

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
Milan-Nerevar: Yeh I tested those - they fit because the texts wraps into 2 lines, but definitely not ideal. No figma design for this either

Milan-Nerevar: But as u say, moving it to an existing tab could work

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/receive/view/ReceiveView.kt
Milan-Nerevar: Totally. In general typo & colors should be taken from theme instead of being hardcoded like this

Screenshot_20240604_175356.png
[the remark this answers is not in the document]
Milan-Nerevar: lol idk how it go there, probably I hit ctrl+v to editor too, instead of only here on github, before pushing the changes :D

Milan-Nerevar:
@HonzaR whats the next coarse of action? Should I fix up the PR or?

HonzaR:
> @HonzaR whats the next coarse of action? Should I fix up the PR or?

Yes, let's update the PR to resolve the review comments, if you like? Make sure to fix the CI actions reports. I'll approve the CI run for you then. Thank you.

Milan-Nerevar:
@HonzaR  u can recheck, but probably lint will fail - even tho I am using project code style

HonzaR:
> @HonzaR u can recheck, but probably lint will fail - even tho I am using project code style

Hi @Milan-Nerevar, Thank you for the update commits! The static code analysis checks are still failing. It's a good approach to check your changes locally before pushing. See the predefined [**Gradle tasks**](https://github.com/Electric-Coin-Company/zashi-android/blob/main/docs/Setup.md#gradle-tasks) in our Setup documentation. Would you like to fix those too?

HonzaR (commented):
I've retested it, and it works well now. Thank you for fixing the commits!

Before merging it, we should resolve one last issue I have. In the Testnet variant (unified and sapling), or when rotated to the landscape orientation, there is an unnecessary blank bottom padding.

HonzaR (approved):

Issue #1421, opened 2024-05-01
Implement localization in Spanish


true-jared:
- needs to be scoped out properly and we need to prepare the translation files

PR #1651, opened 2024-10-23
[#1421] [WIP] es translation

y4ssi:


ui-lib/src/main/res/ui/advanced_settings/values-es/strings.xml
```diff
@@ -1,9 +1,9 @@
 <resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
-    <string name="advanced_settings_title">Advanced Settings</string>
-    <string name="advanced_settings_recovery">Recovery Phrase</string>
-    <string name="advanced_settings_export">Export Private Data</string>
-    <string name="advanced_settings_choose_server">Choose a Server</string>
-    <string name="advanced_settings_currency_conversion">Currency Conversion</string>
-    <string name="advanced_settings_info">You will be asked to confirm on the next screen</string>
-    <string name="advanced_settings_delete_button">Delete Zashi</string>
```
daira, line 8: @true-jared wrote:
> "Delete Zashi" is confusing, as doing this doesn't actually delete the app. "Delete wallet/account" may be better.

What this is actually doing is deleting access to the wallet from the device. It may or may not be backed up.

ui-lib/src/main/res/ui/send_confirmation/values-es/strings.xml
```diff
@@ -1,52 +1,46 @@
...
+    <string name="send_confirmation_address">Enviando a</string>
+    <string name="send_confirmation_memo">Mensaje</string>
+    <string name="send_confirmation_fee">Tarifa</string>
+    <string name="send_confirmation_amount_item">Monto</string>
+    <string name="send_confirmation_send_button">Enviar</string>
+    <string name="send_confirmation_back_button">Cancelar</string>
 
-    <string name="send_confirmation_dialog_error_title">Transaction Failed</string>
-    <string name="send_confirmation_dialog_error_text">An error occurred and the attempt to send funds failed. Try it again, please.</string>
+    <string name="send_confirmation_dialog_error_title">Transacción Fallida</string>
```
HonzaR, line 12: Now I see that here we use camel-case.

ui-lib/src/main/res/ui/export_data/values-es/strings.xml
```diff
@@ -1,14 +1,9 @@
...
-        funds, only the ability to see what you do with your funds.</string>
-    <string name="export_data_confirm">Export private data</string>
-    <string name="export_data_agree">I agree</string>
-    <string name="export_data_export_data_chooser_title">Share internal Zashi data with:</string>
-    <string name="export_data_unable_to_share">Unable to find an application to share with.</string>
-</resources>
\ No newline at end of file
+    <string name="export_data_header">Consentimiento para Exportar Datos Privados</string>
+    <string name="export_data_text_1">Al hacer clic en "Estoy de acuerdo" a continuación, das tu consentimiento para exportar los datos privados de Zashi, lo cual incluye todo el historial de la billetera, toda la información privada, los memos, los montos y las direcciones de los destinatarios, incluso para tu actividad protegida.*\n\nEstos datos privados también permiten ver ciertas acciones futuras que realices con Zashi.\n\nCompartir estos datos privados es irrevocable: una vez que hayas compartido estos datos privados con alguien, no hay forma de revocar su acceso.</string>
```
HonzaR, line 3: We need to keep all in-sentence `"` escaped like this `\"`. Otherwise, they won't be seen after they are laid out by the UI framework.

HonzaR (commented):
Thank you for the translation. It's a lot of good work! I did first pass through the new texts

I just left several comments inline. I also synced the underlying `1421-initial-spanish-localization` branch with main. It brings a few new texts in `res/ui/authentication/values-es/strings.xml`. 

As part of the underlying branch recent changes, I also updated `src/main/res/ui/common` to remove two texts that are not meant to be translated and I moved them to the new location `src/main/res/ui/non_translatable` that we can keep as is. Thank you.

PR #1853, opened 2025-04-14
Say "Shielded Privacy" for UA that contain a T-address

emersonian:
When a user views their Zashi Shielded Address, it is described as having "Maximum Privacy".

In reality, the UA has an embedded T-address, rendering it highly-likely that the user is doxxed by a motivated adversary.

Long-term let's allow Orchard-only UAs like in other ZEC wallets, perhaps as a "lockdown mode" expert feature in settings.

User story:

1. Human rights activist/anon swaps USDC to ZEC using a swap interface which only supports T-addresses (NEAR, Maya).
2. User shields their transparent ZEC into the Orchard pool using Zashi.
3. User posts their "Zashi Shielded Address" publicly on their social media profile, feeling safe by the phrase "Maximum Privacy".
4. Adversary sees their T-address in the UA, searches the address in NEAR or Maya's transaction history, determines their USDC came from a KYC exchange, is able to identify the activist.

### DIFF
--- a/ui-lib/src/main/res/ui/qr_code/values/strings.xml
+++ b/ui-lib/src/main/res/ui/qr_code/values/strings.xml
@@ -7,7 +7,7 @@
     <string name="qr_code_wallet_address_shielded">Zcash Shielded Address</string>
     <string name="qr_code_wallet_address_sapling">Zcash Sapling Address</string>
     <string name="qr_code_wallet_address_transparent">Zcash Transparent Address</string>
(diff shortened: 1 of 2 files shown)


pjv:
If that’s all it takes to see the t-address embedded in the UA, it should say “no privacy”.

true-jared:
@emersonian As discussed on the forum, this will get addressed by https://github.com/Electric-Coin-Company/zashi-android/issues/1871.

PR #2178, opened 2026-04-05
refactor: use success token for RestoreTorView

viktorcrypt:
## Summary

Replace the hardcoded active switch color in the Restore Tor screen with an existing success token from `ZashiColors`.

## Why

- keeps the screen aligned with the design system
- removes a raw hex color from UI code
- keeps the change small and localized


nesence-m (approved):

nesence-m:
updated

PR #2218, opened 2026-05-11
ci: publish to self-hosted F-Droid repo + switch to ZODL GPG signer

y4ssi:
## Summary

- Replace GCS + Cloud KMS GPG fetch with **AWS OIDC + Secrets Manager**. The release job assumes `zodl-android-release-github-actions` and reads `/release/gpg-signing-key` directly — no more `gsutil` or `gcloud kms decrypt`.
- Switch GPG detached signatures from `sysadmin@z.cash` (ECC key) to `sysadmin@zodl.com` (ZODL key `0338…6AB1`). The ECC key is scheduled for revocation 2026-06-23; existing release `.asc` files remain verifiable until then.
- Drop the redundant *Build FOSS (F-Droid) APK* step. A single FOSS `externallibs` build now feeds both Solana and F-Droid.
- New **`publish_to_fdroid`** job (parallel to `publish_to_solana`): pulls the FOSS APK from the GitHub Release, runs `fdroid update` signed by the Secrets-Manager-held index keystore, and pushes the result to `zodl-inc/fdroid` over SSH. GitHub Pages auto-publishes to **foss.zodl.com**.
- **README**: swap the F-Droid badge target to `foss.zodl.com` and embed the repo fingerprint + `fdroidrepos://` deeplink.

## Why our own F-Droid repo

The upstream `gitlab.com/fdroid/fdroiddata` build farm has no access to our build-time secrets (Flexa/CMC/Crashlytics), so its output is feature-stripped relative to the Play Store build. Our self-hosted repo builds with the full secret set while remaining fully-FOSS-compatible (same repo layout, same signing semantics).

## Signing keys — unchanged vs new

| Purpose | Key | Status |
|---|---|---|
| APK signing | `UPLOAD_KEYSTORE_BASE_64` (existing GH secret) | **unchanged** — clients keep getting updates without reinstall |
| `.asc` detached sigs on Release artifacts | ZODL GPG key `sysadmin@zodl.com` | **new** (replaces ECC `sysadmin@z.cash`) |
| F-Droid index (`index-v1.jar`, `index-v2.json`) | `/infra/zodl-android/fdroid_index_keystore` | new |

## Companion PR

Infra (IAM role, Secrets Manager, DNS, repo seed content, runbook): [zodl-inc/infra#2](https://github.com/zodl-inc/infra/pull/2). Merge order does not strictly matter — the new workflow only runs on `release: published`, by which time infra is expected to be applied. Infra has already been applied to AWS.

## Test plan

- [x] YAML is syntactically valid (`python3 -c 'import yaml; yaml.safe_load(...)'`).
- [x] Job names resolve: `validate_gradle_wrapper`, `release`, `publish_to_solana`, `publish_to_fdroid`.
- [x] IAM role, Secrets Manager entries, SSH deploy key on `zodl-inc/fdroid`, and GitHub Pages are provisioned.
- [x] `dig foss.zodl.com` resolves to GitHub Pages.
- [ ] Trigger a GitHub Release (or re-run the last one as `workflow_dispatch`) and confirm:
  - `.asc` detached signatures verify with the ZODL public key.
  - `publish_to_fdroid` pushes a commit to `zodl-inc/fdroid` with `fdroid/repo/{APK, index-v1.jar, index-v2.json}`.
  - F-Droid client at `foss.zodl.com` reports the new version.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,29 +33,19 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Authenticate to Google Cloud for GCS/KMS
-        id: auth_gcs
-        uses: google-github-actions/auth@v3
+      - name: Configure AWS credentials (OIDC)
+        uses: aws-actions/configure-aws-credentials@d979d5b3a71173a29b74b5b88418bfda9437d885 # v6.1.1
         with:
-          credentials_json: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
+          role-to-assume: ${{ secrets.AWS_RELEASE_ROLE_ARN }}
+          aws-region: us-east-1
 
-      - name: Download file from GCS
+      - name: Import GPG signing key from AWS Secrets Manager
         run: |
-          gcloud auth login --cred-file="$GOOGLE_APPLICATION_CREDENTIALS"
-          gsutil -q cp gs://${{ secrets.SOLANA_GCS_BUCKET }}/encrypted_gpg.kms encrypted_gpg.kms
-
-      - name: Decrypt file using KMS
-        run: |
-          gcloud kms decrypt \
-            --key gpg \
-            --keyring gpg \
-            --location global \
-            --plaintext-file private.pgp \
-            --ciphertext-file encrypted_gpg.kms
-
-      - name: Import GPG
-        run: |
-          gpg --import private.pgp
+          aws secretsmanager get-secret-value \
+            --secret-id /release/gpg-signing-key \
+            --query SecretString --output text > private.pgp
+          gpg --batch --import private.pgp
+          shred -u private.pgp
 
       - name: Set up Java
         uses: ./.github/actions/setup-java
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -326,3 +300,135 @@ jobs:
...
+          aws-region: us-east-1
+
+      - name: Load F-Droid index keystore + SSH deploy key from Secrets Manager
+        id: secrets
+        run: |
+          KEYSTORE_JSON=$(aws secretsmanager get-secret-value \
+            --secret-id /infra/zodl-android/fdroid_index_keystore \
+            --query SecretString --output text)
+          echo "$KEYSTORE_JSON" | jq -r '.keystore_b64'         | base64 -d > /tmp/fdroid-index.jks
+          echo "keystore_password=$(echo "$KEYSTORE_JSON" | jq -r '.keystore_password')" >> $GITHUB_OUTPUT
```
noop-sk: **[CRITICAL] `::add-mask::` is called AFTER writing to `$GITHUB_OUTPUT`**

Current order:
```sh
echo "keystore_password=..." >> $GITHUB_OUTPUT   # ← secret already visible in the log
echo "::add-mask::..."                             # ← mask arrives too late
```

The runner may flush output before processing the mask command → the password appears in plain text in the step log.

**Fix — mask first, then write:**
```sh
KEYSTORE_PASS=$(echo "$KEYSTORE_JSON" | jq -r '.keystore_password')
KEY_PASS=$(echo "$KEYSTORE_JSON" | jq -r '.key_password')
echo "::add-mask::${KEYSTORE_PASS}"
echo "::add-mask::${KEY_PASS}"
echo "keystore_password=${KEYSTORE_PASS}" >> $GITHUB_OUTPUT
echo "key_password=${KEY_PASS}"           >> $GITHUB_OUTPUT
```

.github/workflows/release.yaml
```diff
@@ -326,3 +300,135 @@ jobs:
...
+          echo "key_password=$(echo "$KEYSTORE_JSON" | jq -r '.key_password')"           >> $GITHUB_OUTPUT
+          echo "::add-mask::$(echo "$KEYSTORE_JSON" | jq -r '.keystore_password')"
+          echo "::add-mask::$(echo "$KEYSTORE_JSON" | jq -r '.key_password')"
+
+          mkdir -p ~/.ssh
+          aws secretsmanager get-secret-value \
+            --secret-id /infra/zodl-android/fdroid_deploy_ssh_key \
+            --query SecretString --output text > ~/.ssh/id_fdroid
+          chmod 600 ~/.ssh/id_fdroid
+          ssh-keyscan -t ed25519,rsa github.com >> ~/.ssh/known_hosts 2>/dev/null
```
noop-sk: **[Minor] `2>/dev/null` suppresses `ssh-keyscan` errors**

If keyscan fails (DNS issue, timeout), SSH will either connect to an unverified host or fail silently — with no signal in the log.

Remove `2>/dev/null` so real failures surface:
```sh
ssh-keyscan -t ed25519,rsa github.com >> ~/.ssh/known_hosts
```

.github/workflows/release.yaml
```diff
@@ -326,3 +300,135 @@ jobs:
...
+          keystorepass: ${KEYSTORE_PASSWORD}
+          keypass: ${KEY_PASSWORD}
+          EOF
+
+      - name: Run fdroid update
+        working-directory: fdroid-repo/fdroid
+        run: |
+          fdroid update --create-metadata --pretty --verbose
+          # fdroid writes config.yml with secrets — delete before committing.
+          rm -f config.yml
```
noop-sk: **[Minor] `rm -f` on `config.yml` is inconsistent with the rest of the job**

`config.yml` contains plaintext passwords (keystore, key password). Everywhere else in this job, `shred -u` is used for key material — the JKS and the SSH key. Use the same here:
```sh
shred -u config.yml
```

noop-sk (commented):
Clean migration from GCP to AWS overall. One critical security issue with secret masking, two minor notes.

.github/workflows/release.yaml
y4ssi: Fixed in bd27c19. Read the JSON values into shell vars, register the `::add-mask::` for both passwords, **then** write to `GITHUB_OUTPUT`. The runner now sees the mask command before the output line, so the passwords are scrubbed from any subsequent log line. Thanks for catching this — would have leaked the keystore + key passwords on every release run otherwise.

.github/workflows/release.yaml
y4ssi: Fixed in bd27c19. Dropped the `2>/dev/null`; ssh-keyscan errors now surface in the log so a DNS / timeout regression won't silently fall through to an unverified-host SSH push.

.github/workflows/release.yaml
y4ssi: Fixed in bd27c19. Switched to `shred -u config.yml` so the JKS, the SSH deploy key, and now config.yml are all disposed of the same way. Good catch on the inconsistency.

nesence-m (approved):
utACK

PR #2222, opened 2026-05-12
[#2215] Redesign server selection: Automatic/Manual connection mode

### LINKED ISSUE #2215: Redesign server selection: Automatic/Manual connection mode
### Motivation

The current server selection UI forces users to pick a single server from a list. Most users do not know which server is best and should not have to care. There is no "just works" option.

A simpler Automatic/Manual toggle gives users a clear mental model: either the app picks an active server for wallet sync, or the user chooses one explicitly.

### Proposed Solution

* **Automatic mode**: The app evaluates known servers and keeps the wallet endpoint aligned to the best completed server result.
* **Manual mode**: The user selects a single built-in or custom server endpoint for sync.
* Migration: users with a custom server default to Manual, preserving their explicit choice; users on known/default endpoints default to Automatic.
* Persist a `ServerSelection` / `ConnectionMode` model in encrypted preferences under the `server_selection` key.
* Validate custom endpoints before saving manual selections.
* Keep sync and transaction submission on a single selected endpoint for this Android implementation.

czarcas7ic:
Closes: #2215
Resolves MOB-1144.

## Summary

- Adds persisted Automatic and Manual connection modes in Choose Server.
- Automatic mode chooses a known endpoint for wallet sync and submission. Manual mode stores and pins a single endpoint.
- Switching from Automatic to Manual shows the current active endpoint selected. Save persists Manual with that endpoint even if the user does not tap a row.
- Migration keeps missing or default endpoints in Automatic and preserves custom, unknown, or non-default known endpoints in Manual.
- Transaction submission remains single-endpoint in this PR.

## Test plan

- Added model coverage for default, non-default known, custom, missing, JSON round trip, and corrupted JSON migration cases.
- Added Choose Server coverage for Automatic to Manual to Save without an endpoint tap and known/custom endpoint classification.
- Added persistence rollback coverage for endpoint update failure after selection storage.
- Manual testing covered the full server selection flow with the related SDK and Android changes.


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/PersistEndpointUseCase.kt
```diff
@@ -1,31 +0,0 @@
-package co.electriccoin.zcash.ui.common.usecase
```
czarcas7ic, line 1: File replaced by `PersistServerSelectionUseCase`

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/ServerSelection.kt
```diff
@@ -0,0 +1,98 @@
...
+    val isCustom: Boolean = false
+) {
+    init {
+        require(mode == ConnectionMode.AUTOMATIC || endpoint != null) {
+            "Manual server selection requires an endpoint"
+        }
+    }
+
+    fun toJson() =
+        JSONObject().apply {
```
nesence-m: use kotlin serialization instead.

nesence-m (changes requested):

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/ServerSelection.kt
czarcas7ic: https://github.com/zodl-inc/zodl-android/pull/2222/commits/3fbcc3f974dc38ab8d35e7df8b54dd80804c3d99

czarcas7ic:
Followed up on the Android side based on the zodl-ios review feedback:

- [a428dec](https://github.com/zodl-inc/zodl-android/commit/a428dec796eaf55d52a93fb2a34849e23f6fdc89) only reports Flexa `transactionSent` after a successful submit result, matching the iOS `.grpcFailure(txIds:)` feedback.
- [bf0605e](https://github.com/zodl-inc/zodl-android/commit/bf0605ec819246994c0d6cf9bff8b2a9a9055d1e) disables and guards choose-server controls while save is in progress, matching the iOS server setup feedback so the selection cannot change mid-save.

nesence-m:
Separated tx broadcast and automatic server to 2 new PRs

Issue #2216, opened 2026-05-09
Add automatic multi-server transaction broadcast


czarcas7ic:
### Motivation

Android currently submits each transaction to a single lightwalletd server. If that server is slow, unavailable, or rejects a transaction that another server would accept, the send can fail to propagate even though the transaction was created successfully.

Broadcasting through multiple servers in Automatic mode makes transaction submission more reliable while still preserving Manual mode for users who explicitly choose one endpoint.

### Proposed Solution

* Use the Android SDK Broadcaster API to create transactions locally before submitting them.
* **Automatic mode**: submit each created transaction to all known endpoints in parallel.
* **Manual mode**: submit each created transaction only to the user's selected endpoint.
* Treat the first successful endpoint response as the accepted result, then allow a short grace period for best-effort propagation to the remaining endpoints.
* Preserve existing send-result behavior for failures, partial results, and multi-transaction proposals.
* Apply the shared path to regular sends, ZIP321 sends, shielding, swaps, and Keystone/PCZT submissions.

PR #2223, opened 2026-05-12
[#2216] Add multi-server transaction submission

czarcas7ic:
Closes: #2216

## Summary

- Replaces proposal submission internals with a create-then-submit multi-endpoint broadcaster.
- Manual mode submits each created transaction to the selected endpoint. Automatic mode fans out each created transaction to all known endpoints.
- The first successful endpoint wins, remaining endpoint calls get a short grace period, and the endpoint group has a timeout plus drain window before reporting timeout.
- If no endpoint responds before timeout, the result stays pending with timeout-specific copy because the transaction may still have been broadcast.
- First non-gRPC rejection details are preserved, non-timeout gRPC pending results use the default pending copy, partial multi-transaction submissions route to support/error state, and regular sends, ZIP321, shielding, swaps, and Keystone submissions all use the same path.

## Dependency Note

This still depends on the unreleased Android SDK broadcaster API. Local validation uses the SDK included build instead of Maven Central until the released SDK tag is available. Before merge, replace the local SDK dependency path with the released tag and repeat final validation.

## Testing

- End-to-end manual validation covered the app flow with the SDK broadcaster API and the automatic/manual server-mode branch.
- `git diff --check`
- `./gradlew :ui-lib:compileZcashmainnetFossDebugUnitTestKotlin`
- `./gradlew -PJACOCO_VERSION=0.8.14 :ui-lib:testZcashmainnetFossDebugUnitTest --tests co.electriccoin.zcash.ui.common.datasource.MultiEndpointTransactionSubmitterTest --rerun-tasks`
- `./gradlew detektAll`
- `./gradlew ktlint` currently fails locally only on inherited base-branch formatting in `MockSynchronizer.kt`. The PR-scoped files are clean.

Current Android CI failures are inherited from the base branch and expected until the SDK tag and base cleanup land.

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
@@ -10,7 +10,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.async
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
-import java.util.Collections
+import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.atomic.AtomicInteger
 
 internal class MultiEndpointTransactionSubmitter(
@@ -89,7 +89,7 @@ internal class MultiEndpointTransactionSubmitter(
     ): TransactionSubmitResult {
         val completion = CompletableDeferred<BroadcastCompletion>()
         val failureCount = AtomicInteger(0)
-        val failures = Collections.synchronizedList(mutableListOf<TransactionSubmitResult>())
+        val failures = ConcurrentLinkedQueue<TransactionSubmitResult>()
         val jobs =
             endpoints.map { endpoint ->
                 scope
@@ -133,7 +133,7 @@ internal class MultiEndpointTransactionSubmitter(
                                         if (failureCount.incrementAndGet() >= endpoints.size) {
                                             completion.complete(
                                                 BroadcastCompletion.Rejected(
-                                                    result = selectRejectedResult(transaction, failures)
+                                                    result = selectRejectedResult(transaction, failures.toList())
                                                 )
                                             )
                                         }
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
@@ -1,0 +1,295 @@
+package co.electriccoin.zcash.ui.common.datasource
+
+import cash.z.ecc.android.sdk.model.CreatedTransaction
+import cash.z.ecc.android.sdk.model.TransactionSubmitResult
+import co.electriccoin.lightwallet.client.model.LightWalletEndpoint
+import co.electriccoin.zcash.spackle.Twig
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.async
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import java.util.Collections
+import java.util.concurrent.atomic.AtomicInteger
+
+internal class MultiEndpointTransactionSubmitter(
+    private val scope: CoroutineScope,
+    private val globalTimeoutMillis: Long = MULTI_SUBMIT_GLOBAL_TIMEOUT_MILLIS,
+    private val gracePeriodMillis: Long = MULTI_SUBMIT_GRACE_PERIOD_MILLIS,
+    private val submit: suspend (CreatedTransaction, LightWalletEndpoint) -> TransactionSubmitResult
+) {
+    suspend fun submitTransactions(
+        transactions: List<CreatedTransaction>,
+        endpoints: List<LightWalletEndpoint>,
+        logTag: String
+    ): List<TransactionSubmitResult> {
+        var anySubmissionFailed = false
+
+        return transactions.mapIndexed { index, transaction ->
+            if (anySubmissionFailed) {
+                TransactionSubmitResult.NotAttempted(transaction.txId)
+            } else {
+                val result =
+                    submitTransaction(
+                        transaction = transaction,
+                        endpoints = endpoints,
+                        index = index,
+                        transactionCount = transactions.size,
+                        logTag = logTag
+                    )
+
+                if (result !is TransactionSubmitResult.Success) {
+                    anySubmissionFailed = true
+                }
+
+                result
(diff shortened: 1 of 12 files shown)


czarcas7ic:
The one merge blocking thing here is that we still need the [Android SDK](https://github.com/zcash/zcash-android-wallet-sdk/pull/1937) PR merged and tagged and replaced in this PR, but am opening it up for initial review anyway.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
```diff
@@ -0,0 +1,365 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.async
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
+
+internal class MultiEndpointTransactionSubmitter(
```
nesence-m, line 20: You can make the class implement 'CloseableScopeHolder by CloseableScopeHolderImpl()'.

This way the whole class is closeable and will close its own scope when you are done with it.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
```diff
@@ -268,85 +278,69 @@ class ProposalDataSourceImpl(
...
             result
         }
 
+    private suspend fun submitCreatedTransactions(
+        synchronizer: SdkSynchronizer,
+        transactions: List<CreatedTransaction>,
+        endpoints: List<LightWalletEndpoint>,
+        logTag: String
+    ): List<TransactionSubmitResult> =
+        MultiEndpointTransactionSubmitter(
```
nesence-m, line 320: because MultiEndpointTransactionSubmitter is closeable you can invoke 'use' function.

Also remove broadcastScope

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
```diff
@@ -0,0 +1,365 @@
...
+                                submitToEndpoint(
+                                    transaction = transaction,
+                                    endpoint = endpoint,
+                                    endpointLabel = endpointLabel,
+                                    transactionLabel = transactionLabel,
+                                    logTag = logTag
+                                )
+                        )
+                    }.also { job ->
+                        job.invokeOnCompletion { throwable ->
```
nesence-m: i recommend avoiding invokeOnCompletion and launching new jobs within a suspend function.

You can move this logic within scope.launch {} to the async block

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
```diff
@@ -0,0 +1,365 @@
...
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+            }
+        val timeoutJob =
+            scope.launch {
+                delay(globalTimeoutMillis)
```
nesence-m: u can join the delays delay(globalTimeoutMillis + timeoutDrainMillis)

nesence-m (changes requested):
have you published the sdk branch ? I was not able to find it

czarcas7ic:
> have you published the sdk branch ? I was not able to find it

@nesence-m  https://github.com/zcash/zcash-android-wallet-sdk/pull/1937 was merged but someone will still need to tag a release for it. Once that is tagged, I can replace in this PR.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
czarcas7ic: https://github.com/zodl-inc/zodl-android/commit/dfb626d04879a5c177ac96d19b27612b5aac4bfc

Thanks, done here

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
czarcas7ic: https://github.com/zodl-inc/zodl-android/commit/2754db131fff3acffb468da2c28a6bb05d74022e

Addressed here

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/MultiEndpointTransactionSubmitter.kt
czarcas7ic: https://github.com/zodl-inc/zodl-android/commit/89988ad297a8aa9435e1af3f7c0b088daa283b2f

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
czarcas7ic: https://github.com/zodl-inc/zodl-android/commit/dfb626d04879a5c177ac96d19b27612b5aac4bfc

czarcas7ic: One thing I noticed after applying this: wrapping the submitter in `use` closes its scope as soon as `submitTransactions` returns. That cancels the remaining endpoint jobs immediately, so the existing 5s grace period after the first accepted endpoint no longer runs.

I think we still want that grace period so the call can return on first success while other endpoint submissions get a short chance to finish. I’m going to adjust the lifecycle so the submitter is still closeable, but isn’t closed immediately on first return unless you think immediate cancellation is the intended behavior.

czarcas7ic: https://github.com/zodl-inc/zodl-android/pull/2223/commits/544d43bc0ad758b8d18df681ebcff6a92c2b075b

Addressed my above concern here

nesence-m (approved):
utACK

nesence-m:
I've approved the PR. 

There are a few nitpicks & I also think a few things could be done slightly better - but we are not going for 100% perfect code and we can address this iteratively :) 

Functionally the code is correct and handles edge-cases well - this is the most important for any TX logic - the most sensitive stuff of the whole app!

czarcas7ic:
@nesence-m Are you able to leave your comments somewhere so that I can create a follow up PR, so that we could make this closer to what your vision is?

PR #2232, opened 2026-05-17
[#1816] Replace resourceConfigurations with androidResources.localeFilters

### LINKED ISSUE #1816: Gradle `resourceConfigurations` deprecation
The `resourceConfigurations` is deprecated in favor of `androidResources.localeFilters`.

We need to resolve:
```
w: file:///Users/user/Projects/zashi-android/build-conventions-secant/src/main/kotlin/secant.android-build-conventions.gradle.kts:43:13 'resourceConfigurations: MutableSet<String>' is deprecated. Support for resource configurations will be removed. For language resource configurations in applications, use androidResources.localeFilters.
```
and 
```
w: file:///Users/user/Projects/zashi-android/build-conventions-secant/src/main/kotlin/secant.android-build-conventions.gradle.kts:72:13 'resourceConfigurations: MutableSet<String>' is deprecated. Support for resource configurations will be removed. For language resource configurations in applications, use androidResources.localeFilters.
```

## Additional context
Our solution using `resourceConfigurations` is still recommended, though. See https://developer.android.com/build/shrink-code#unused-alt-resources

ChinmayGopal931:
Closes #1816

`resourceConfigurations` is deprecated in AGP 8. This replaces it with `androidResources.localeFilters`.

```diff
-resourceConfigurations.addAll(listOf("en", "en-rUS", "en-rGB", "en-rAU", "es", "en_XA", "ar_XB"))
+localeFilters += listOf("en", "en-rUS", "en-rGB", "en-rAU", "es", "en-rXA", "ar-rXB")
```

### Why two of the tags changed spelling

The pseudolocales went from `en_XA` / `ar_XB` to `en-rXA` / `ar-rXB`. This isn't cosmetic: `localeFilters` validates its entries and `resourceConfigurations` didn't, so the underscore form now fails the build outright.

```
> Task :app:processZcashtestnetFossDebugResources FAILED
  > The locale in localeFilters "en_XA" is invalid.
```

> [!WARNING]
> Please don't "modernize" these to BCP-47 (`en-US`, `en-XA`). AGP hands the list to `aapt2 -c`, which only takes Android resource qualifiers and rejects BCP-47 with `invalid config 'en-US' for -c option`. The `-r` region form is the only spelling that satisfies both. That's also why `en-rUS` / `en-rGB` / `en-rAU` are untouched.

### Why library and test modules just drop the call

`localeFilters` only exists on `ApplicationAndroidResources`, so library and test variants can't set it. They don't need to: the app module's filter applies to everything merged into the APK, including resources from libraries. Library modules here aren't published as AARs and the `com.android.test` modules target `:app`, so nothing is lost.

The `@Suppress("UnstableApiUsage")` matches what `managedDevices` already does in the same file.

### Verification

Built the APK twice from the same tree, once with the old API and once with the new one, then diffed `aapt2 dump configurations` on both. **Identical, 43 configurations each**: `en-rAU`, `en-rGB`, `es` and the `en-rXA` / `ar-rXB` pseudolocales all kept, third party locales still stripped. So this is a like-for-like swap, not just something that compiles.

`:app:assembleZcashtestnetFossDebug` passes, as does `processZcashtestnetFossDebugResources` on `:ui-integration-test` and `:ui-screenshot-test` (the modules the call was removed from).

### DIFF
--- a/build-conventions-secant/src/main/kotlin/secant.android-build-conventions.gradle.kts
+++ b/build-conventions-secant/src/main/kotlin/secant.android-build-conventions.gradle.kts
@@ -13,12 +13,6 @@ pluginManager.withPlugin("com.android.application") {
             minSdk = project.property("ANDROID_MIN_SDK_VERSION").toString().toInt()
             targetSdk = project.property("ANDROID_TARGET_SDK_VERSION").toString().toInt()
 
-            // en_XA and ar_XB are pseudolocales for debugging.
-            // The rest of the locales provides an explicit list of the languages to keep in the
-            // final app.  Doing this will strip out additional locales from libraries like
-            // Google Play Services and Firebase, which add unnecessary bloat.
-            resourceConfigurations.addAll(listOf("en", "en-rUS", "en-rGB", "en-rAU", "es", "en_XA", "ar_XB"))
-
             testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
             testInstrumentationRunnerArguments["useTestStorageService"] = "true"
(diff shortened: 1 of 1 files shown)


nesence-m (dismissed):
utACK

ChinmayGopal931:
> utACK


@nesence-m Is there anything I can do to move this pr along? how can I help get this merged?

nesence-m:
@zodl-review

ChinmayGopal931:
Hey @nesence-m, sorry about the dismissed approval. I pushed a follow-up commit because the version you approved doesn't actually build anymore (pr has been approved and open for a while)

The two pseudolocale tags came straight across as `en_XA` and `ar_XB`. It turns out `localeFilters` checks every entry where `resourceConfigurations` never did, so those two now kill resource linking:

```
> Task :app:processZcashtestnetFossDebugResources FAILED
  > The locale in localeFilters "en_XA" is invalid.
```

I'd only run `compileKotlin` and `help` before, and neither of those links resources, which is why it looked fine.

The fix is to spell them `en-rXA` and `ar-rXB`. I tried the BCP-47 spelling first (`en-XA` and `ar-XB`) and that fails too, just later on, with `invalid config 'en-XA' for -c option`. So the `-r` form is the only one that works here, and that's also why I left `en-rUS`, `en-rGB` and `en-rAU` as they were.

I did a bit of cleanup on the comments while I was in there. The one on the library block said libraries inherit the filter, which isn't really right. The app module's filter just applies to everything merged into the APK, libraries included.


Mind taking another look when you get a chance?

Issue #323, opened 2022-04-04
Localization


ccjernigan:
Covers both addition of foreign language translations, bug fixes for UI layouts that break with multiple languages, and RTL language support.

PR #2233, opened 2026-05-17
[#323] Add content descriptions for Pay and Swap address book and QR scanner buttons

ChinmayGopal931:
Part of #323

Adds missing `contentDescription` values to the address book and QR scanner icon buttons on the Pay (ExactOutput) and Swap screens. These four buttons are icon-only with no adjacent text label, so without a description TalkBack users have no way to know what they do.

Both reuse the existing `send_address_book_content_description` and `send_scan_content_description` strings already defined in `ui-lib`, with Spanish translations in place. No new string resources are added.

The info and swap-direction toggle buttons on these same screens are also unlabeled, but addressing them requires new string resources and translations, so they are deferred to a follow up PR to keep this change small.


nesence-m (approved):
utACK

ChinmayGopal931:
> utACK

@nesence-m Is there anything I can do to move this pr along? how can I help get this merged?

nesence-m (approved):
lgtm. Thanks!

PR #2234, opened 2026-05-17
[#1477] Add showBackground to design-lib previews

### LINKED ISSUE #1477: Add background to design-lib previews
We'd like to leverage the background parameters of the Compose Preview API to eliminate preview customizations using another layer.

For example:
```
@Preview(showBackground = true, backgroundColor = 0xFF231F20)
```

ChinmayGopal931:
Closes #1477

Adds `showBackground = true` to every standalone `@Preview` annotation in `ui-design-lib/src/main/java/co/electriccoin/zcash/ui/design/component/` that did not already have it, so the Android Studio preview panel renders a surface behind each composable instead of a transparent checkerboard. Thirty-seven annotations are updated across thirteen files, all mechanical, with no functional or runtime impact.

The multi-preview meta-annotations in `newcomponent/PreviewScreens.kt` are untouched because their nested `@Preview` entries already declare `showBackground = true`.

I deliberately did not set the `backgroundColor` argument shown in the issue example. The hardcoded hex `0xFF231F20` is the dark-mode brand surface and would render incorrectly behind previews that wrap their content in `ZcashTheme(forceDarkMode = false)`. Leaving `backgroundColor` unset lets the system supply the correct light or dark surface based on the wrapping theme.

### DIFF
--- a/ui-design-lib/src/main/java/co/electriccoin/zcash/ui/design/component/Balance.kt
+++ b/ui-design-lib/src/main/java/co/electriccoin/zcash/ui/design/component/Balance.kt
@@ -135,7 +135,7 @@ object StyledBalanceDefaults {
     )
 }
 
-@Preview
+@Preview(showBackground = true)
 @Composable
 private fun StyledBalancePreview() =
     ZcashTheme(forceDarkMode = false) {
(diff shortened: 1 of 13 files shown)


nesence-m (approved):
lgtm. Thanks!

nesence-m:
outdated - we will use new compose annotation system

PR #2255, opened 2026-05-27
Skip Ktor timeouts for Tor HTTP clients

czarcas7ic:
Skips Ktor timeout capabilities when the shared HTTP client is backed by the SDK Tor engine, which does not support them. Direct OkHttp clients keep the existing client and helper request timeouts, and voting helper requests now gate their per-request timeout configuration the same way.

Validation: `:ui-lib:compileZcashmainnetFossDebugKotlin` and `:ui-lib:testZcashmainnetFossDebugUnitTest --tests co.electriccoin.zcash.ui.common.provider.KtorVotingApiProviderTest` with the pinned SDK checkout.

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/HttpClientProvider.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/HttpClientProvider.kt
@@ -16,6 +16,8 @@ import io.ktor.serialization.kotlinx.json.json
 
 interface HttpClientProvider {
     suspend fun create(): HttpClient
+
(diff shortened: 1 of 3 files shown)


czarcas7ic:
I am not fully confident this is needed, but was recommended as a fix that worked.

Be on tor, on the `feature/chp_vote_ks_signing` branch. Attempt to update the vote config, are unable to with a timeout error of some kind. Update to this branch, can change the vote config.

cc @Cosmos-Harry @noop-sk if you guys come across this issue of being on tor and not being able to update the vote config, I would check this PR out

noop-sk (approved):

PR #2260, opened 2026-05-27
bugfix: No polls bottom sheet instead of empty screen

noop-sk:
## Summary
- When polling is enabled but no rounds exist, replace the inline empty screen with a bottom sheet over a loading shimmer background
- Bottom sheet uses `ZashiConfirmationState.error()` consistent with the rest of the voting flow
- **Refresh** dismisses the sheet and reloads polling data; **Got it** / tap outside navigates back (closes CHP)

## Test plan
- [ ] Enable polling, ensure no active/past rounds → shimmer background + "No polls right now" bottom sheet appears
- [ ] Tap **Refresh** → sheet closes, data reloads
- [ ] Tap **Got it** or tap outside → navigates back out of CHP
- [ ] When rounds exist → normal poll list renders, no sheet shown


LukasKorba (approved):
utACK

PR #2278, opened 2026-06-01
ci: auto-create GitHub Release on version-tag push

y4ssi:
Adds a workflow that creates a GitHub Release automatically when a `<version>-<versionCode>` tag is pushed.

Today the flow is:

1. Engineer drafts the changelog in `docs/whatsNew/WHATS_NEW_EN.md`.
2. Engineer creates a tag.
3. Engineer hand-creates a Release with title `Release X.Y.Z (CODE)` and pastes the matching section as the body.
4. The existing `release.yaml` workflow fires on `release: published` and does the build + APK signing + Solana / F-Droid publishing.

After this PR step (3) is automated:

1-2 unchanged.
3. **Pushing the tag triggers `create-release.yml`**, which:
   - validates tag shape (`[0-9]+.[0-9]+.[0-9]+-[0-9]+`)
   - parses out `<version>` and `<versionCode>`
   - extracts the `## [<version> (<versionCode>)]` section from `docs/whatsNew/WHATS_NEW_EN.md`
   - creates a published Release with title `Release <version> (<versionCode>)`
4. `release.yaml` fires the moment that release is published - exactly the same way it does today.

## What stays manual

- Adding the `## [<version> (<versionCode>)] - <date>` section to `WHATS_NEW_EN.md` before tagging. If the section is missing, the workflow fails loudly before creating any release, so an engineer can fix the changelog and re-push.

## Title / body format

Title: `Release 3.5.1 (1740)` (matches every release since 3.0.1).

Body: verbatim copy of the changelog section. The awk extraction was tested against the real WHATS_NEW_EN.md and produces output identical to the existing 3.4.0 / 3.4.1 release bodies.

## What this does NOT change

- `release.yaml` is untouched. The build / signing / Play / Solana / F-Droid pipeline runs the same way.
- WHATS_NEW_EN.md format is unchanged.
- No new repo secrets required.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,0 +1,81 @@
+name: Create Release from tag
+
+# Fires when a `<version>-<versionCode>` tag is pushed (e.g. `3.5.1-1740`).
+# The workflow:
+#   1. Validates the tag shape.
+#   2. Extracts the matching `## [<version> (<versionCode>)] - <date>` section
+#      from `docs/whatsNew/WHATS_NEW_EN.md`.
+#   3. Creates a published GitHub Release with title `Release <version> (<versionCode>)`
+#      and the changelog section as the body.
+#
+# Publishing the release fires the existing `release.yaml` workflow, which
+# does the actual build + APK signing + Solana / F-Droid publishing. We do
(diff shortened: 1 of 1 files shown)


nesence-m (approved):
utACK.

PR #2297, opened 2026-06-06
[E2E]: parallelize via phase matrix + build APK once

zcash-harry:
## Summary

Restructure the e2e smoke workflow to (a) parallelize the suite across
two matrix jobs and (b) build the APK only once instead of repeating
it per matrix job.

### Phase matrix [1, 2]

Suite splits into two independent halves:
- **Phase 1**: create-new-wallet, navigate-to-receive,
  address-book/delete-contact, currency-conversion
- **Phase 2**: restore-existing-wallet, recovery-phrase,
  export-private-data, send-transaction

Each phase runs on its own emulator in a separate matrix job. Both
zodl-qa wrappers (run-smoke-android.sh, run-smoke-ios.sh) accept
`ZODL_TEST_PHASE=1|2` and select the right `SUITE_FLOWS` subset.

### Build once

Previously each phase job ran `gradle assembleZcashmainnetInternalDebug`
(~5-7 min each). Now a single `build` job uploads the APK as an
artifact; phase matrix jobs `actions/download-artifact` and
`adb install` it.

## Expected impact

| | Before | After |
|---|---|---|
| Wall time | ~40 min | ~17-22 min |
| CI minutes / run | ~40 min | ~25 min |

## Test plan

- [ ] CI green on both phase 1 and phase 2 jobs.
- [ ] APK download + `adb install apk/*.apk` works inside the
      android-emulator-runner script.
- [ ] Each phase's emulator boot doesn't conflict with the other
      (separate runners — should be fine).

## Companion PR

zodl-ios `harry/e2e-biometric-toggle` carries the same restructure
for the iOS workflow.

### DIFF
--- a/.github/workflows/e2e-smoke.yml
+++ b/.github/workflows/e2e-smoke.yml
@@ -36,11 +36,17 @@ jobs:
       - name: Restore cached APK
         # Keyed on hashes of every input that affects the APK output.
         # On hit, skip the gradle build and just upload the restored APK.
+        # buildSrc/src/** + build-conventions-secant/src/** hold Kotlin
+        # source (Git.kt, ChangelogParser.kt, Dimensions.kt, BuildType.kt)
+        # that affects the build but isn't matched by the
+        # **/*.gradle.kts glob. Without those entries a change there
+        # would leave the cache valid and the workflow could install a
+        # stale APK.
         id: apk-cache
         uses: actions/cache@v5
         with:
           path: app/build/outputs/apk/zcashmainnetInternal/debug
-          key: ${{ runner.os }}-android-apk-${{ hashFiles('app/src/**', 'ui-lib/src/**', 'sdk-ext-lib/src/**', 'ui-design-lib/src/**', '**/*.gradle.kts', 'gradle.properties', 'gradle/libs.versions.toml') }}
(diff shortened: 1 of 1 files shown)


noop-sk:
## Review

Overall a clean and well-motivated CI optimization. One issue worth fixing before merge.

### Issue: cache key misses `buildSrc/src/**/*.kt`

The repo has `buildSrc/src/main/kotlin/co/electriccoin/zcash/Git.kt`. The current hash covers `**/*.gradle.kts` but not `.kt` source files inside `buildSrc/src/`. A change to those files won't invalidate the APK cache, so the workflow could install a stale APK.

**Fix** — add `buildSrc/src/**` to the `hashFiles` call:

\`\`\`yaml
key: ${{ runner.os }}-android-apk-${{ hashFiles('app/src/**', 'ui-lib/src/**', 'sdk-ext-lib/src/**', 'ui-design-lib/src/**', 'buildSrc/src/**', '**/*.gradle.kts', 'gradle.properties', 'gradle/libs.versions.toml') }}
\`\`\`

### Everything else looks good

- APK upload is unconditional (runs on both cache hit and miss) — correct.
- No app `checkout` in phase jobs — clean, APK arrives via `download-artifact`.
- `fail-fast: false` — correct, a flaky phase shouldn't cancel the other.
- Artifact name suffixed with `-phase-${{ matrix.phase }}` — avoids upload conflict.
- `retention-days: 1` on the APK artifact — minimal and appropriate.

noop-sk (approved):

PR #2295, opened 2026-06-05
Release 3.5.3

noop-sk:
Closes #2294

### Changed:
- Removed servers scheduled for decommissioning from the server list.

PR #2300, opened 2026-06-08
ci(create-release): fire release workflow for prerelease-branch tags

y4ssi:
## Problem

Tag `3.5.3-1745` was pushed (merge of #2295) but **no GitHub Release was created**, so the build+publish chain never ran — there is no release at `releases/tag/3.5.3-1745` and `release.yaml` (which builds + signs + publishes to Play / F-Droid) never fired.

### Root cause

The release chain is two links:

```
push tag 3.5.3-1745
   └─► create-release.yml   (tag → published GitHub Release)
          └─► release.yaml   (on: release published → build + sign + publish)
```

For `push: tags` events, GitHub runs the workflow file **as it exists in the tagged commit**, not on `main`. `create-release.yml` was added to `main` on 2026-06-01 (`a673060a`). The tag `3.5.3-1745` lives on `prerelease/3.5.3-point`, which **forked from `main` before that commit and was never reintegrated** (`git compare` reports `diverged`, 18 commits behind). So `create-release.yml` is simply **not present in the tagged tree** — `GET contents/.github/workflows/create-release.yml?ref=<tag>` returns 404 — and the push trigger has nothing to run.

This is why prior releases (`3.5.1`, `3.5.2`) were all created **manually**; the single real run of `create-release.yml` (for `3.5.1`) failed. The auto-release has never actually worked end-to-end.

## Fix

1. **`workflow_dispatch` escape hatch** — add a `tag` input so the workflow can be run **from `main`** (where the file always lives) against any existing tag, regardless of whether the tagged commit carries the workflow. This covers every prerelease branch that forked before the workflow existed:
   ```
   gh workflow run create-release.yml -f tag=3.5.3-1745
   ```
2. **Checkout the tag ref** so the changelog section is read from the tagged tree (not from whatever the dispatch branch happens to be).
3. **Document + mitigate the cascade caveat** — releases created with the default `GITHUB_TOKEN` do **not** trigger other workflows (GitHub anti-recursion), so `release.yaml` won't auto-fire. The workflow now prefers an optional `RELEASE_PAT` secret (fine-grained PAT / App token with `contents: write`) and falls back to `GITHUB_TOKEN`.

## Immediate remediation (already done)

I manually created the `3.5.3-1745` release, which kicked off the `release.yaml` build (run 27147687255). This PR makes sure the next prerelease tag doesn't need that.

## Follow-ups (not in this PR)

- Add a `RELEASE_PAT` repo secret so the dispatch/auto path cascades into `release.yaml` without a manual re-publish.
- Consider merging `main` into `prerelease/*` branches (or branching them from `main`) before tagging so the `push:tags` path also works.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -33,10 +33,14 @@ name: Create Release from tag
 # CASCADE CAVEAT: releases created with the default `GITHUB_TOKEN` do NOT
 # trigger other workflows (GitHub's anti-recursion rule), so `release.yaml`
 # (on: release published) will NOT auto-fire from this workflow's release.
-# To make the build+publish chain run automatically, set a `RELEASE_PAT`
-# repo secret (a fine-grained PAT or GitHub App token with `contents: write`)
-# and this workflow will use it; otherwise it falls back to GITHUB_TOKEN and
-# an operator must (re)publish the release manually to kick off release.yaml.
+# To make the build+publish chain run automatically, this workflow mints a
+# short-lived GitHub App installation token (preferred over a long-lived PAT
+# in a PUBLIC repo) and creates the release with it. Wiring:
+#   * repo variable  RELEASE_BOT_APP_ID    — the App's numeric id (not secret)
+#   * repo secret     RELEASE_BOT_APP_KEY  — the App's PEM private key
+# The App needs only `contents: write` on this single repo. If the variable
+# is unset, the workflow falls back to GITHUB_TOKEN and an operator must
+# (re)publish the release manually to kick off release.yaml.
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


noop-sk (dismissed):

.github/workflows/create-release.yml
```diff
@@ -69,13 +122,26 @@ jobs:
             echo "__BODY_EOF__"
           } >> "$GITHUB_OUTPUT"
 
+      # Mint a short-lived App installation token so the published release
+      # cascades into release.yaml. Skipped when RELEASE_BOT_APP_ID is unset;
+      # the create step then falls back to GITHUB_TOKEN (no cascade).
+      - name: Mint App token
+        id: app_token
+        if: ${{ vars.RELEASE_BOT_APP_ID != '' }}
+        uses: actions/create-github-app-token@v2
```
noop-sk: Consider SHA-pinning this action to match the `actions/checkout` convention a few lines above (`@de0fac2e… # v6.0.2`).

This is the most security-sensitive step in the file — it reads the App private key and mints a token with `contents: write`. A floating `@v2` tag can be re-pointed (compromised maintainer, force-pushed tag) and silently swap the code that handles those credentials.

Suggested:

```yaml
uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2.2.2
```

(v2 currently resolves to v2.2.2 / `fee1f7d`. v3.2.0 also exists if you prefer to jump majors — `bcd2ba49218906704ab6c1aa796996da409d3eb1`.)

noop-sk (approved):

PR #2314, opened 2026-06-11
[#1378] Fetch exchange rates over Tor only

nesence-m:
Exchange-rate fetching was not gated on Tor state: CMCApiProvider used HttpClientProvider.create(), which silently falls back to a direct (clearnet) client when Tor is disabled, exposing the user's IP and request timing to the rate provider — while the UI promises rates are fetched over Tor.

- HttpClientProvider: add createTor() to the interface, returning a client that always routes over Tor regardless of the user's toggle.
- CMCApiProvider: use createTor() instead of create() so the CMC quote request can never traverse clearnet.
- HttpClientProvider.createDirect(): refuse requests to the CMC host at the network layer (defense-in-depth), mirroring the existing isVotingHelperPath() special-casing, and skip retries for it.
- Add CMCApiProviderTest asserting CMC uses the Tor-only client and never create().

> **Note**

# Author

> **Note**

# Reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ### Fixed:
 - Network request/response bodies and voting diagnostics are no longer written to logs in release builds, preventing sensitive data (recipient/refund addresses, amounts, transaction hashes) from leaking to logcat, bug reports, and crash dumps. Credential headers are also redacted from logs.
+- Exchange rates are no longer requested over a direct (non-Tor) connection. When Tor Protection is disabled the request to the rate provider is now blocked instead of falling back to clearnet, preventing the user's IP address and request timing from being exposed.
 
 ## [3.5.3 (1745)] - 2026-06-05
 
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/CMCApiProvider.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/CMCApiProvider.kt
@@ -22,7 +22,7 @@ class CMCApiProviderImpl(
 ) : CMCApiProvider {
     override suspend fun getExchangeRateQuote(apiKey: String): GetCMCQuoteResponse =
         execute {
-            get("https://pro-api.coinmarketcap.com/v1/cryptocurrency/quotes/latest") {
+            get("https://$CMC_API_HOST/v1/cryptocurrency/quotes/latest") {
                 parameter("symbol", "ZEC")
                 parameter("convert", "USD")
                 contentType(ContentType.Application.Json)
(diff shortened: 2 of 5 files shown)


noop-sk (commented):
## Review

**Overview:** Fixes a privacy bug where CMC exchange-rate requests could traverse clearnet when Tor was disabled. The fix has two layers: CMCApiProvider now calls `createTor()` (always Tor), and `createDirect()` gets an OkHttp interceptor that throws `IOException` for any CMC request as defense-in-depth.

---

### Correctness

- **Core fix is correct.** `createTor()` promoted to the interface and used in `CMCApiProvider` — the clearnet path can no longer be taken from that call site.

- **Exception contract mismatch.** `CMCApiProvider.getExchangeRateQuote` is annotated `@Throws(ResponseException::class)`, but `createTor()` can now throw `IOException` (or whatever `getTorHttpClient` throws when Tor is off/failing). The annotation is now misleading. In practice it's safe because `ExchangeRateRepository` has `catch (_: Exception)` which swallows any non-`ExchangeRateUnavailable` exception and silently fails — but this is implicit and fragile. Consider either widening the annotation to `@Throws(IOException::class, ResponseException::class)` or wrapping non-`ResponseException` exceptions in `ExchangeRateDataSource.getCMCExchangeRate()`.

### Code Quality

- **Issue-number references in code comments.** Comments in `CMCApiProvider.kt` and `HttpClientProvider.kt` include `// MOB-1378:` prefixes. Comments should explain the *why* (constraint/invariant) without referencing tickets — tickets rot and comments that reference ticket numbers lose meaning over time. The explanatory text that follows the prefix is valuable; strip the prefix.

- **KDoc on `createTor()` interface method** also references `MOB-1378` — same concern.

- **`CMC_API_HOST` constant is private and duplicated.** `HttpClientProvider.kt` has `private const val CMC_API_HOST` and `CMCApiProviderTest.kt` declares its own `private const val CMC_HOST` with the same value. Not a bug, but if the host ever changes it's two places.

### Defense-in-Depth

The `IOException`-throwing interceptor in `createDirect()` + the retry exclusion for exchange-rate paths is a solid safety net. Appreciated.

### Tests

- `CMCApiProviderTest` is well-structured and covers the critical invariant.
- `RecordingHttpClientProvider.create()` calls `error()` — the test will throw `IllegalStateException` (not `AssertionError`) if `create()` is called, meaning the `assertEquals(0, createCalls, ...)` assertion below it would never run. Consider relying on either the counter assertion or `error()`, not both.
- `KtorVotingApiProviderTest` fake correctly delegates `createTor()` to `create()`.

---

Approve with minor nits. The security fix is correct and the layered approach is the right call. The `@Throws` contract mismatch is the only thing worth addressing before merge; the rest are style points.

noop-sk (dismissed):

noop-sk (approved):

PR #2318, opened 2026-06-13
[MOB-1339] Treat "already in mempool" submit rejections as success

zcash-harry:
## Summary
- A TEX send (or any send that's rebroadcast after the original submit already landed in the network) currently surfaces a misleading "transaction failed" screen.
- Root cause: lightwalletd / zcashd returns `transaction already exists in mempool` or `transaction dropped because it is already queued for download` as a non-zero submit error, which the SDK reports as `TransactionSubmitResult.Failure`. Those descriptions actually mean the tx is already on the network — i.e. the broadcast succeeded.
- Fix: `ProposalDataSourceImpl.submitTransactionInternal` now classifies a `Failure` with one of those two descriptions as effective success — counts toward `successCount`, excluded from the resubmittable-failure set, status string rewritten so debug logs still surface the underlying description.
- This is the app-layer guard. The matching SDK-level fix lives at [Cosmos-Harry/zcash-android-wallet-sdk → harry/treat-already-in-mempool-as-success](https://github.com/Cosmos-Harry/zcash-android-wallet-sdk/pull/new/harry/treat-already-in-mempool-as-success) and becomes redundant (harmless) once that lands in our dependency pin.

Linear: MOB-1339

## Test plan
- [ ] Send to a TEX address from a v3.5.1+ wallet where the previous attempt is known to have broadcast successfully (e.g. retried after a confirmation timeout).
- [ ] Confirm the success screen renders instead of the failure screen.
- [ ] Confirm transaction list still shows the tx in the expected state.
- [ ] No regression for genuine failures: send with insufficient funds → failure screen still appears with the real error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 - The cross-chain swap slippage guarantee is now always enforced: the swap provider's minimum-amount bounds are treated as required, so a malicious or tampered response can no longer skip the slippage check by omitting them.
 - Swap quotes rejected by the amount-consistency safety check are now reported to monitoring (without any amounts), so a future swap-provider format change surfaces as an observable signal instead of silently blocking swaps.
 - Exchange rates are no longer requested over a direct (non-Tor) connection. When Tor Protection is disabled the request to the rate provider is now blocked instead of falling back to clearnet, preventing the user's IP address and request timing from being exposed.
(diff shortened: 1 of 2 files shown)


nesence-m (approved):
utACK. Lgtm

nesence-m:
@Cosmos-Harry this PR is not needed with the newest SDK PR is it?

zcash-harry:
> @Cosmos-Harry this PR is not needed with the newest SDK PR is it?

Yea, this was a hot fix to have the change asap but SDK is the main one.

PR #2323, opened 2026-06-16
Release 3.6.0

noop-sk:
### Added:
- Until now, balances only displayed in US dollars. We added multi-currency support — pick your preferred fiat in Currency Conversion settings and on the Wallet Status Widget opt-in.

### Changed:
- New Automatic/Manual server selection improves reliability — Automatic keeps you on the best server. Users on custom servers stay on Manual.
- New wallets now sync from the chain tip instead of a bundled checkpoint, skipping the first-launch scan.

### Changed:
- Removed servers scheduled for decommissioning from the server list.

## [3.5.2 (1742)] - 2026-06-04


nesence-m (approved):
utACK

PR #2324, opened 2026-06-16
ci(release): publish F-Droid via GitLab Pages trigger

y4ssi:
The `publish_to_fdroid` job built the F-Droid index locally and pushed it to `zodl-inc/fdroid` for **GitHub Pages**. That never worked: the FOSS APK is ~181 MB, over GitHub Pages' 100 MB/file hard limit, so the repo could never actually serve the APK.

The F-Droid repo now lives on **GitLab Pages** (project `zodl-inc/fdroid` on gitlab.com, id `83433564`). A GitLab CI `pages` job downloads this release's FOSS APK from the GitHub release, builds + signs the index (index key held as a GitLab CI variable, fingerprint `7e751ab7…` unchanged), and publishes — GitLab Pages serves the large APK fine.

## What changed
- `publish_to_fdroid` is now a **single step** that triggers the GitLab pipeline with `variables[APK_TAG]=${GITHUB_REF_NAME}` via a scoped pipeline **trigger token** (new repo secret `GITLAB_FDROID_TRIGGER_TOKEN`).
- Dropped: AWS OIDC + Secrets Manager keystore load, `fdroidserver` install, fdroid index build, SSH clone + git-push to `zodl-inc/fdroid`, and the unused `environment: deployment`.
- `needs: [release]` kept (runs after the APK is on the release).
- **Best-effort**: a non-2xx trigger response logs a warning but does not fail the release.

## Out of band (already done)
- Trigger token created on the GitLab project and stored as repo secret `GITLAB_FDROID_TRIGGER_TOKEN`.
- `zodl-inc/fdroid` `.gitlab-ci.yml` updated to build the APK download URL from `APK_TAG` (defaults to the latest known tag).
- Verified end-to-end with a test trigger of `APK_TAG=3.5.3-1745` (idempotent re-publish).

Do not squash-merge away the trigger best-effort handling.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -322,144 +322,40 @@ jobs:
           echo "Config updated in GCS"
           echo "Backup: gs://${{ secrets.SOLANA_GCS_BUCKET }}/backups/encrypted_config.yaml.kms.${TIMESTAMP}"
 
+  # Publishes the FOSS APK to the self-hosted F-Droid repo at foss.zodl.com.
+  #
+  # The repo is served by GitLab Pages (project zodl-inc/fdroid on gitlab.com).
+  # A GitLab CI "pages" job downloads this release's FOSS APK from the GitHub
+  # release, builds + signs the F-Droid index (with the long-lived index key
+  # held as a GitLab CI variable), and publishes via Pages -- which, unlike
+  # GitHub Pages (100 MB/file hard limit), can serve the ~181 MB APK.
+  #
+  # This job's only responsibility is to TRIGGER that GitLab pipeline with the
+  # release tag, so the right APK version gets published. All the AWS Secrets
+  # Manager / keystore / fdroidserver / git-push work now lives in GitLab CI.
+  #
+  # Best-effort: a failed trigger must not fail the release.
   publish_to_fdroid:
     needs: [release]
-    environment: deployment
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      id-token: write  # Required for OIDC -> AWS
     steps:
-      - name: Get release tag
-        id: release_info
+      - name: Trigger GitLab Pages F-Droid publish
         env:
-          GH_TOKEN: ${{ github.token }}
+          GITLAB_FDROID_TRIGGER_TOKEN: ${{ secrets.GITLAB_FDROID_TRIGGER_TOKEN }}
         run: |
-          TAG="${GITHUB_REF_NAME}"
-          VERSION_NAME=$(echo "$TAG" | sed 's/^v//' | cut -d'-' -f1)
-          VERSION_CODE=$(echo "$TAG" | cut -d'-' -f2)
-
-          echo "tag=${TAG}" >> $GITHUB_OUTPUT
-          echo "version_name=${VERSION_NAME}" >> $GITHUB_OUTPUT
-          echo "version_code=${VERSION_CODE}" >> $GITHUB_OUTPUT
-
-      - name: Configure AWS credentials (OIDC)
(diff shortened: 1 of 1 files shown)


nesence-m (dismissed):
utACK

.github/workflows/release.yaml
```diff
@@ -322,144 +322,40 @@ jobs:
...
-        working-directory: fdroid-repo
-        env:
-          TAG: ${{ steps.release_info.outputs.tag }}
-        run: |
-          git add fdroid/repo fdroid/archive fdroid/metadata || true
-          if git diff --cached --quiet; then
-            echo "No changes to publish."
-            exit 0
+          echo "Triggering GitLab F-Droid pipeline for ${GITHUB_REF_NAME}"
+          HTTP_CODE=$(curl -s -o /tmp/trigger.json -w '%{http_code}' \
```
noop-sk: Missing `-m` timeout — if GitLab is unreachable, this `curl` will block until the runner's job-level timeout fires (typically 6 h). Add `-m 30` to bound the wait:
```bash
HTTP_CODE=$(curl -s -m 30 -o /tmp/trigger.json -w '%{http_code}' \
```

.github/workflows/release.yaml
```diff
@@ -322,144 +322,40 @@ jobs:
...
+            -X POST "https://gitlab.com/api/v4/projects/83433564/trigger/pipeline" \
```
noop-sk: Hardcoded GitLab project ID — functional, but if the project is ever migrated/recreated this silently breaks. Consider promoting to a repo variable (`GITLAB_FDROID_PROJECT_ID`) alongside `GITLAB_FDROID_TRIGGER_TOKEN` so it can be updated in one place.

.github/workflows/release.yaml
```diff
@@ -322,144 +322,40 @@ jobs:
...
+            -F token="${GITLAB_FDROID_TRIGGER_TOKEN}" \
+            -F ref=main \
```
noop-sk: `ref=main` is an implicit contract with the GitLab repo. If the GitLab pipeline ever moves to an environment branch (e.g. `production`) this will silently trigger the wrong ref. A repo variable would make it explicit and easy to change.

.github/workflows/release.yaml
```diff
@@ -322,144 +322,40 @@ jobs:
...
+  # GitHub Pages (100 MB/file hard limit), can serve the ~181 MB APK.
+  #
+  # This job's only responsibility is to TRIGGER that GitLab pipeline with the
+  # release tag, so the right APK version gets published. All the AWS Secrets
+  # Manager / keystore / fdroidserver / git-push work now lives in GitLab CI.
+  #
+  # Best-effort: a failed trigger must not fail the release.
   publish_to_fdroid:
     needs: [release]
-    environment: deployment
```
noop-sk, line 327: Removing `environment: deployment` drops any required-reviewer gates that were set on this GitHub environment. Confirm the environment had no mandatory approvals — if it did, those protections are now gone for the trigger step.

y4ssi:
Thanks @noop-sk — addressed all four:

1. **curl timeout** — added `-m 30` so an unreachable GitLab can't block the runner until the 6h job timeout. ✅
2. **Project id** — now read from repo var `GITLAB_FDROID_PROJECT_ID` (set to 83433564), defaulting to the literal if unset. ✅
3. **ref** — now `GITLAB_FDROID_REF` (set to `main`), same default pattern. ✅
4. **dropped `environment: deployment`** — confirmed via the environments API it only had a `branch_policy` (no required reviewers), so no approval gate was lost. ✅

Pushed in the latest commit.

nesence-m (approved):
utACK

PR #2327, opened 2026-06-18
ci: fix publish_to_solana KMS key + migrate to dapp-store-cli 1.0.0

y4ssi:
## Why

The `publish_to_solana` job in `.github/workflows/release.yaml` has **never succeeded**. Two confirmed root causes, fixed here.

## 1. Wrong KMS key (decrypt + re-encrypt of config.yaml)

The config.yaml **decrypt** step and the config.yaml **re-encrypt** step both used `--key=config`. That CryptoKey does **not exist** — the GCP KMS keyring `gpg` (project from `secrets.GCP_PROJECT_ID_PROD`) has exactly one key: `gpg` (directly verified). The keypair decrypt step already (correctly) used `--key=gpg`.

Both `--key=config` lines are now `--key=gpg`. The keypair decrypt step is left untouched.

## 2. Obsolete Solana dApp Store CLI (0.15.x → 1.0.0)

The job called the legacy interface:

```
npx @solana-mobile/dapp-store-cli@latest create release -k <keypair> ... --config config.yaml
npx @solana-mobile/dapp-store-cli@latest publish update  -k <keypair> ... --config config.yaml
```

Solana Mobile shipped `dapp-store-cli@1.0.0`, a portal-backed CLI that **removed** `create release`, `publish update`, `-k`, and `--config`. The old `0.15.0` is now hard-gated server-side and refuses to run, so neither `@latest` nor the pinned old version works.

Migrated to the single 1.0.0 invocation against the already-downloaded FOSS APK (`stores/solana/solana.apk`):

```
npx @solana-mobile/dapp-store-cli@1.0.0 \
  --apk-file solana.apk \
  --keypair /tmp/solana-keypair.json \
  --whats-new "New features and improvements in this release."
```

with the portal key supplied via env `DAPP_STORE_API_KEY`. A comment block documents the 0.15→1.0 migration and the API-key prerequisite.

The config.yaml metadata flow (decrypt/re-encrypt from GCS) is largely vestigial in the 1.0.0 portal model (the portal holds the metadata now), but it is **left in place** — the KMS key fix is correct and harmless, and the keypair decrypt is still required to sign the on-chain submission. Stripping the scaffolding can be a follow-up.

`continue-on-error: true` is kept on the job (it is best-effort).

## ⚠️ HUMAN PREREQUISITE — `DAPP_STORE_API_KEY` does not exist yet

**The `DAPP_STORE_API_KEY` repo secret does NOT exist.** The 1.0.0 CLI requires it, so this job will fail at the publish step until a human creates it:

1. Generate it at **publish.solanamobile.com → settings → API keys**
2. `gh secret set DAPP_STORE_API_KEY`

This is expected — the job is best-effort (`continue-on-error: true`), so its absence does not fail the release. But Solana publishing will not actually work until the secret is set.

## Validation

`python3 -c "import yaml; yaml.safe_load(...)"` — valid YAML.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -234,7 +234,7 @@ jobs:
 
           echo "Decrypting config.yaml..."
           gcloud kms decrypt \
-            --key=config \
+            --key=gpg \
             --keyring=gpg \
             --location=global \
             --plaintext-file=/tmp/config.yaml \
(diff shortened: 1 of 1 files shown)


nesence-m (approved):
utACK

Issue #1815, opened 2025-03-10
Gradle `targetSdk` deprecated


HonzaR:
## Is your feature request related to a problem? Please describe.
'targetSdk: Int?' is deprecated. Will be removed from library DSL in v9.0. Use testOptions. targetSdk or/ and lint. targetSdk instead

## Describe the solution you'd like
This is deprecated but we don't have a replacement for the instrumentation APKs yet

## Additional context
https://developer.android.com/guide/topics/manifest/uses-sdk-element.html

PR #2330, opened 2026-06-20
Resolve build-conventions AGP DSL deprecations (#1816, #1815)

ChinmayGopal931:
Both of these were sitting in the Android build conventions as TODOs, and the AGP version we're on now (8.13.2) has the replacements, so I cleared them out.

**#1816 — `resourceConfigurations` → `androidResources.localeFilters`**

`resourceConfigurations` is deprecated in favor of `androidResources.localeFilters`. The wrinkle is that `localeFilters` only exists on the application module (it's declared on `ApplicationAndroidResources`, not the library or test variants), which lines up with the deprecation message about filtering locales "in applications". So the locale list moves to the app module.

The library and test modules have no equivalent, and their copy only ever trimmed locales out of the instrumentation and benchmark test APKs, so it's dropped there. Nothing the shipped app needs can disappear from that, since removing a filter only keeps more locales.

**#1815 — library `targetSdk` → `testOptions.targetSdk`**

The library module's `defaultConfig.targetSdk` is deprecated and gets removed from the library DSL in AGP 9. For a library it only sets the target for the instrumentation test APK, so it moves to `testOptions.targetSdk`. The old "no replacement yet" comment is stale on the current AGP. The app and test modules keep `defaultConfig.targetSdk` since it's only deprecated for libraries.

The locale values are carried over exactly as they were, so this is a rename of the same filtering rather than a behavior change.

Verified the convention plugins compile against AGP 8.13.2 and that the app module configures with the new `localeFilters`.

### DIFF
--- a/build-conventions-secant/src/main/kotlin/secant.android-build-conventions.gradle.kts
+++ b/build-conventions-secant/src/main/kotlin/secant.android-build-conventions.gradle.kts
@@ -13,12 +13,6 @@ pluginManager.withPlugin("com.android.application") {
             minSdk = project.property("ANDROID_MIN_SDK_VERSION").toString().toInt()
             targetSdk = project.property("ANDROID_TARGET_SDK_VERSION").toString().toInt()
 
-            // en_XA and ar_XB are pseudolocales for debugging.
-            // The rest of the locales provides an explicit list of the languages to keep in the
-            // final app.  Doing this will strip out additional locales from libraries like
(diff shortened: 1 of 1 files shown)


ChinmayGopal931:
Closing as a duplicate. #2232 already migrates resourceConfigurations to androidResources.localeFilters (#1816) with the same approach and is approved and mergeable, so this is redundant. I'll handle the targetSdk deprecation (#1815) as its own small PR rather than bundling it here.

PR #2337, opened 2026-06-23
Track Linear releases on release:published (run from main)

zcash-harry:
## What
Move Linear release tracking out of `create-release.yml` into a standalone **`linear-release.yml`** triggered on **`release: published`**. Mirrors the iOS fix (zodl-ios#1855) and replaces the reverted glob change (#2335 → #2336).

## Why
The Linear step was a trailing step of `create-release.yml`, whose automatic `push: tags` path uses the workflow **as it exists on the tagged commit** — so it silently missed tags cut from prerelease branches that predate the file (the limitation already documented in that file's header). "From main" was only reachable via the manual `workflow_dispatch`.

`release: published` runs from **`main`** and fires on the same event `release.yaml` already consumes (emitted by `create-release.yml`'s App-token cascade), so Linear tracking now runs automatically from `main` on every published release — no dependency on the tagged commit, no manual step.

## Changes
- **New** `linear-release.yml`: `on: release: { types: [published] }`, `permissions: contents: read`, checkout pinned to `github.event.release.tag_name` with `fetch-depth: 0`.
- **`create-release.yml`**: remove the embedded Linear step; drop the now-unused `fetch-depth: 0` (only that step needed it — changelog extraction reads the tree, not history); update header notes.

## Review hardening
- `concurrency` guard keyed on the tag — `create-release.yml` emits `published` up to twice (create-as-published + the force-publish toggle), so runs are serialized to avoid overlapping Linear syncs.
- SHA-pinned `linear/linear-release-action` (was floating `@v0`) to match the repo's pin-every-action convention.
- Header qualified: the auto-cascade only happens on the App-token path (`RELEASE_BOT_APP_ID` is configured); the default `GITHUB_TOKEN` doesn't trigger downstream workflows.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -11,8 +11,12 @@ name: Create Release from tag
 #      and the changelog section as the body.
 #
 # Publishing the release fires the existing `release.yaml` workflow, which
-# does the actual build + APK signing + Solana / F-Droid publishing. We do
+# does the actual build + APK signing + Solana / F-Droid publishing, and the
+# `linear-release.yml` workflow, which records the release in Linear. We do
 # NOT do any of that here — this workflow's only job is "tag -> release".
+# (Linear tracking used to be a trailing step here; it now lives in
+# `linear-release.yml` on `release: published` so it runs from `main` rather
+# than depending on this file being present on the tagged commit.)
 #
 # Two trigger paths:
 #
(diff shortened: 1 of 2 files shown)


zcash-harry:
@zodl-review please review if now linear release will work

noop-sk (approved):

PR #2341, opened 2026-06-24
MOB-1396 Swap code cleanup and test coverage

nesence-m:
@noop-sk this is the first PR before MOB-1396 implementation. This PR separation will help to review subsequent one later.

> **Note**

# Author

> **Note**

# Reviewer


nesence-m:
@zodl-review

nesence-m:
@zodl-review

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetPreselectedSwapAssetUseCase.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val swapRepository: SwapRepository,
+    private val metadataRepository: MetadataRepository,
+    private val simpleSwapAssetProvider: SimpleSwapAssetProvider
+) {
+    /**
+     * Resolves the asset that should be preselected — the most recently used asset from history,
+     * falling back to a hardcoded default — and suspends until the loaded asset list actually
+     * contains it, then returns that [SwapAsset]. The caller (ViewModel) owns the selection state
+     * and decides whether to apply this result.
+     */
```
noop-sk, line 23: **Suspend-forever risk**: if `SwapRepository.assets` never emits a non-empty list (e.g. network down on first launch), this call never returns and the VM's `init` coroutine hangs permanently — producing a blank UI with no error shown. Consider wrapping the call with `withTimeoutOrNull` in the caller, or letting the VM fall back to a default on timeout.

The same applies to `getAssetFromHistory()` below: `filterNotNull().first()` also suspends forever if the metadata flow never emits non-null. In practice MetadataRepository emits `null` only transiently before the first DB read, so this is lower risk — but worth noting in case that assumption ever breaks.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/picker/SwapAssetPickerScreen.kt
```diff
@@ -6,6 +6,7 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import kotlinx.serialization.Serializable
 import org.koin.androidx.compose.koinViewModel
 import org.koin.core.parameter.parametersOf
+import java.util.UUID
 
 @Composable
 fun SwapAssetPickerScreen(args: SwapAssetPickerArgs) {
```
noop-sk, line 12: Default parameter expressions are evaluated at every call site in Kotlin, so each `SwapAssetPickerArgs()` construction gets a fresh UUID — which is the correct behaviour for the pipeline (each navigation request needs a unique ID). Worth noting that `copy()` preserves the existing `requestId`, which is also intentional (the pipeline must match results back to the original request, not a new one).

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/slippage/SwapSlippageScreen.kt
```diff
@@ -17,6 +18,8 @@ fun SwapSlippageScreen(args: SwapSlippageArgs) {
```
noop-sk, line 16: Same note as `SwapAssetPickerArgs.requestId`: evaluated at each call site (fresh UUID per instance), `copy()` intentionally preserves the existing ID for pipeline result matching.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/SwapVM.kt
```diff
@@ -9,38 +9,36 @@ import co.electriccoin.zcash.ui.R
...
-import co.electriccoin.zcash.ui.screen.swap.slippage.SwapSlippageArgs
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.WhileSubscribed
-import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.update
```
noop-sk, line 44: Reading from `swapRepository.assets.value.data` directly here rather than `internalState.value.swapAssets` is intentional: `internalState.swapAssets` is only updated by the `combine` above on the next emission, so it can lag behind. The repository snapshot is always current at this synchronous call point.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetPreselectedSwapAssetUseCase.kt
nesence-m: Good call-out — keeping the current behaviour for this PR. A few mitigating points so we can proceed:

- `invoke()` is awaited inside `viewModelScope.launch { }` in the VM `init`, so the suspended coroutine is tied to the VM lifecycle and cancelled on `onCleared()` — it is not a process-lifetime leak, just a pending preselect for the screen's lifetime.
- The preselect only seeds the *initial* selected asset. The asset list and its loading/error are driven independently by `getSwapAssetsUseCase.observe()` (`SwapAssetsData.isLoading` / `.error`), so a slow or failed first load is represented in state rather than gated solely on the preselect resolving.
- `SwapRepository.assets` is refreshed on a retrying loop, so the preselect resolves once assets load.

So no functional regression here — good to proceed. If we want belt-and-suspenders later, a `withTimeoutOrNull` fallback in the caller is a small follow-up, but it isn't needed for this change.

noop-sk (approved):

PR #2347, opened 2026-06-26
ci(release): add workflow_dispatch recovery lever for alt-store publish

y4ssi:
## What happened

Tag `3.7.0-1986` was cut from the long-lived `prerelease/3.7.0` branch (commit `b8879fa`, merged via #2339). That branch **diverged from `main` and never carried the cascade fix** from #2328 (`ci: force release published event so release.yaml always cascades`).

Tag-push workflows run from the **tagged tree**, so the *old* `create-release.yml` ran for `3.7.0-1986`. It lacks the **"Force published event"** step, so:
- the GitHub Release (created out-of-band on 06-23) was only **updated**, not created-as-published,
- updating a release does **not** emit `release: published`,
- so `release.yaml` (Google Play + Solana + F-Droid publish) **never fired** — nothing showed up in Actions.

There was no direct way to trigger `release.yaml` (it only had `release: published`), so recovery meant re-dispatching `create-release.yml` from `main` to re-emit the published event.

## Fix

Add a `workflow_dispatch(tag)` recovery lever to `release.yaml` so an operator can publish the stores directly for any existing tag:

```
gh workflow run release.yaml -f tag=3.7.0-1986
```

Both checkouts and the GitLab F-Droid trigger now resolve the tag from `inputs.tag || github.ref(_name)`, so the dispatch path targets the requested tag rather than the workflow's default branch. The normal `release: published` path is unchanged.

## Note

The deeper cause — prerelease branches lagging `main`'s workflow updates — is mitigated separately by keeping `prerelease/*` branches rebased on `main` before tagging. This PR adds the recovery lever that was missing when that invariant slips.

The `3.7.0-1986` publish has already been recovered by re-dispatching `create-release.yml` (which forced the published event from `main`'s fixed workflow); this PR ensures the simpler `release.yaml` dispatch lever exists next time.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,9 +1,28 @@
 name: Release
 
+# Normally triggered by `release: published`, which `create-release.yml` emits
+# (it flips the release draft->published with the App token so the event
+# cascades — the default GITHUB_TOKEN cannot trigger downstream workflows).
+#
+# RECOVERY LEVER (workflow_dispatch): a tag-push runs create-release.yml from
+# the TAGGED tree, so a tag cut from a branch that predates the cascade fix
+# (e.g. a long-lived `prerelease/x.y.z` branch that never merged main's
+# workflow updates) runs an OLD create-release.yml that never forces the
+# published event — and this workflow then never fires. That is exactly what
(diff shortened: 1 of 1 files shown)


zcash-harry (approved):

zcash-harry:
UtACK

PR #2349, opened 2026-06-29
Fix crash: EncryptedSharedPreferences fails after device-to-device transfer (MOB-1452)

noop-sk:
## Problem

24 affected users (41% increase) on Android 16 experience a crash loop on every app launch:

```
javax.crypto.AEADBadTagException
  at EncryptedSharedPreferences.create(...)
```

**Root cause:** Device-to-device transfer copies the `EncryptedSharedPreferences` file but not the Android Keystore key used to decrypt it. On the new device, the app tries to decrypt the file with a non-existent key → `AEADBadTagException` → crash.

The key is NOT bound to biometric authentication — the issue is purely the Keystore being hardware-bound and non-transferable.

## Fix

Wrap `EncryptedSharedPreferences.create()` in a try/catch. On failure:
1. Delete the corrupted prefs file
2. Delete the stale Android Keystore master key entry
3. Recreate fresh

The user will need to re-enter their seed phrase once, after which everything is restored (address book and metadata keys are re-derived from seed, so no permanent data loss for those).

## Impact

- Voting recovery state and migration plan state are lost (not seed-derivable)
- Address book and transaction metadata are recoverable after seed re-entry
- No wallet funds affected

## Files Changed

- `preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt`

### DIFF
--- a/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
+++ b/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION")
+
 package co.electriccoin.zcash.preference
 
 import android.content.Context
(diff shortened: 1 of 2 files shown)


nesence-m (approved):
utACK

Issue #13, opened 2021-10-05
Set up Preference Infrastructure


ccjernigan:
We need a consistent methodology for storing preferences in the app.  Although the initial use case will be a boolean for whether the user has gone though the app first launch, there will be other values that need to be stored.  Without over designing, some plan is needed to deal with preferences in a reasonable way.

 - How do we assert preference keys are not duplicated accidentally (copy-paste)?
 - How do we ensure type safety for preference key-value pairs?
 - How do we store default values associated with a preference key?
 - Where is the preference filename defined?
 - Do we need to differentiate between preferences that should and should not be backed up?  (E.g. store them in a separate file?)

Specific answers to those questions will be defined as part of this implementation. We may be able to leverage recent commits to the old wallet app to inspire answers to some of those questions, e.g. reflection-based tests for the keys being unique and model objects representing a default value.

Although we'd like to use [Datastore](https://developer.android.com/jetpack/androidx/releases/datastore), it doesn't appear to support encryption so [EncryptedSharedPreferences](https://developer.android.com/topic/security/data) might be the way to go currently.

PR #2351, opened 2026-06-29
MOB-1397: Auto-recover from SeedNotRelevant without user interaction

noop-sk:
When WalletCoordinator emits isSeedMismatch=true, WalletViewModel.init automatically calls RecoverFromSeedMismatchUseCase which uses deleteSdkDataFlow (now lockout-aware) to wipe the mismatched DB, then re-persists the current wallet seed so the synchronizer rescans from scratch.

Also adds a debug menu button to simulate SeedNotRelevant by storing a random fake seed and killing the process, reproducing the real MT6769 scenario.

> **Note**

# Author

> **Note**

# Reviewer


nesence-m (changes requested):
The recovery design itself survived an adversarial pass — the trigger is a deterministic backend verdict (not a transient state), only SDK databases are wiped while app prefs/the seed are never touched, the wiped DB belongs to a seed not on the device so nothing recoverable is destroyed, the import-different-seed race resolves correctly, and reentrancy is covered (single activity-scoped collector, SDK-side lockout mutex, wipe running in the SDK's own scope). But the PR can't land as-is:

**Blocking**

1. **Unbuildable against any reachable SDK revision.** `SynchronizerProvider.kt:82` and `RecoverFromSeedMismatchUseCase.kt:17` consume `WalletCoordinator.isSeedMismatch` / lockout-aware `deleteSdkDataFlow`, which exist only on the head of unmerged SDK PR zcash/zcash-android-wallet-sdk#2007 — verified by fetching `WalletCoordinator.kt` at both the head branch's effective `SDK_COMMIT_PIN` and the base's current pin: neither has the API, and this PR doesn't touch `gradle.properties`. That fully explains the red builds; it only compiled locally via `SDK_INCLUDED_BUILD_PATH`. Needs #2007 merged + a pin bump first. **Also note:** the review on #2007 found its catch is dead code on the default Slipstream build (`SeedNotRelevant` latches into `onSetupErrorHandler`, never leaving `new()`) — so even with a pin bump, `isSeedMismatch` would never fire in production and this auto-recovery would never trigger; the two PRs need to be fixed together.

2. `WalletViewModelTest.kt:75` — still constructs `WalletViewModel(synchronizerProvider, walletRepository)` while the PR adds a required third constructor param; ui-lib unit tests don't compile (explains the `test_kotlin_modules` failure). Verified untouched at head.

3. Merge state `CONFLICTING`/`DIRTY`, 20 commits behind `maint/v3.10.x` — needs a rebase. The feature commit itself should rebase cleanly (the patched `WalletViewModel` shape is identical on base); the conflicts most likely come from the unrelated second commit (see 10).

4. ktlint/detekt red, reproducible from the diff: `AndroidPreferenceProvider.kt:23` (`android.util.Log` import mid-kotlinx block; also `Log` instead of Twig at 217/226/235/239), `UseCaseModule.kt:144-145` and `SimulateSeedNotRelevantUseCase.kt:7` (import ordering).

**Non-blocking**

5. `RecoverFromSeedMismatchUseCase.kt:17` — the `deleteSdkDataFlow()` Boolean is ignored. The SDK reports a failed erase as `false` (e.g. an engine synchronizer still keyed) rather than throwing; proceeding to `store()` then re-hitting `SeedNotRelevant` produces an **unthrottled automatic wipe-retry loop**. Check the result and cap retries or surface an error state.

6. `RecoverFromSeedMismatchUseCase.kt:15-16` — the comment "Erase clears all prefs — seed is re-persisted immediately after" is factually wrong for this SDK: `deleteSdkDataFlow` explicitly preserves wallet secrets; `store()` is only needed to flip `walletInitMode`. A wrong safety comment on the PR's most safety-critical line will mislead future readers into seeing a seed-loss window that doesn't exist.

7. `RecoverFromSeedMismatchUseCase.kt:17-18` — between lockout release and `store()`, `flatMapLatest` restarts a synchronizer from the old stored wallet against the fresh DB, then `store()` rebuilds again — wasted double init plus a benign process-death window where the init-mode flip is unapplied.

8. `WalletViewModel.kt:80-90` — recovery anchored to an activity-scoped VM means a mismatch hit by background synchronizer use won't recover until the UI opens; the provider/repository layer would be a more robust home. Nit: if the wallet is concurrently nulled (Reset Zashi), `deleteSdkDataFlow` never emits and `.first()` suspends forever.

9. `SimulateSeedNotRelevantUseCase.kt:24-25` — the debug action irreversibly overwrites the real stored seed with random entropy, no confirmation; a tester with a funded debug wallet loses it unless backed up. Consider stashing the real seed or adding a confirm dialog.

10. Commit 8473c7961 ("Redact PII from migration doc comments") is unrelated scope — 5 feature-migration files of comment rewording, the likely conflict source; better as its own PR.

11. No CHANGELOG entry (crash loop → silent auto-recovery is user-facing) and no tests for the use case or the auto-trigger (checklist box unticked).

nesence-m (changes requested):
The fix round is genuinely substantial — the test-constructor break, the lint findings, the rebase, the wipe-retry cap (with good tests), the confirmation gate on the debug seed-overwrite, the corrected erase-semantics comment, and the CHANGELOG/tests gap are all addressed, and the declined nits (double-init, VM-scoped anchoring) are documented with reasoning I can accept as follow-ups. Two blockers from the previous review remain, however:

1. **Still unbuildable.** `SDK_COMMIT_PIN` is untouched and `ZCASH_SDK_VERSION=3.1.1-SNAPSHOT` doesn't contain `WalletCoordinator.isSeedMismatch` — the companion SDK PR [#13 MOB-1397: catch SeedNotRelevant and fix Slipstream dead-code path](https://github.com/zodl-inc/zodl-android-wallet-sdk/pull/13) is still open. All four red CI jobs fail on the same `Unresolved reference 'isSeedMismatch'` at `SynchronizerProvider.kt:82`. This needs SDK #13 landed and either a snapshot republish or an `SDK_COMMIT_PIN` pointing at the merged sha (per the pin's own doc comment, not the PR-branch sha).

2. **The PII-redact commit was duplicated, not dropped.** Commit "Redact PII from migration doc comments" ([a8271878e](https://github.com/zodl-inc/zodl-android/pull/2351/commits/a8271878ec9f725445175317ab34445f1b715f49)) is still the head of this PR while [#2493 Redact PII from migration doc comments](https://github.com/zodl-inc/zodl-android/pull/2493) carries the identical changes. Please drop it from this branch and let #2493 own it — as it stands, whichever merges second will conflict.

Non-blocking notes for a follow-up (no action needed here): the retry counter resets on a successful erase, so the erase-succeeds-then-mismatch-recurs path is still technically unbounded (self-limiting in practice); and the declined item 8 (VM-scoped recovery + the Reset-Zashi `.first()` hang) would benefit from a tracked MOB ticket rather than only a code comment.

noop-sk:
Re: the PII-redact duplicate concern in the review — tested it directly rather than guessing. Simulated merging #2493 into `maint/v3.10.x`, then rebased this branch (which already carries the same redaction, cherry-picked from `8473c7961`) onto the result: `git rebase` detects the duplicate patch via patch-id and auto-skips it (`skipped previously applied commit`) — no conflict, tree stays clean. So no action needed here; once #2493 merges, the next real rebase of this branch (needed anyway once SDK #13 lands) will skip the redundant commit on its own.

Still blocked on the other item: SDK #13 (`isSeedMismatch`) hasn't merged yet.

noop-sk:
Rebased onto `maint/v3.10.x` now that SDK #13 (`isSeedMismatch`) has merged — CHANGELOG conflict resolved cleanly, no other conflicts.

Still not buildable in CI though, and the reason is different from before: `.github/actions/setup-sdk-companion/action.yml` still hardcodes `repository: zcash/zcash-android-wallet-sdk` (the old upstream repo) for its `SDK_COMMIT_PIN` checkout. SDK #13 merged into the *new* `zodl-inc/zodl-android-wallet-sdk`'s `maint/v3.1.x` — a different repo the companion action doesn't know how to point at. Without a real SDK_INCLUDED_BUILD_PATH substitution, this branch resolves the SDK from Maven Central at whatever version is currently published there, which doesn't have `isSeedMismatch` yet.

Two ways forward, neither of which I want to do unilaterally right now:
1. Make the companion action's `repository:` input configurable (touches shared CI infra used by every PR — needs a real review, not a rushed pin-hack; a previous attempt at a narrower version of this was reverted for exactly that reason, see the PII-redact/rebase history on this branch).
2. Cut an actual SDK snapshot/version from `zodl-inc/zodl-android-wallet-sdk` and bump the app's Maven dependency to it.

Leaving this for Milan — flagged in the handoff summary too. Not blocked on anything else in this PR's own diff.

nesence-m (commented):
Re-verified at `5f9d8a5c2`.

**Buildability — still blocking, and I confirm your diagnosis.** `SDK_COMMIT_PIN` is empty and `ZCASH_SDK_VERSION=3.1.1-SNAPSHOT` still lacks `isSeedMismatch`; all red jobs fail on the same `Unresolved reference 'isSeedMismatch'` at `SynchronizerProvider.kt:82`. SDK #13's merge commit `0865690d473e4719523b3b0d3e3d59234527e596` is the tip of `maint/v3.1.x` in `zodl-inc/zodl-android-wallet-sdk` — the right pin value per the pin's doc comment — but `.github/actions/setup-sdk-companion/action.yml` hardcodes `repository: zcash/zcash-android-wallet-sdk`, where that sha doesn't exist (API-verified), so the pin can't work until the action is repointed or a real SDK version is published and bumped. I'll take the call on which route; agreed neither belongs in this PR unreviewed.

One heads-up for whichever route lands: the merged SDK adds abstract `Synchronizer.setupError`, so `ui-lib/src/androidTest/.../fixture/MockSynchronizer.kt` will need `override val setupError: StateFlow<Throwable?> = MutableStateFlow(null)` or androidTest compilation breaks the moment the dependency moves.

**PII-redact commit** — I reproduced your reasoning: patch-id auto-skip on rebase is real, and identical hunks merge cleanly either way, so I'll accept keeping it on the branch until #2493 merges; please do let the next rebase drop it rather than carrying it into the merge.

Everything else from the review round checks out in the diff (wipe-retry cap, confirmation on the debug simulate, tests). Please also open a MOB ticket for the two follow-ups now documented in the `WalletViewModel` comment (VM-scoped recovery; Reset-Zashi `.first()` hang) — they're written down in code but tracked nowhere.

Keeping my earlier request-changes in force until the SDK dependency question is resolved and CI is green.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RecoverFromSeedMismatchUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+            failedAttempts += 1
+            Twig.error {
+                "Seed-mismatch recovery: deleteSdkDataFlow reported failure " +
+                    "(attempt $failedAttempts/$MAX_RECOVERY_ATTEMPTS)"
+            }
+            throw SeedMismatchRecoveryException(failedAttempts)
+        }
+
+        failedAttempts = 0
+        persistableWalletProvider.store(recoveryWallet)
```
nesence-m, line 71: **CONFIRMED race — deleted-wallet resurrection.** `invoke()` snapshots the wallet at entry, then suspends in `deleteSdkDataFlow().first()`; the SDK's lockout mutex queues this erase *behind* a concurrently running Reset Zashi erase. Reset finishes and clears encrypted prefs (seed gone), then this erase succeeds against the clean DB and `store(recoveryWallet)` unconditionally writes the pre-recovery wallet JSON — seed phrase included — back into encrypted prefs. Next launch, the wallet the user just deleted is back with its secret intact. The KDoc's framing ("store only flips WalletInitMode") holds in the happy path but not across this interleaving. Minimal guard: re-read the persisted wallet after the erase and skip `store()` unless it still identity-matches the entry snapshot (cleared or replaced ⇒ bail out).

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RecoverFromSeedMismatchUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+            Twig.error {
+                "Seed-mismatch recovery: giving up after $failedAttempts failed attempt(s), not retrying erase again"
+            }
+            throw SeedMismatchRecoveryException(failedAttempts)
+        }
+
+        val recoveryWallet = wallet.copy(walletInitMode = WalletInitMode.RestoreWallet)
+
+        // Use lockout-aware erase so any lingering synchronizer is stopped before the DB wipe.
+        val erased = walletCoordinator.deleteSdkDataFlow().first()
```
nesence-m: **CONFIRMED — one hung erase permanently disables auto-recovery.** At the pinned SDK rev, `deleteSdkDataFlow`'s inner launch does `persistableWallet.first()?.network ?: return@launch` — with a concurrently nulled wallet the `callbackFlow` never emits and never closes, so this `first()` suspends for the collector's lifetime. Because `WalletViewModel`'s init collector is a plain sequential `collect`, the wedged call then blocks every subsequent `isSeedMismatch = true` emission for the rest of the session (reset → restore different wallet → new legitimate mismatch → no recovery ever fires). The PR comment concedes the hang but not that it one-shots the whole feature. Cheap guard: `withTimeout` around the erase, counting a timeout as a failed attempt.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RecoverFromSeedMismatchUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+ * erase is retried up to [MAX_RECOVERY_ATTEMPTS] times (tracked for the lifetime of this instance,
+ * which is owned for the lifetime of the injecting ViewModel) and, once exhausted, further
+ * invocations throw [SeedMismatchRecoveryException] without attempting another erase — an
+ * explicit, observable failure instead of a silent infinite loop.
+ */
+class RecoverFromSeedMismatchUseCase(
+    private val persistableWalletProvider: PersistableWalletProvider,
+    private val walletCoordinator: WalletCoordinator,
+) {
+    private var failedAttempts = 0
```
nesence-m, line 25: `MAX_RECOVERY_ATTEMPTS` guards an in-memory counter on a **Koin `factoryOf`-scoped** class (UseCaseModule.kt:296), so the bound resets on every ViewModel recreation and process restart — the exact conditions of the crash-loop being fixed. A persistently failing erase costs 3 destructive wipe attempts per app open, forever; with the leaked second `WalletViewModel` instance (see comment there) the effective concurrent budget doubles. The counter is the whole point of the class — it needs process scope (single) or persistence.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/viewmodel/WalletViewModel.kt
```diff
@@ -70,6 +77,27 @@ class WalletViewModel(
...
+    // 1. This recovery is anchored to an activity-scoped ViewModel, so a mismatch hit by
+    //    background synchronizer use won't auto-recover until the UI is next opened. The
+    //    provider/repository layer (e.g. SynchronizerProvider itself) would be a more robust
+    //    home for this collector so it runs regardless of UI lifecycle.
+    // 2. If the wallet is concurrently nulled out (Reset Zashi races this), deleteSdkDataFlow()
+    //    may never emit and RecoverFromSeedMismatchUseCase's `.first()` suspends forever. Not
+    //    observed in practice, but worth guarding (e.g. a timeout or cooperating with the reset
+    //    flow) in a follow-up.
+    init {
+        viewModelScope.launch {
```
nesence-m: This eager init collector gets **multiplied by a Koin quirk**: `AuthenticationViewModel` constructor-injects `WalletViewModel` (ViewModelModule.kt:104/106), so Koin builds a second, store-less instance whose scope is never cancelled (`onCleared` never runs). Pre-PR that extra instance was inert (all its flows `WhileSubscribed`); now each instance runs this collector for the process lifetime — on a mismatch, two-plus collectors invoke the destructive recovery concurrently (overlapping erases, racing `store()` calls), and every activity relaunch leaks another. The follow-up comment above already names the right home: the app-scoped `SynchronizerProviderImpl` (it already hosts an equivalent init-block watcher and injects `WalletCoordinator` + `PersistableWalletProvider`), which also fixes recovery never running during background-only synchronizer use (MigrationWorker). Given the destructive-duplication angle, this is worth doing now rather than as a follow-up.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/viewmodel/WalletViewModel.kt
```diff
@@ -70,6 +77,27 @@ class WalletViewModel(
...
+            isSeedMismatch
+                .filter { it }
+                .collect {
+                    runCatching { recoverFromSeedMismatch() }
```
nesence-m: Two issues in this handler: (1) `runCatching` also catches `CancellationException`, so ordinary ViewModel teardown mid-erase logs a spurious "Auto-recovery from seed mismatch failed" with a cancellation stack trace — polluting exactly the log channel used to triage this feature; rethrow cancellation. (2) Terminal failure (`SeedMismatchRecoveryException` after exhausted retries) is **log-only**: the app keeps running with `synchronizer = null`, nothing syncs, and no UI explains why — support gets unactionable "wallet stuck" reports. The codebase already has the mechanism (`NavigateToErrorUseCase` with `ErrorArgs.General`, as used by `OnZip321ScannedUseCase`/`RequestSwapQuoteUseCase`) — the terminal branch should surface through it.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/viewmodel/WalletViewModel.kt
```diff
@@ -70,6 +77,27 @@ class WalletViewModel(
                 initialValue = false
             )
 
+    // MOB-1397 review, follow-up not addressed here (lower-priority, needs a larger restructure
```
nesence-m: Style (team Kotlin convention — no `//` line comments, KDoc or nothing): this new block plus the other `//` comments added in this PR (DebugVM.kt:33/:228, RecoverFromSeedMismatchUseCase.kt:61, WalletViewModelTest.kt:111/:131, RecoverFromSeedMismatchUseCaseTest.kt:68/:74, EncryptedPreferenceProviderTest.kt:170-171) should be KDoc on the declarations or dropped. This particular block largely dissolves once the collector moves to `SynchronizerProviderImpl` and the erase gets a timeout — both items it defers.

gradle.properties
```diff
@@ -124,7 +124,7 @@ SDK_INCLUDED_BUILD_PATH=
 # from something more durable than the SDK's own PR branch (which stops being fetchable once
 # that branch is deleted): repoint to the merged sha on origin/main (or the relevant
 # origin/maint/vX.Y.x line) once the companion SDK PR lands.
-SDK_COMMIT_PIN=
+SDK_COMMIT_PIN=0865690d473e4719523b3b0d3e3d59234527e596
```
nesence-m, line 127: Process note on the pin itself (co-development flow is intended, but untracked): with `SDK_COMMIT_PIN` live on `maint/v3.10.x`, the maintenance line is **unreleasable** — create-release/deploy/release workflows all hard-fail on a non-blank pin — until SDK 3.1.1 ships and a follow-up PR clears this. Separately, the `isSeedMismatch`/`deleteSdkDataFlow` API this PR consumes exists only on the SDK's `maint/v3.1.x` (merge `0865690d`), not SDK main — without the SDK-side forward cascade, a later main-line SDK bump silently loses the API. Both need an explicit tracking item before an urgent v3.10.2 becomes necessary.

nesence-m (commented):
Deep review pass (Claude-assisted, verified findings only). Approval stands; posting these so they're on record — fixes for the confirmed items are being pushed to this branch as a follow-up commit. Two findings anchor outside the diff, so they live here:

**1. `ResetZashiUseCase.closeSynchronizer()` deadlocks in the latched seed-mismatch state (CONFIRMED).** With `isSeedMismatch` latched, `WalletCoordinator`'s `synchronizer` StateFlow stays `null` (verified at SDK pin `0865690d`). Once recovery exhausts its 3 attempts, Delete wallet: biometric succeeds, `deleteLocalFiles()` removes the address book/metadata, then `closeSynchronizer()` suspends forever in `getSynchronizer()` (`filterNotNull().first()`, no timeout) — seed and SDK data are never wiped even though the user authenticated the deletion. Pre-PR this state crash-looped at startup, so the deadlock is newly reachable. Fix: bound the wait (`withTimeoutOrNull`) and proceed with the wipe when no synchronizer materializes — `clearSDK()`'s lockout-aware erase handles teardown anyway.

**2. `feature-voting` unit tests no longer compile (CONFIRMED).** The `FakeSynchronizerProvider` in `VotingKeystoneRepositoryTest.kt` was not updated for the new abstract `isSeedMismatch` (the other two fakes were). PR CI doesn't catch it because `pull-request.yml` only runs `:ui-lib:testZcashmainnetStoreDebugUnitTest`; any local/full-suite run of `:feature-voting` unit tests fails.

Two review candidates were investigated and refuted (wrong-init-mode/no-rescan claim; StateFlow-conflation retry stall), so the inline comments below are the verified surface.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
```diff
@@ -72,11 +80,15 @@ class SynchronizerProviderImpl(
     // crashes startup with a StackOverflowError (caught on-emulator 2026-07-28). Resolved at
     // first hook fire instead, when the graph is fully built.
     private val migrationSyncedHook: Lazy<MigrationSyncedHook>,
+    private val recoverFromSeedMismatch: RecoverFromSeedMismatchUseCase,
```
nesence-m: provider is lower level than use-case. So UseCase cannot be seen by provider

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
```diff
@@ -127,6 +139,47 @@ class SynchronizerProviderImpl(
                     }
                 }
         }
+        startSeedMismatchRecoveryCollector()
```
nesence-m: startSeedMismatchRecoveryCollector needs to go to a separate use-case which will be fired from app class

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RecoverFromSeedMismatchUseCase.kt
```diff
@@ -0,0 +1,141 @@
...
+import cash.z.ecc.android.sdk.model.PersistableWallet
+import co.electriccoin.zcash.spackle.Twig
+import co.electriccoin.zcash.ui.common.provider.PersistableWalletProvider
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.withTimeout
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * Recovers from the SDK reporting a seed mismatch (the `SeedNotRelevant` crash trigger, surfaced
```
nesence-m: too many comments, noone gonna read this

nesence-m (changes requested):

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
nesence-m: Fixed in https://github.com/zodl-inc/zodl-android/pull/2351/commits/209063965a6bd8e8198464682b53ecd3d257f93c — `SynchronizerProviderImpl` no longer injects `RecoverFromSeedMismatchUseCase` / `NavigateToErrorUseCase`; it only exposes `isSeedMismatch` as state again.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
nesence-m: Fixed in https://github.com/zodl-inc/zodl-android/pull/2351/commits/209063965a6bd8e8198464682b53ecd3d257f93c — the collector is now `ObserveSeedMismatchUseCase` (ui-lib, Koin factory), launched once from `ZcashApplication.onCreate` on `applicationScope` next to `observeSynchronizerError()`. Tests moved to `ObserveSeedMismatchUseCaseTest` (deterministic `runTest`, no real dispatchers).

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RecoverFromSeedMismatchUseCase.kt
nesence-m: Trimmed in https://github.com/zodl-inc/zodl-android/pull/2351/commits/209063965a6bd8e8198464682b53ecd3d257f93c — class KDoc is 6 lines now; the helper and exception KDocs are 1–3 lines.

app/src/main/java/co/electriccoin/zcash/app/ZcashApplication.kt
```diff
@@ -120,6 +123,10 @@ class ZcashApplication : CoroutineApplication() {
         }
     }
 
+    private fun observeSeedMismatch() {
+        applicationScope.launch { observeSeedMismatchUseCase() }
```
nesence-m: observeSeedMismatch() instead of observeSeedMismatchUseCase() pls

nesence-m (dismissed):

app/src/main/java/co/electriccoin/zcash/app/ZcashApplication.kt
nesence-m: Done in https://github.com/zodl-inc/zodl-android/pull/2351/commits/f6b4a505a8f00f7f3d72d7e5067bca5b7f8b0a69 — the injected property is `observeSeedMismatch` now and `onCreate` launches it directly; the wrapper is gone.

nesence-m (approved):
Re-approving after the observeSeedMismatch rename nit; ruleset dismissed the earlier approval on push.

PR #2357, opened 2026-07-03
MOB-1403: cold-start — batch tx reads, recipient as String, native-lib preload

nesence-m:
### Changes
1. **batch tx outputs/recipients, Transaction.recipient as String** — warm per-tx output and recipient caches with single batched SDK queries instead of a per-transaction DB call on every activity-list emission; model `Transaction.recipient` as a `String` (validated to a `WalletAddress` only where needed, in `GetTransactionDetailByIdUseCase`).
2. **preload SDK native library at app startup** — call `Synchronizer.preloadNativeLibrary()` off-main from `Application.onCreate` so the one-time native load overlaps splash/auth.
3. **shimmer loading on balance and account app bar** — show shimmer placeholders while balances and account data load instead of blocking the UI.
4. **performance optimizations** — app startup and recomposition tweaks, new `Synchronizer` batch API overrides in test mocks, and parallel Gradle sync (`org.gradle.tooling.parallel=true`).

### ⚠️ Depends on the SDK PR
Depends on https://github.com/zcash/zcash-android-wallet-sdk/pull/2012 — relies on new SDK APIs (batched `getRecipients`/`getTransactionOutputs`, `TransactionId.txIdString`, `Synchronizer.preloadNativeLibrary()`, lazy Tor client). **Blocked** until that SDK PR merges, publishes, and `ZCASH_SDK_VERSION` is bumped to the version that includes them — CI will not compile before that.


zcash-harry:
Pushed b80c10c0c: the Activity list never loaded on any wallet that hadn't explicitly switched accounts. `SelectedAccountUUIDProvider.uuid` is only ever written by `AccountDataSource.selectAccount`, so on a fresh/default-account wallet it stays `null` and `TransactionRepository.transactions` short-circuits to a permanent `null` (endless shimmer). Restored `AccountDataSource.selectedAccount` as the source, which treats the default account as selected when the preference is unset. Verified on emulator — activity list populates again.

zcash-harry:
Before: 
Screenshot 2026-07-16 at 9 05 29 PM

After:
Screenshot 2026-07-17 at 9 43 34 AM

PR #2365, opened 2026-07-20
fix(tor): set isSecure=false for .onion custom server endpoints

y4ssi:
## Problem

`ZashiEndpointTextFieldParser.toEndpointOrNull()` always set `isSecure = true` for custom server entries. When the host ends in `.onion`, this causes `CombinedWalletClientImpl` to build an `https://` URI for the gRPC channel, which fails with `InvalidContentType` because the Tor stream is plaintext HTTP/2.

## Change

In `ZashiEndpointTextField.kt`, derive `isSecure` from the host name:

```kotlin
val isSecure = !host.endsWith(".onion") && uri.scheme != "grpc"
```

This matches the existing `grpc://` scheme override and adds `.onion` as a second plaintext case.

## Testing

Tested end-to-end on an Android emulator (x86_64, API 34). With this fix and the accompanying SDK changes (zcash/zcash-android-wallet-sdk#2020), the wallet successfully syncs compact blocks through a Tor v3 onion lightwalletd endpoint configured as a custom server.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -220,3 +220,8 @@ JVM_TOOLCHAIN=17
 # This shouldn't be changed, as Android doesn't support targets beyond Java 8
 KOTLIN_JVM_TARGET=8
 ANDROID_JVM_TARGET=1.8
+
+# Zodl self-hosted lightwalletd Tor hidden service.
+# To use: enable Tor in Settings, then enter this in the custom server field:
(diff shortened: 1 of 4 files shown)


Otr437 (approved):

nesence-m (changes requested):
@y4ssi can u point this to maint/3.8.x branch?

nesence-m (approved):
utACK

PR #2366, opened 2026-07-22
[MOB-1475] Stop implying refunds always come back as USDC on NEAR

zcash-harry:
## Summary
- The refund-address explainer used "USDC on NEAR" as an illustrative example of the source currency/network, but users reported reading it as a literal, fixed refund destination and feared their funds would be sent to the wrong asset/network.
- Changes the wording to "in the source currency on the same network," which is true for any swap.

Fixes MOB-1475.

## Test plan
- [x] `ktlintFormat`/`ktlint`/`detekt` clean
- [x] Verified strings.xml (en + es) are well-formed XML


nesence-m (approved):
lgtm

PR #2373, opened 2026-07-27
Lower the minimum accepted Keystone firmware to 3.0.1

Chlup:
## Summary

Lowers the minimum Keystone firmware version accepted when scanning a signed transaction from **3.0.3** to **3.0.1**.

- `KeystoneFirmwareVersion.MINIMUM_SUPPORTED` → `3.0.1` (the single point of change; the firmware gate, error prompt, exception message, and log line all render the minimum from this constant, so nothing else needed touching)
- Updated `minimumSupportedMatchesProductRequirement` in `KeystoneFirmwareVersionTest`, which pins the constant
- Updated the 3.8.0 changelog entry so the release notes state the shipped minimum
- Dropped the now-stale KDoc clause that justified pinning the minimum at 3.0.3

## Test plan

- Not built or run locally (machine not set up for Android dev) — relying on CI for the unit test suite; `KeystoneFirmwareVersionTest` covers the updated pin, and the remaining comparison/normalization tests are unaffected by the new minimum.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,18 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+## [3.8.0 (2023)] - 2026-07-27
+
+### Changed:
+- We lowered the minimum supported Keystone firmware for signing from 3.0.3 to 3.0.1.
+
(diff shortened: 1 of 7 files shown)


nesence-m (approved):
utACK

noop-sk (approved):

PR #2387, opened 2026-07-30
ci(release): remove Solana dApp Store publish job

y4ssi:
Solana store deploys are being done manually. Removing the automated publish job from the release pipeline.


nesence-m (changes requested):
@y4ssi can u point the PR to maint/v3.8.x?

Also gplay deploy always times out on 10m - can we increase it to 30m while we at it? :)

nesence-m (approved):
utACK

PR #2404, opened 2026-08-10
MOB-1669: skip eager live-driver start after committing a migration plan

noop-sk:
## Summary
- Skip eagerly starting `MigrationLiveDriver` synchronously right after `finalizeMigrationSchedule` commits a plan — matches iOS, which never eagerly proves/drives after storing a signed batch (verified live: this was blocking a `finalizeReadyTransfers` call for 36.6s on a physical device). The already-armed `MigrationScheduler` job still drives the plan forward; proving is deferred, not skipped.
- Fix 7 `feature-migration` unit tests that were silently coupled to which Gradle flavor variant compiled them (`isTestnetBuildFlavor()`'s ambient `BuildConfig.FLAVOR` default resolved differently under `testZcashmainnetStoreDebugUnitTest` vs `testZcashtestnetStoreDebugUnitTest`) — pinned `fineGrained` explicitly at the affected call sites instead of relying on the ambient default.
- Fix a Compose layout bug in the "Prepare Your Balance" detail sheet where a long dependency list ("Waits on steps 1, 2, ..., 14") had no width bound, squeezing the weighted title column down to one character per line.
- Add a loading spinner to the Keystone Sign Transaction screen's QR area — it previously rendered nothing at all while the first animated-QR part was being built, making the screen look broken during any non-instant QR generation.
- Fix an Android string-resource whitespace bug: `migrationSetup_bodyPrefix`/`_bodyFiatSuffix`/`_bodySuffix` each needed a leading/trailing space to join correctly, but Android's string-resource compiler silently trims un-quoted `<string>` whitespace — quoted all three (en + es) to preserve it. Live symptom: `"your0.22786965 ZEC($119.72)from the Orchard pool..."`.
- Minor: add a spacer between the migration-setup body text and the "Find out more" link.

## Scope note
This is app-only. The *original* MOB-1669 root cause — "Android took 3.5 minutes to generate 3 batched QRs" during Keystone signing — turned out to live in the SDK's Rust migration engine (a spendable-note DB query re-run on every call instead of cached once per JNI invocation), not in this app-side path. That fix is a separate, unpushed SDK commit (`bugfix/MOB-1669` on `zcash-android-wallet-sdk`) and will need its own SDK release before the app can pick up the underlying speedup. The eager-live-driver-start fix in this PR is a real, independently-confirmed bug (not the one Harry originally reported), kept because it's a genuine improvement matching iOS's architecture.

## Test plan
- [x] `./gradlew :feature-migration:testZcashmainnetStoreDebugUnitTest :feature-migration:testZcashtestnetStoreDebugUnitTest :ui-lib:testZcashmainnetStoreDebugUnitTest` — all green
- [x] `detektAll`, `ktlint` — clean
- [x] Live-verified on a physical Pixel (mainnet internal debug): eager-driver-start stall eliminated, QR spinner renders correctly, prep-details sheet layout fixed, setup-screen text spacing fixed

### DIFF
--- a/feature-migration/src/main/java/co/electriccoin/zcash/ui/common/model/migration/MigrationPreparationDetails.kt
+++ b/feature-migration/src/main/java/co/electriccoin/zcash/ui/common/model/migration/MigrationPreparationDetails.kt
@@ -47,9 +47,17 @@ fun preparationStepTitle(number: Int, stepCount: Int): StringResource = stringRe
  * on EVERY row inside this sheet, including the one that's already due. [formatMigrationDuration]'s
  * own `coerceAtLeast` floor already turns a zero/negative [secondsUntil] into its network floor, so
  * no separate clamping is needed here.
+ *
+ * [fineGrained] forwards to [formatMigrationDuration] — exposed here (default unchanged: ambient
+ * [isTestnetBuildFlavor]) purely so a caller (a test) can pin it explicitly instead of depending on
+ * which Gradle flavor variant happened to compile the caller (MOB-1669 test-flakiness follow-up,
+ * 2026-08-09: this default silently resolved differently under `testZcashmainnetStoreDebugUnitTest`
+ * vs `testZcashtestnetStoreDebugUnitTest`, since both compile and run this same JVM test).
  */
-fun preparationStepTimeLabel(secondsUntil: Long): StringResource =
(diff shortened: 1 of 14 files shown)


nesence-m (approved):
utACK

PR #2406, opened 2026-08-10
MOB-1678: Fix multi-bundle CHP voting delegation

noop-sk:
## Summary
- Multi-bundle CHP voting rounds could crash constructing bundle 1+ (`no alpha for round=..., bundle=1`) because a zodl-only round-level phase column was advanced during PCZT construction, which then hard-rejected any later bundle's construction as a "phase regression" once an earlier bundle had been proved. Removed all round-level phase writes/gates from the voting flow and switched every "should I (re)construct/prove/submit this bundle" decision to the SDK's own per-bundle `DelegationPhase`, with an overwrite-refusal-aware retry so a `resetVotingSessionState` recovery can't leave a stale `Proved` read pointing at NULL data.
- `delegate-vote` submissions were rejected with 400 ("invalid message field: tx1 effects must be 821 bytes, got 0") because the vote-chain server now requires a `tx1_effects` field; the app was still sending the legacy `sighash` field. Threaded `tx1_effects` through from the SDK JNI layer to the API request body.
- Carries one unrelated pre-existing fix already on this branch: `MigrationSendingVM` no longer bypasses `DRIVE_LOCK`.

Companion SDK PR: zcash/zcash-android-wallet-sdk#2165

## Test plan
- [x] Built DIT (`assembleZcashtestnetInternalDebug`) and ran a live multi-bundle CHP voting round end-to-end on-device (`emulator-5554`) — delegation construction, proving, and submission all succeeded for bundle 1+ after bundle 0 was already proved; confirmed working by user on-device.
- [x] Full androidTest compile check passes.


feature-migration/src/main/java/co/electriccoin/zcash/ui/common/usecase/FinalizeMigrationScheduleUseCase.kt
```diff
@@ -79,14 +55,7 @@ class FinalizeMigrationScheduleUseCase(
...
         // plan's first bucket.
-        if (startLiveDriverImmediately) {
-            migrationLiveDriver.startIfNotRunning(accountKeyId)
-        } else {
-            migrationLog(
-                "FinalizeMigrationSchedule: skipping eager live-driver start (MOB-1669) — " +
-                    "the scheduled WorkManager job will drive this plan forward instead."
-            )
-        }
+        migrationLiveDriver.startIfNotRunning(accountKeyId)
```
nesence-m: **Major — merge silently reverts MOB-1669 (main, 2026-08-09).**

The merge of `origin/main` (`06f58c238`) resolved conflicts by keeping this branch's pre-MOB-1669 copies of the migration files, so merging this PR reverts four things that are on main today, with no replacement anywhere in the head tree (`git grep startLiveDriverImmediately` → 0 hits):

1. This line: the `startLiveDriverImmediately` opt-out is gone and `migrationLiveDriver.startIfNotRunning(accountKeyId)` is unconditional again — bringing back the ~3.5-minute blocking `finalizeReadyTransfers` pass right after committing a large Keystone batch (the WorkManager job armed by `migrationScheduler.schedule(...)` already covers this work). Callers in `MigrationReviewVM` and `MigrationScheduledVM` lost their `startLiveDriverImmediately = false` args.
2. The `fineGrained: Boolean = isTestnetBuildFlavor()` pinning params in `MigrationProgressVM`/`MigrationPreparationDetails` — their removal makes the migration unit tests flavor-dependent again (pass on testnet variants, fail on mainnet).
3. The `TITLE_COLUMN_WEIGHT`/status-column layout fix in `MigrationPreparationDetailsBottomSheet`.
4. The mid-batch-resume scenario in `MigrationKeystoneSignVMTest.failureSheetDismissOnlyHidesSheet` (incl. the `assertNotNull(pendingPczts.get(...))` assertion) — the accumulated-signatures-survive-dismiss invariant is now untested.

Suggest re-merging main and resolving in favor of main's versions of these files (none of them conflict with the voting work semantically).

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/repository/VotingKeystoneRepository.kt
```diff
@@ -203,18 +190,49 @@ class VotingKeystoneRepositoryImpl(
...
+                                roundName = session.title
+                            )
+                        }.recoverCatching { throwable ->
+                            if (!throwable.isDelegationSetupOverwrite()) throw throwable
+                            Log.i(
+                                TAG,
+                                "Keystone governance PCZT setup for round $roundId bundle $bundleIndex " +
+                                    "looks corrupted; resetting and rebuilding once"
+                            )
+                            votingCryptoClient.resetVotingSessionState(dbHandle, roundId)
```
nesence-m, line 230: **Major — round-wide reset can permanently invalidate already-scanned Keystone signatures.**

`resetVotingSessionState(roundId)` is round-scoped: at the pinned crate rev it clears setup (`alpha`, `pczt_sighash`, `rk`, `tx1_effects`, …) for **every** bundle with `delegation_tx_hash IS NULL AND bundle_index NOT IN keystone_signatures`. But the crate's `keystone_signatures` preservation guard is dead code in this integration — no SDK JNI wraps `store_keystone_signature` (checked PR #2165's `VotingRustBackend.kt`), and `storeBundleSignature` here writes only to the app-side `votingRecoveryRepository`. So signed-but-unsubmitted sibling bundles are *not* protected.

Reachable trigger: `buildGovernancePczt` persists Rust-side state inside the native call, and `storePendingKeystoneRequest` only runs after `redactPcztForSigner` — process death (or a redaction exception) in that window leaves Rust setup persisted with no app-side pending request. On resume, the rebuild hits the crate's overwrite refusal and this line wipes the whole round, including a sibling bundle 0 the user already signed on the device. Rebuilds draw `rand::thread_rng()`, so the new alpha/sighash can never match the scanned signature — and since bundle 0 is still in `keystoneBundleSignatures`, the app never re-requests it. The round wedges until fully cleared and re-signed.

Fix direction: scope the recovery to the failing bundle only, or persist scanned signatures into the crate's `keystone_signatures` table (the preservation hook already ships in the crate; needs a small JNI in the SDK PR) before any reset.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/usecase/SubmitVotesUseCase.kt
```diff
@@ -391,49 +407,55 @@ class SubmitVotesUseCase(
...
+            }.recoverCatching { throwable ->
+                // The crate refuses to silently overwrite already-persisted PCZT fields with
+                // different data — this bundle's setup is present and intact, nothing to do.
+                if (!throwable.isDelegationSetupOverwrite()) throw throwable
+            }.getOrThrow()
         }
 
-        val updatedRoundState = votingCryptoClient.getRoundState(dbHandle, roundId)
-        if (updatedRoundState?.proofGenerated != true) {
+        val alreadyProved =
```
nesence-m, line 453: **Major — Keystone path can pair a stale (old-alpha) proof with a rebuilt PCZT, silently.**

The comment above is right that a stale `proofs` row survives `resetVotingSessionState` (verified: the crate's reset is an `UPDATE` on `bundles` only; `proofs` rows are never touched), and `DelegationPhase` reads PROVED from the mere existence of a `success=1` proofs row (`phase_from_columns` does no consistency check against current bundle state). The `setupJustBuilt` guard covers this for software wallets — but it can only be set in the `!context.isKeystone` branch. The Keystone reset-and-rebuild happens in `VotingKeystoneRepository` (fresh alpha/rk/sighash written there), and no invalidation signal reaches this decision.

Result for a Keystone bundle rebuilt after a reset: `alreadyProved` is true from the surviving stale row, re-proving is skipped, and `getDelegationSubmissionWithKeystoneSig` assembles the **old-alpha proof** with the **new** bundle fields. The in-app sighash/rk checks pass (both sides see the new values), so the inconsistent delegation goes to the server, gets rejected — and every retry repeats identically, because no code path ever clears a `success=1` proofs row short of deleting bundles. Wedged round.

Fix direction: make the reset (or the rebuild write) invalidate the bundle's proofs row crate/SDK-side, or thread a rebuilt-since-prove signal from `VotingKeystoneRepository` into this check.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/usecase/SubmitVotesUseCase.kt
```diff
@@ -524,14 +547,17 @@ class SubmitVotesUseCase(
...
 
-        // A resumed round may already be past delegation/proving in Rust.
-        // Skip stale rebuild work instead of trying to regress the phase.
-        val rustRoundState = votingCryptoClient.getRoundState(dbHandle, roundId)
-        if (rustRoundState?.phase.hasVoteReady()) {
+        // A resumed round may already have this specific bundle submitted/confirmed. Per-bundle
+        // phase, not round-level: a multi-bundle round can have bundle 0 confirmed while bundle 1
+        // is still pending, and the round-level phase alone can't tell those apart.
+        if (currentDelegationPhase(dbHandle, roundId, bundleIndex).let {
+                it == DelegationPhase.SUBMITTED || it == DelegationPhase.CONFIRMED
```
nesence-m: **Major — SUBMITTED early-return skips the confirmation wait the old code performed; can wedge the round.**

SUBMITTED is derived Rust-side as `delegation_tx_hash IS NOT NULL AND van_leaf_position IS NULL` — i.e. it is true in exactly the crash window between `storeDelegationTxHash` and `storeVanPosition`. On a resumed submit in that state, the cached-hash fast-path above probes with `maxAttempts = 1`; if the tx isn't queryable yet it falls through — and this early-return then skips the full `awaitTxConfirmation` (~90s budget) + `storeVanPosition` that the pre-PR flow ran in this state. `submitVoteCommitmentsAndShares` next calls `generateVanWitnessJson`, which fails on `load_van_position` ("van_leaf_position not yet set") — a spurious user-facing failure where the old code just waited. (The "fall through and re-run delegation from scratch" comment above is now dead letter — this return prevents exactly that.)

Harsher sub-case: if the delegation tx landed with `code != 0`, the probe also yields no van position, and since nothing outside a full session reset ever clears `delegation_tx_hash`, every retry re-reads SUBMITTED and early-returns — permanently wedged.

Fix direction: on SUBMITTED (as opposed to CONFIRMED), await confirmation with the full retry budget and store the van position; clear the stored hash if the tx confirmed as failed so delegation can re-run.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/model/voting/StaticVotingConfig.kt
```diff
@@ -53,10 +53,14 @@ data class StaticVotingConfig(
...
+        // The old pin (commit 2785311d, checksum bed0116f) is now frozen for backwards
+        // compatibility with wallets that predate the v2 config schema (John's SDK gap analysis,
+        // #core-wallet 6.8., forwarded to #wallet-team 7.8.) — it still parses, but points at a
+        // stale trusted_keys set. This is the same config valargroup shared as the live "prod"
+        // custom-config URL for testing an unlisted round (verified live).
         const val BUNDLED_PINNED_SOURCE =
-            "https://raw.githubusercontent.com/valargroup/token-holder-voting-config/" +
-                "2785311d45758e85567d70a1f13709fa01b62c6b/prod/static-voting-config.json" +
-                "?checksum=sha256:bed0116f961226b256a574b52461ce81d9f5294a57e190987dc155f07eb1e431"
+            "https://voting.valargroup.org/prod/static-voting-config.json" +
```
nesence-m: **Major — compile-time sha256 pin now points at a mutable URL; first republish bricks voting for default users.**

Despite the `BUNDLED_` name, this is a live fetch and the **primary** source: `resolvePinnedConfigSource` returns it whenever the user has no custom source and remote config's `voting_config_url` is unset, and `fetchStaticConfig` does a no-cache GET + `decodeAndVerify`, which throws on any hash mismatch (fail-closed for the whole `fetchTrustedConfig` chain).

The previous value was a commit-pinned `raw.githubusercontent.com/...` URL — content-addressed, so the embedded checksum was valid forever. `https://voting.valargroup.org/prod/static-voting-config.json` is a mutable prod path: the first routine republish (new round, key rotation) makes every shipped build's checksum check fail and voting goes down for all default-configured users until a remote-config override or app update. Also worth noting: this change landed in `367405625`, whose commit message ("fix(migration): remove duplicate MigrationDriveOnce import") doesn't mention it.

Fix: keep the compiled-in pin content-addressed (commit-pinned raw URL), and let the mutable server path come via remote config only.

ui-design-lib/src/main/res/ui/common/values/strings.xml
```diff
@@ -1380,9 +1380,9 @@ Your vote can no longer be submitted.</string>
     <string name="migrationTorFailure_riskBody">Without Tor, your IP address is exposed to the server and could be linked to the migration amounts.</string>
 
     <string name="migrationSetup_title">Move to Ironwood</string>
-    <string name="migrationSetup_bodyPrefix">"Latest Zcash network upgrade requires moving your "</string>
-    <string name="migrationSetup_bodyFiatSuffix">" (%1$s)"</string>
-    <string name="migrationSetup_bodySuffix">" from the Orchard pool to the new Ironwood pool. Your funds are safe."</string>
+    <string name="migrationSetup_bodyPrefix">Latest Zcash network upgrade requires moving your </string>
```
nesence-m: **Major — removing the surrounding quotes drops load-bearing whitespace (EN and ES).**

`migrationSetup_bodyPrefix` / `_bodyFiatSuffix` / `_bodySuffix` are concatenated verbatim in `buildMigrationBodyText` (`MigrationSetupScreen.kt` — `append(prefix); append(zecAmount); append(fiatSuffix); append(suffix)`, no code-side separators), so their leading/trailing spaces are the only word separation. AAPT trims leading/trailing whitespace from *unquoted* string resources — the double quotes this PR removes (here and in `values-es/strings.xml` ~1102) were the mechanism preserving those spaces.

Post-PR the screen renders like: "Latest Zcash network upgrade requires moving your1.23 ZEC($45)from the Orchard pool…" in every locale.

Fix: restore the quotes in both locale files (or move the spacing into code).

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/model/voting/RoundPhaseExtensions.kt
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * True when constructing a bundle's governance PCZT failed because it was already built with
+ * different data (the crate's `store_delegation_data` refuses to silently overwrite persisted
+ * `pczt_sighash`/`padded_note_secrets`/`tx1_effects` for a bundle). This means the bundle's setup
+ * is genuinely present and intact — the caller should treat the existing setup as authoritative,
+ * not retry blindly. The sanctioned recovery from a *corrupted* setup (not this — this is the
+ * healthy case) is [co.electriccoin.zcash.ui.common.provider.VotingCryptoClient.resetVotingSessionState].
+ */
+fun Throwable.isDelegationSetupOverwrite(): Boolean =
+    message.orEmpty().contains("refusing to overwrite", ignoreCase = true)
```
nesence-m: **Minor (hardening)** — this matches only the top-level `message`, while its sibling `isQueryReturnedNoRows` in `VotingCryptoClient.kt` deliberately walks the cause chain with `generateSequence(this) { it.cause }` because native errors can arrive wrapped. Traced end-to-end today the refusal text does arrive top-level (crate `VotingError` → JNI `anyhow` flatten → `RuntimeException`, all Kotlin layers pass-through), so this works — but if any SDK/app layer ever starts wrapping native errors, all three overwrite-recovery paths silently turn into hard failures on resumed rounds. Cheap fix: reuse the same cause-walk here.

nesence-m (changes requested):
This is a deep review of the PR at head `7ea634b66`; every finding below was adversarially verified against the actual code (app + SDK PR #2165's JNI layer + the pinned `zcash_voting` rev `60f468a`) before being reported. Six major findings are attached as inline comments; requesting changes primarily for the merge-clobber and the two Keystone wedge bugs.

**Not inline-anchorable:**
- **Missing changelog entry** — `CHANGELOG.md` is untouched, but this is a user-facing bugfix PR (`docs/CODE_REVIEW_GUIDELINES.md`: "All changes should include this", plus the PR-template checkbox).

**Verified-and-cleared, for the record:** the ktlint import-order issue was already fixed by `7ea634b66`; the tx1_effects legacy-DB concern is moot (the crate's v13 migration wipes older voting DBs on open); the `DelegationPhase.fromWireValue` hard-fail and the hotkey-regeneration gate change were both verified as correct design, not defects.

**Cleanup-tier (non-blocking):** the `SUBMITTED || CONFIRMED` settled-predicate is copy-pasted at 4 sites; the overwrite-swallow idiom at 3 sites; the 9-arg `buildGovernancePczt` call is duplicated verbatim in the Keystone retry; `delegationPhases` is re-fetched up to 3× per bundle on the submit path; PROVED bundles lost their early-exit (redundant witness regen on retries); and the dead round-level helpers (`hasVoteReady`, the write-only `RoundStateInfo` fields + wire-int mapping) invite re-adoption of exactly the round-level gating this PR outlaws — worth deleting.

🤖 Review prepared with [Claude Code](https://claude.com/claude-code); all findings human-triaged.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/repository/VotingKeystoneRepository.kt
noop-sk: Status: companion SDK PR (zcash/zcash-android-wallet-sdk#2165) now ships `storeKeystoneSignature` (the missing preservation hook this finding calls for) — see that PR's commit `a92e7012`. Not yet wired up on this side: `VotingKeystoneRepository.storeBundleSignature` still only writes to the app-side `votingRecoveryRepository`, not the new SDK call. Tracking as a follow-up; leaving this open until wired.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/usecase/SubmitVotesUseCase.kt
noop-sk: Status: investigated on the SDK side (companion PR #2165) whether `resetVotingSessionState` could also invalidate the stale `proofs` row. No public crate function exists for it, and backend-lib has no precedent for hand-writing production SQL against the crate's schema, so it's documented as a known gap on `resetVotingSessionState`'s KDoc at both the TypesafeVotingDb and public VotingSdk layers rather than fixed. This one's still genuinely open — no fix path found yet on either side.

.github/actions/setup-sdk-companion/action.yml
```diff
@@ -0,0 +1,87 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Read SDK_COMMIT_PIN and enforce the branch gate
+      id: pin
+      shell: bash
+      run: |
+        PIN="$(grep -m1 '^SDK_COMMIT_PIN=' gradle.properties | cut -d= -f2 | tr -d '[:space:]')"
+        TARGET="${{ github.base_ref || github.ref_name }}"
+        if [ -n "${PIN}" ] && [ "${TARGET}" != "main" ]; then
```
nesence-m, line 23: **Major — this gate is PR-path only; the deploy path never runs it.** `gradle.properties` promises "CI hard-fails if it finds this non-blank" on maint/release/review branches, but this composite action is only invoked from `pull-request.yml` and `e2e-smoke.yml`. `deploy.yml` triggers on *push* to `release/**` — it runs the moment the release branch is pushed, independent of any PR — and contains no pin check; and since Gradle itself never reads `SDK_COMMIT_PIN`, a deploy from a tree with a live pin doesn't fail — it silently builds against the published Maven `ZCASH_SDK_VERSION`, i.e. **not** the SDK revision every green checkmark on the PR actually tested. An API-compatible SDK behavioral fix ships to Play with no error of any kind.

This is the same failure class as cutting an SDK release with a live `[patch.crates-io]` stanza, and the fix is about as cheap:

```yaml
- name: Refuse to deploy with a live SDK_COMMIT_PIN
  run: |
    if grep -qE '^SDK_COMMIT_PIN=.+' gradle.properties; then
      echo '::error::SDK_COMMIT_PIN is set — releases must build against a published SDK.'; exit 1
    fi
```

at the top of `deploy.yml` (and `release.yaml`/`create-release.yml` for symmetry).

.github/actions/setup-sdk-companion/action.yml
```diff
@@ -0,0 +1,87 @@
...
+    description: "'true' if the companion SDK build was activated this run, else empty."
+    value: ${{ steps.pin.outputs.commit != '' && 'true' || '' }}
+runs:
+  using: "composite"
+  steps:
+    - name: Read SDK_COMMIT_PIN and enforce the branch gate
+      id: pin
+      shell: bash
+      run: |
+        PIN="$(grep -m1 '^SDK_COMMIT_PIN=' gradle.properties | cut -d= -f2 | tr -d '[:space:]')"
```
nesence-m, line 20: Hardening: no format validation on the pin — a branch name here checks out fine and silently tracks a moving ref, defeating the entire point of a *commit* pin (same "branches move underneath you" principle as the SDK's Cargo rev-pin policy). One line after the read:

```bash
[ -z "${PIN}" ] || [[ "${PIN}" =~ ^[0-9a-f]{40}$ ]] || { echo '::error::SDK_COMMIT_PIN must be a full 40-char commit sha'; exit 1; }
```

(The current value is a full sha — good.)

gradle.properties
```diff
@@ -109,6 +109,17 @@ ZCASH_GOOGLE_PLAY_DEPLOY_STATUS=draft
...
+# this exact commit and builds against it (via SDK_INCLUDED_BUILD_PATH) instead of the
+# published Maven artifact — for co-developing an app change against unreleased SDK work,
+# without waiting on an SDK release. Locally this has no effect (use
+# ORG_GRADLE_PROJECT_SDK_INCLUDED_BUILD_PATH instead, see docs/CONTRIBUTING or CLAUDE.md).
+#
+# main and branches targeting main only. Must be blank on maint/release/review branches — CI
+# hard-fails if it finds this non-blank there, since those branches must always build against
+# an audited, published SDK version, never an arbitrary pinned commit. Clear this before
+# merging into maint.
+SDK_COMMIT_PIN=d7aa3423a9c62d84a1a2156c660fdf99fc29d266
```
nesence-m: **Blocking-trivial — stale pin.** `d7aa3423` is two commits behind the companion SDK head `88065245` (it's missing that PR's last two fixes), so the green checks on this PR don't cover the SDK revision that will actually merge. Please bump to the current head before merge — and once zcash/zcash-android-wallet-sdk#2165 lands, repoint to the **merged** sha (or clear the pin): `d7aa3423` is reachable only from the SDK's PR branch, and branch-only commits stop being reliably fetchable after the branch is deleted, which would quietly make old app commits unbuildable.

Also, minor: the pointer "see docs/CONTRIBUTING or CLAUDE.md" on L116 — neither file mentions `ORG_GRADLE_PROJECT_SDK_INCLUDED_BUILD_PATH` (grep-verified).

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/usecase/SubmitVotesUseCase.kt
```diff
@@ -524,14 +547,26 @@ class SubmitVotesUseCase(
...
+        // van_leaf_position isn't yet — the normal in-flight state while awaiting confirmation, not
+        // an edge case. Falling through is safe: buildDelegationProofIfNeeded no-ops for an
+        // already-SUBMITTED bundle (see its own phase check), and getDelegationSubmission(...)
+        // deterministically re-derives the same submission bytes from the already-persisted PCZT/
+        // proof state (the Keystone sig/sighash/rk equality asserts below rely on exactly this), so
+        // re-running the submit step below is an idempotent re-broadcast. It then proceeds to the
+        // full awaitTxConfirmation retry budget and storeVanPosition, instead of returning early and
+        // leaving van_leaf_position unset (which previously wedged the round downstream in
+        // submitFreshVoteBundle → syncVoteTree → generateVanWitnessJson).
+        if (currentDelegationPhase(dbHandle, roundId, bundleIndex) == DelegationPhase.CONFIRMED) {
```
nesence-m, line 573: The in-flight SUBMITTED fall-through is right, and I verified it end to end: no-op re-prove → deterministic re-derivation → idempotent re-broadcast → full `TX_CONFIRMATION_RETRIES × TX_CONFIRMATION_POLL_MS` budget → `requireAccepted` → `storeVanPosition`. That fixes the wedge I flagged.

But the **confirmed-as-failed sub-case from that comment is still unhandled**: cached-hash probe returns a confirmation with `code != 0` → `vanPosition` stays null (`takeIf { it.code == 0 }`) → falls through here (phase is SUBMITTED, not CONFIRMED) → re-derives the *same deterministic bytes* → the identical failing re-broadcast, on every retry, forever. Nothing ever clears `delegation_tx_hash` — no per-bundle clear primitive exists anywhere (grep: zero hits), which the commit message itself acknowledges. The wedge went from silent-and-downstream to visible-and-earlier; that's an improvement, not a fix.

Fine as a follow-up since it needs crate/SDK work — but please file the ticket and link it here so it doesn't evaporate when this thread closes.

nesence-m (changes requested):
Re-review at head `cbf4b20f`, all claims verified against the head tree. The fixes from the first pass are real: the MOB-1669 merge-clobber is fully restored (all four items, down to the `assertNotNull(pendingPczts.get(...))` assertion and the `coVerify(startLiveDriverImmediately = false)`); the bundled config source is content-addressed again (commit + checksum pinned together, mutable URL only in a comment); the strings.xml quotes are back in both locales; the cause-chain walk and changelog entries landed. The SUBMITTED early-return rework is correct for the in-flight case — full 45×2s confirmation budget, `requireAccepted`, `storeVanPosition` (see inline for the sub-case that remains).

The commit-pinned companion SDK mechanism is genuinely well built — single source of truth in `gradle.properties`, rides the existing `SDK_INCLUDED_BUILD_PATH` composite-build wiring with zero new Gradle code, exact-commit cache keying so a stale cache costs time but never a wrong-rev artifact, `persist-credentials: false` and sha-pinned actions throughout. Requesting changes for three targeted items, all cheap:

1. **Deploy-path pin guard** (inline on `setup-sdk-companion/action.yml`): the release-safety promise in `gradle.properties` is only enforced on the PR path; `deploy.yml` (push-triggered on `release/**`) never checks the pin and Gradle never reads it, so a release cut with a live pin silently ships against the last published Maven snapshot. Same failure class as releasing with a live `[patch.crates-io]` stanza; ~3-line fix.
2. **Stale pin** (inline on `gradle.properties`): CI currently validates against SDK `d7aa3423`, two commits behind the companion head `88065245` — the green checkmark doesn't cover what will actually merge.
3. **Wire `storeKeystoneSignature`** (existing thread on `VotingKeystoneRepository.kt`): the SDK primitive this fix was blocked on now exists and is verified working (the crate's preservation `WHERE` clause is real and crate-tested) — but the app has zero call sites; it isn't even surfaced on `VotingCryptoClient`. The round-wide reset still permanently orphans already-scanned Keystone signatures. With the blocker gone this is a small app-side change, and it's the worst reachable bug in the set.

Follow-up tier (not blocking, please track):
- **PR jobs still split across two SDK revisions**: `test_kotlin_modules`, `test_robo_debug`, and `test_robo_release`'s Flank config run Gradle without the companion action, i.e. against the Maven snapshot. When an app PR actually uses a new unpublished SDK API — the mechanism's whole purpose — those jobs fail to compile; when it's API-compatible, one green PR aggregates results from two different SDKs. Wire the action in or document the limitation.
- Pin hardening: no sha-format validation (a branch name would check out fine and track a moving ref — `[[ "$PIN" =~ ^[0-9a-f]{40}$ ]]`); after the SDK PR merges, repoint to the merged sha or clear promptly (branch-only shas stop being reliably fetchable after branch deletion); the doc pointer "see docs/CONTRIBUTING or CLAUDE.md" references docs that don't mention `ORG_GRADLE_PROJECT_SDK_INCLUDED_BUILD_PATH`.
- The stale-proofs pairing (my other open thread) stays a documented gap — the KDoc that landed SDK-side is accurate, and the fix genuinely needs crate work; keeping the thread open as the tracker.
- Cleanup tier from the first pass: all still open (settled-predicate now at 3 copy-pasted sites, overwrite-swallow at 3, duplicated 9-arg `buildGovernancePczt`, `delegationPhases` re-fetched up to 3× per bundle, dead `hasVoteReady`).

CI at this head, for the record: `test_robo_release` (Flank credentials NPE + 25-min timeout) and `e2e-android (phase 2)` (restore-wallet never reaches `home.syncComplete`) fail identically on `feature/slipstream-module-sdk`, so they're fleet/infra, not this PR. The 2/12 `ScreenshotTest` failures in `test_android_modules_emulator (app)` look like emulator flake (ColorBuffer/snapshot errors; passes on the sibling) — worth a re-run to confirm before merge.

🤖 Review prepared with [Claude Code](https://claude.com/claude-code); all findings human-triaged.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/repository/VotingKeystoneRepository.kt
nesence-m: Re-verified at head `cbf4b20f`: still zero call sites for the SDK's `storeKeystoneSignature` — it isn't surfaced on `VotingCryptoClient` at all — and the reset at ~L223 is still round-wide with nothing persisted crate-side first. Meanwhile I reviewed the SDK-side primitive in #2165 and it checks out: the crate's preservation `WHERE ... bundle_index NOT IN (SELECT ... FROM keystone_signatures)` is real and covered by the crate's own test, and the JNI adds stricter-than-crate validation. So the cited blocker is gone; this is now a small app-side change (surface it on `VotingCryptoClient`, call it in `storeBundleSignature` before any reset can run). This is ask #3 of my re-review — keeping the thread open until wired.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/usecase/SubmitVotesUseCase.kt
nesence-m: Re-verified at head `cbf4b20f`: confirmed doc-only, and the KDoc that landed SDK-side is accurate (nothing clears `proofs`; the only clearing helper in the crate is `pub(crate)`). The `setupJustBuilt` guard still can't reach the Keystone branch, so the stale-proof pairing remains reachable there. Accepting doc-only for this PR since the fix genuinely needs crate work — but this and the confirmed-as-failed `delegation_tx_hash` wedge (my comment on `SubmitVotesUseCase`) need the same crate-side clear/invalidate primitive, so please track them under one follow-up ticket. Keeping the thread open as the tracker.

noop-sk: Filed: https://linear.app/zodl/issue/MOB-1699 — tracks both this and the delegation_tx_hash wedge below under one ticket, since both need the same crate-side clear/invalidate primitive.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/usecase/SubmitVotesUseCase.kt
noop-sk: Filed: https://linear.app/zodl/issue/MOB-1699

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/repository/VotingKeystoneRepository.kt
noop-sk: Wired up in 5bae4fcb8.

Surfaced `storeKeystoneSignature` on `VotingCryptoClient` (interface + impl, mirroring `recordShareDelegation`'s `withContext(Dispatchers.IO) { session(dbHandle).storeKeystoneSignature(...) }` pattern):

```kotlin
suspend fun storeKeystoneSignature(
    dbHandle: Long,
    roundId: String,
    bundleIndex: Int,
    keystoneSig: ByteArray,
    keystoneSighash: ByteArray,
    rk: ByteArray
)
```

Called from `VotingKeystoneRepositoryImpl.storeBundleSignature`, right after `extractSpendAuthSignatureFromSignedPczt` and before the local `votingRecoveryRepository.storeKeystoneBundleSignature` write (so the crate-side guard is populated first — protected even if the local write never happens). Values passed are the same triple already used app-side: `spendAuthSig` (extracted from the signed PCZT), `sighash` (checked against the pending request's expected sighash just above), and `pendingRequest.decodeExpectedRk()` — the crate-computed `rk` from the governance PCZT this signature was produced for.

One nuance worth flagging: the SDK KDoc says to pass the `rk`/`sighash` pair that `getDelegationSubmissionWithKeystoneSig` already verified the signature against. That call only becomes possible once a delegation proof exists (built later, asynchronously, via `buildDelegationProofIfNeeded`), so it isn't available yet at `storeBundleSignature` time (right after scanning). I used the best available proxy — the crate's own PCZT-computed rk/sighash, already sighash-checked via `rejectMismatchedKeystoneSighash` — and left a comment noting that `SubmitVotesUseCase`'s later `getDelegationSubmissionWithKeystoneSignature` call independently re-asserts this same triple against the crate's submission result, so a wrong value here would surface loudly at submission time rather than silently corrupt anything. Also handles the legacy case where a resumed pending request has no persisted `rk` (nullable field) by skipping the crate-side call gracefully — the local recovery-repository record is unaffected either way.

Added two new unit tests to `VotingKeystoneRepositoryTest` covering the crate-side call (args + open/close db lifecycle) and the null-rk skip path. Verified with `:feature-voting:compileZcashmainnetStoreDebugKotlin` and `:feature-voting:testZcashmainnetStoreDebugUnitTest` against the composite SDK build (both green).

.github/actions/setup-sdk-companion/action.yml
noop-sk: Fixed in ec31d5dd9.

Added an early "Refuse to deploy with a live SDK_COMMIT_PIN" guard step to `deploy.yml` (in the `build` job, right after checkout, before `Set up Java`), `release.yaml` (in the `release` job, right after checkout, before the AWS credentials step), and `create-release.yml` (right after "Checkout tag", before parsing the changelog) — using your exact snippet. All three now hard-fail before any build/release work if `SDK_COMMIT_PIN` is non-blank in `gradle.properties`, closing the gap where the deploy path could silently fall back to the published Maven SDK.

Validated all three with `python3 -c "import yaml; yaml.safe_load(...)"` and dry-ran the guard's bash logic locally against the current `gradle.properties` (which does still have a live pin, pending separate cleanup) — confirms it correctly detects and would fail on it.

.github/actions/setup-sdk-companion/action.yml
noop-sk: Fixed in ec31d5dd9.

Added the format validation to the existing "Read SDK_COMMIT_PIN and enforce the branch gate" step, right after the `PIN="$(grep ...)"` line and before the branch-gate check, using your exact snippet:

```bash
[ -z "${PIN}" ] || [[ "${PIN}" =~ ^[0-9a-f]{40}$ ]] || { echo '::error::SDK_COMMIT_PIN must be a full 40-char commit sha'; exit 1; }
```

Validated YAML syntax, and dry-ran the check locally against the current (valid, 40-char) pin — passes without a false positive — and against a synthetic branch-name-shaped value — correctly rejects with exit 1.

gradle.properties
noop-sk: Bumped to `e177a881` (commit `fd269de53`) — the SDK head after its third review round (all 4 comments there fixed). Also fixed the stale docs/CONTRIBUTING-or-CLAUDE.md pointer and expanded the comment to note the pin must be repointed to a merged commit on SDK main once #2165 lands, since a PR-branch-only sha stops being fetchable once that branch is deleted.

feature-voting/src/main/java/co/electriccoin/zcash/ui/common/usecase/SubmitVotesUseCase.kt
noop-sk: Fixed in c479ef368.

Root cause confirmed as described: `resetVotingSessionState` only clears the `bundles` table, never `proofs`, and `DelegationPhase` treats a `proofs` row's mere existence as PROVED with no check against the current `alpha`. `setupJustBuilt` already handled this for software wallets, but nothing set it for Keystone, since the reset-and-rebuild-on-overwrite-refusal recovery lives entirely inside `VotingKeystoneRepository.createPcztEncoder`, invisible to `SubmitVotesUseCase`.

Fix (app-side only, per the agreed direction):

- `VotingRecoverySnapshot` (`VotingRecoveryRepository.kt`) gains a new persisted field `rebuiltSinceProofBundles: Set<Int>`, plus two new repository methods following the existing `keystoneBundleSignatures`/`submittedProposalIds` mutation conventions:
  - `suspend fun markBundleRebuiltSinceProof(accountUuid: String, roundId: String, bundleIndex: Int)`
  - `suspend fun clearBundleRebuiltSinceProof(accountUuid: String, roundId: String, bundleIndex: Int)`
  Backed by `VotingRecoverySnapshotExt.withBundleRebuiltSinceProof`/`withBundleRebuiltSinceProofCleared`, and wired into the existing JSON encode/decode for the snapshot's persisted storage - it's persisted, not in-memory, since process death is the actual trigger scenario here.
- `VotingKeystoneRepository.createPcztEncoder`'s reset-and-rebuild recovery block calls `markBundleRebuiltSinceProof` right after the rebuilt `buildGovernancePczt` call succeeds (i.e. once a fresh `alpha` is actually in hand).
- `SubmitVotesUseCase`'s `alreadyProved` computation now also requires `bundleIndex !in context.recovery.rebuiltSinceProofBundles`, and clears the flag via `clearBundleRebuiltSinceProof` right after a fresh proof is generated for that bundle.

Also corrected the misleading comment above the recovery block in `VotingKeystoneRepository.kt` - it previously said "(e.g. reinstall)"; that's not actually reachable, since the crate's SQLite DB and the app's `VotingRecoveryRepository` cache live in the same app-private storage and get wiped symmetrically by a real reinstall. The real trigger is Android process death between `buildGovernancePczt`'s synchronous native-side commit and the later `storePendingKeystoneRequest` call that persists the app's own redacted-PCZT cache.

Added tests: `VotingKeystoneRepositoryTest.resetAndRebuildPathMarksBundleRebuiltSinceProof` (confirms `markBundleRebuiltSinceProof` fires on the reset-and-rebuild path) and two new cases in `VotingRecoverySnapshotExtTest` for the new extension functions.

Verified with a real build against the composite SDK (`:feature-voting:compileZcashmainnetStoreDebugKotlin`, `:feature-voting:testZcashmainnetStoreDebugUnitTest`, `ktlintFormat`, `detektAll`) - all clean for the touched files.

nesence-m (approved):
Lgtm

PR #2407, opened 2026-08-10
[MOB-1679] Add manual Firebase App Distribution deploy workflow

nesence-m:
Adds a manually-triggered GitHub Actions workflow that deploys chosen app variants to Firebase App Distribution, so any branch can be shipped to testers from the Actions tab instead of from someone's laptop via the local `deploy.sh`.

Linear: [MOB-1679](https://linear.app/zodl/issue/MOB-1679/manual-firebase-app-distribution-deploy-workflow-branch-variant-picker) (implementation sub-issue: [MOB-1680](https://linear.app/zodl/issue/MOB-1680/add-firebase-app-distributionyml-workflow-ci-docs))

## What the workflow does

`workflow_dispatch` only. The branch comes from the standard "Run workflow" branch picker; the inputs are:

| Input | Required | Default | Meaning |
|---|---|---|---|
| `variants` | yes | `zcashmainnetInternalRelease,zcashtestnetFossDebug` | `all`, or a comma-separated subset of `zcashtestnetStoreDebug`, `zcashmainnetStoreDebug`, `zcashmainnetFossDebug`, `zcashtestnetFossDebug`, `zcashmainnetStoreRelease`, `zcashtestnetStoreRelease`, `zcashmainnetInternalRelease` |
| `groups` | no | `ec,zodl` | Firebase App Distribution tester groups (comma-separated); the default matches the local `deploy.sh` |
| `release_notes` | no | *(empty)* | Empty auto-generates from variant, branch and short SHA |
| `version_code` | no | *(empty)* | Empty uses the build's own default (git commit count of the selected branch) |

Jobs: `validate_gradle_wrapper` → `prepare_matrix` → `build_and_deploy` (matrix) → `notify_slack`.

- `prepare_matrix` validates the `variants` input against the seven distributable variants and expands it into a JSON matrix; an unknown name fails the run immediately and prints the valid values.
- `build_and_deploy` runs one matrix job per selected variant with `max-parallel: 1`, so the variants build **sequentially**, one runner at a time; `fail-fast: false` keeps a failing variant from cancelling the rest. The workflow-level `concurrency` group only serializes separate dispatches. Per-variant Firebase app-ID mapping is identical to the local `deploy.sh`. All distributed builds — debug **and** release build types — are signed with the shared debug keystore (`FIREBASE_DEBUG_KEYSTORE_BASE64`), matching how locally-distributed builds have been signed so far, so testers keep a stable signature and can update-install; the job fails fast if the keystore secret is missing. Each APK is uploaded as a run artifact and then distributed with `firebase-tools@15.26.0 appdistribution:distribute`.
- `notify_slack` is best-effort: `if: always()`, `continue-on-error: true`, and it exits cleanly when the webhook secret is unset, so it can never fail the run. It posts status, branch, variants, groups, actor and the run link to the wallet-team channel.

`docs/CI.md` gains a "Firebase App Distribution deployment" section covering the trigger, inputs and secrets.

## Known behavior change vs. the local `deploy.sh`

The workflow performs **no already-uploaded/dedupe check** before distributing: re-dispatching the workflow (or re-running failed jobs) for the same commit uploads a new release and notifies testers again. This is deliberate — CI builds are not byte-reproducible, so a re-run produces a genuinely new binary, and a duplicate distribution is cheaper than a silently skipped one.

## New secrets required

Provisioning is tracked in [MOB-1681](https://linear.app/zodl/issue/MOB-1681/provision-firebase-app-distribution-service-account-secrets-in-zashi) (assigned to Yasser); the deploy step fails for every variant until the three Firebase secrets exist.

- `FIREBASE_APP_DIST_DEBUG_KEY_BASE64` — `firebase-distribution` environment
- `FIREBASE_APP_DIST_FOSS_KEY_BASE64` — `firebase-distribution` environment
- `FIREBASE_APP_DIST_RELEASE_KEY_BASE64` — `firebase-distribution` environment
- `FIREBASE_DEBUG_KEYSTORE_BASE64` — `firebase-distribution` environment; shared debug keystore that signs every distributed build
- `SLACK_WEBHOOK_URL_WALLET_TEAM` — repository-level, optional (only the completion notification is skipped without it)

The `firebase-distribution` environment does not exist yet and must be created **with no deployment branch policy** (also tracked in [MOB-1681](https://linear.app/zodl/issue/MOB-1681/provision-firebase-app-distribution-service-account-secrets-in-zashi)); the existing `Deployment` environment restricts jobs to `main`, `release/*` and release tags, which would block dispatching this workflow for a feature branch.

All other secrets the workflow references already exist for `deploy.yml`.

## Availability note

GitHub only exposes a `workflow_dispatch` workflow in the Actions UI when the file is present on the **default branch**. This PR targets `maint/v3.9.x`, so the "Run workflow" button appears only once that branch is merged forward into `main` through the normal maintenance merge-forward.

---

> **Note**

# Author

> **Note**

# Reviewer


noop-sk (commented):
Reviewed against our internal design notes for this exact conversion (deploy.sh → GitHub Action). Two findings below are load-bearing for the stated goal of this PR ("any branch can be shipped to testers"); the rest are should-consider/nitpick. Also flagging some things this PR does *better* than our own draft design, for the record.

.github/workflows/firebase-app-distribution.yml
```diff
@@ -0,0 +1,316 @@
...
+              exit 1
+            fi
+          fi
+          VARIANTS_JSON=$(printf '%s\n' "${SELECTED[@]}" | jq -R . | jq -cs .)
+          echo "variants=${VARIANTS_JSON}" >> "$GITHUB_OUTPUT"
+          echo "Selected variants: ${SELECTED[*]}"
+
+  build_and_deploy:
+    needs: [validate_gradle_wrapper, prepare_matrix]
+    environment: deployment
```
noop-sk: **Critical: `environment: deployment` defeats the "run from any branch" goal this PR exists for.**

I checked live via `gh api repos/zodl-inc/zodl-android/environments/deployment/deployment-branch-policies`: this environment's branch policy is `main`, `release/*.*.*`, plus a few tag patterns. Since `build_and_deploy` (the whole job, not just the release-variant legs) declares this environment, GitHub will refuse to run it at all for any ref outside that policy — i.e. exactly the case this PR's own description calls out as the point of the workflow ("so any branch can be shipped to testers"). The PR body correctly documents a different, unrelated GitHub limitation (workflow_dispatch files must live on the default branch to show the Run workflow button), but doesn't address this one.

Suggest a second environment (e.g. `firebase-distribution`) with no deployment branch policy, reusing the same secret names as `deployment` so key rotation doesn't have to touch two places with different names.

.github/workflows/firebase-app-distribution.yml
```diff
@@ -0,0 +1,316 @@
...
+        env:
+          FIREBASE_DEBUG_JSON_BASE64: ${{ secrets.FIREBASE_DEBUG_JSON_BASE64 }}
+          FIREBASE_RELEASE_JSON_BASE64: ${{ secrets.FIREBASE_RELEASE_JSON_BASE64 }}
+        if: "${{ env.FIREBASE_DEBUG_JSON_BASE64 != '' && env.FIREBASE_RELEASE_JSON_BASE64 != '' }}"
+        shell: bash
+        run: |
+          mkdir -p app/src/debug/
+          mkdir -p app/src/release/
+          echo ${FIREBASE_DEBUG_JSON_BASE64} | base64 --decode > app/src/debug/google-services.json
+          echo ${FIREBASE_RELEASE_JSON_BASE64} | base64 --decode > app/src/release/google-services.json
```
noop-sk, line 217: **High: this will very likely break the FOSS variants.** This step decodes `FIREBASE_DEBUG_JSON_BASE64`/`FIREBASE_RELEASE_JSON_BASE64` unconditionally for every matrix leg, including `zcashmainnetFossDebug`/`zcashtestnetFossDebug`.

`app/build.gradle.kts:17-34` applies the `com.google.gms.google-services` plugin to the whole `app` module (not per-variant) once both files exist, and its per-variant `processXxxGoogleServices` task fails the build with "No matching client found for package name" for any applicationId missing from the JSON. FOSS runs under a third, separate Firebase project (project number `228274394465`, vs. `72075012504`/`1004234202961` for debug/release) — these two secrets almost certainly have no client entry for the FOSS package.

This isn't hypothetical: `scripts/prepare-foss-build.sh:8-9` already deletes exactly these two files (and goes further, stripping the google-services/firebase plugin lines from build files) before FOSS builds elsewhere in this repo's CI (`release.yaml`). This workflow doesn't call that script or otherwise special-case FOSS.

Suggest gating this step with `if: ${{ !contains(matrix.variant, 'Foss') }}`.

.github/workflows/firebase-app-distribution.yml
```diff
@@ -0,0 +1,316 @@
...
+          mkdir -p "$HOME/.android"
+          echo "$DEBUG_KEYSTORE_BASE64" | base64 --decode > "$HOME/.android/debug.keystore"
+      - name: Build variant
+        timeout-minutes: 30
+        env:
+          VARIANT: ${{ matrix.variant }}
+          VERSION_CODE_OVERRIDE: ${{ inputs.version_code }}
+          # Signs release build types with the shared debug keystore installed above,
+          # matching how locally-distributed builds have been signed so far.
+          ORG_GRADLE_PROJECT_IS_SIGN_RELEASE_BUILD_WITH_DEBUG_KEY: true
```
noop-sk, line 245: Worth double-checking the claim in the comment above ("matching how locally-distributed builds have been signed so far") against reality: `deploy.sh` itself never sets any `ZCASH_RELEASE_KEYSTORE_*`/`IS_SIGN_RELEASE_BUILD_WITH_DEBUG_KEY` — whatever it produced locally depended entirely on whoever ran it having (or not having) the real release keystore configured globally in their own `~/.gradle/gradle.properties`. If any local release-variant testers were actually signed with the real production key before, they won't be able to update-install over a build signed with this shared debug keystore (signature mismatch) — worth confirming that's an accepted/known transition rather than an assumption baked into the comment.

Separately: this is a nice piece of design regardless — by never touching the real release signing key at all, it sidesteps a risk we flagged in our own earlier draft of this workflow (any-branch dispatch + real release key in env == anyone with dispatch rights can exfiltrate the production signing key via a crafted branch). This approach closes that off entirely instead of just mitigating it. Good call.

.github/workflows/firebase-app-distribution.yml
```diff
@@ -0,0 +1,316 @@
...
+            exit 1
+          fi
+          if [ -z "$RELEASE_NOTES" ]; then
+            RELEASE_NOTES="CI build of ${VARIANT} from ${GITHUB_REF_NAME} (${GITHUB_SHA::7})"
+          fi
+          DISTRIBUTE_ARGS=(appdistribution:distribute "$APK_PATH" --app "$APP_ID" --release-notes "$RELEASE_NOTES")
+          if [ -n "$FIREBASE_GROUPS" ]; then
+            DISTRIBUTE_ARGS+=(--groups "$FIREBASE_GROUPS")
+          fi
+          npx --yes firebase-tools@14 "${DISTRIBUTE_ARGS[@]}"
```
noop-sk: `deploy.sh` checks Firebase App Distribution for an existing release with the same `buildVersion` before building/uploading, and skips if found. This port has no equivalent — re-running this workflow (e.g. GitHub's "Re-run failed jobs", or two dispatches for the same commit) will just upload a duplicate release and re-notify testers rather than skipping cleanly. `concurrency: firebase-app-distribution` (line 46) only serializes *overlapping* dispatches, it doesn't prevent a second, later run from re-uploading the same version. Worth deciding if that's an acceptable simplification (firebase-tools makes doing this idempotently non-trivial) or should be ported over.

Minor/unrelated: `firebase-tools@14` pins only the major version, so the exact CLI build isn't reproducible between two runs a month apart — probably fine, just flagging.

.github/workflows/firebase-app-distribution.yml
```diff
@@ -0,0 +1,316 @@
...
+
+  notify_slack:
+    # Best-effort notification: continue-on-error keeps a Slack failure from
+    # failing the workflow run, and the step exits cleanly when the webhook
+    # secret is not configured.
+    needs: [prepare_matrix, build_and_deploy]
+    if: ${{ always() && needs.build_and_deploy.result != 'skipped' }}
```
noop-sk: Should-consider: this condition means `notify_slack` never fires if `validate_gradle_wrapper` or `prepare_matrix` fail (e.g. a typo'd `variants` input, or a tampered gradle wrapper on some branch) — those failures are only visible in the Actions UI, not in `#wallet-team`. Might be worth reporting on those too, not just once `build_and_deploy` has actually run.

.github/workflows/firebase-app-distribution.yml
nesence-m: Good catch, and confirmed live: `gh api repos/zodl-inc/zodl-android/environments` shows the sole environment `Deployment` carries branch policies (`main`, `release/*.*.*`, tags `v*.*.*` / `*.*-*` / `*.*.*-*`), so the job would have been rejected for any feature branch — exactly what this workflow exists to support.

Fixed in ef37030: the job now uses a dedicated `firebase-distribution` environment, which must be created with **no** deployment branch policy. That keeps the branch protection on `Deployment` (Play deploys) intact instead of loosening it. Creating the environment and moving the secrets into it needs admin rights, so it is tracked in MOB-1681 alongside the secret provisioning; `docs/CI.md` and the workflow header now state the no-branch-policy requirement and why.

.github/workflows/firebase-app-distribution.yml
nesence-m: Confirmed and fixed in ef37030. `app/build.gradle.kts` applies `com.google.gms.google-services` (and Crashlytics) module-wide as soon as both `app/src/debug/google-services.json` and `app/src/release/google-services.json` exist, so `process*GoogleServices` would then fail for the FOSS applicationIds (different Firebase project, `228274394465`).

The export step now carries `&& !contains(matrix.variant, 'Foss')`, so FOSS variants build without the files — the same thing `scripts/prepare-foss-build.sh` does (it deletes exactly these two files) before FOSS builds in `release.yaml`. Documented in `docs/CI.md`.

.github/workflows/firebase-app-distribution.yml
nesence-m: Verified, the claim holds. All local Firebase distribution to date has run from a single machine, whose `~/.gradle/gradle.properties` sets `IS_SIGN_RELEASE_BUILD_WITH_DEBUG_KEY=true` with no release keystore configured — so every previously distributed build, debug and release alike, is signed by that machine's `~/.android/debug.keystore`. `FIREBASE_DEBUG_KEYSTORE_BASE64` is that exact file, which is what makes the transition update-installable for existing testers rather than a wipe.

Agreed on your point about isolation from the Play signing key: this keystore is deliberately the throwaway debug one and never touches the Play upload/release key. I have expanded the comment above `ORG_GRADLE_PROJECT_IS_SIGN_RELEASE_BUILD_WITH_DEBUG_KEY` in ef37030 to state that provenance explicitly instead of the vague "so far".

.github/workflows/firebase-app-distribution.yml
nesence-m: Dropping the duplicate-version skip is deliberate. An idempotent pre-check would need a raw OAuth call against the App Distribution API (the gcloud path `deploy.sh` uses is not worth porting into the workflow), and it would buy little: App Distribution deduplicates byte-identical uploads by content hash, and CI rebuilds are not byte-reproducible anyway, so a re-dispatch of the same commit produces a genuinely new binary that testers should get. A duplicate distribution is cheap; a silently skipped one is confusing.

The version pin is a fair hit, though — fixed in ef37030: `firebase-tools@14` → `firebase-tools@15.26.0`, exact, so a CLI release cannot change deploy behaviour underneath us. Bumping it is now an explicit commit.

.github/workflows/firebase-app-distribution.yml
nesence-m: Right — with `needs.build_and_deploy.result != 'skipped'` a failure in `validate_gradle_wrapper` or `prepare_matrix` (e.g. a typo in the `variants` input) produced no Slack message at all. Fixed in ef37030: `needs: [validate_gradle_wrapper, prepare_matrix, build_and_deploy]` with plain `if: ${{ always() }}`, so every dispatch notifies.

A skipped deploy now reports `:x: ... failed before building (gradle wrapper validation: <result>, variant parsing: <result>)`, and the variant list falls back to `<input> (requested)` when `prepare_matrix` never produced a matrix. `continue-on-error: true` and the unset-webhook early exit are unchanged, so the notification still cannot fail the run.

.github/workflows/firebase-app-distribution.yml
```diff
@@ -0,0 +1,338 @@
...
+            exit 1
+          fi
+          if [ -z "$RELEASE_NOTES" ]; then
+            RELEASE_NOTES="CI build of ${VARIANT} from ${GITHUB_REF_NAME} (${GITHUB_SHA::7})"
+          fi
+          DISTRIBUTE_ARGS=(appdistribution:distribute "$APK_PATH" --app "$APP_ID" --release-notes "$RELEASE_NOTES")
+          if [ -n "$FIREBASE_GROUPS" ]; then
+            DISTRIBUTE_ARGS+=(--groups "$FIREBASE_GROUPS")
+          fi
+          npx --yes firebase-tools@15.26.0 "${DISTRIBUTE_ARGS[@]}"
```
noop-sk, line 291: No dedupe/already-uploaded check before calling `appdistribution:distribute`, unlike the local `deploy.sh` this replaces (which does a check-then-act dedupe against Firebase before uploading). The workflow-level `concurrency: firebase-app-distribution` group prevents two *simultaneous* dispatches from double-uploading, but it doesn't prevent an intentional or mistaken re-dispatch of the same variant/branch from posting a duplicate release + duplicate tester notification. Worth a one-line callout in the PR description as a known behavior change, or adding the dedupe check back if it matters for this team's workflow.

.github/workflows/firebase-app-distribution.yml
```diff
@@ -0,0 +1,338 @@
...
+    inputs:
+      variants:
+        description: "Variants to build and upload: 'all' or a comma-separated subset of: zcashtestnetStoreDebug, zcashmainnetStoreDebug, zcashmainnetFossDebug, zcashtestnetFossDebug, zcashmainnetStoreRelease, zcashtestnetStoreRelease, zcashmainnetInternalRelease"
+        required: true
+        default: 'zcashmainnetInternalRelease,zcashtestnetFossDebug'
+        type: string
+      groups:
+        description: 'Firebase App Distribution tester groups (comma-separated)'
+        required: false
+        default: 'zodl'
```
noop-sk: Default here is `zodl` only. Worth confirming against the original `deploy.sh`/manual agent flow — if that defaulted to `ec, zodl` for `all`-variant runs, the `ec` group would now silently stop receiving builds unless someone remembers to type it into this input.

noop-sk (commented):
A few things worth double-checking before merge (not blockers, workflow logic itself checks out against `app/build.gradle.kts` — versionCode fallback and debug-key signing chain both resolve correctly).

.github/workflows/firebase-app-distribution.yml
nesence-m: Good catch — checked against the original `deploy.sh`: it defaults to `FIREBASE_GROUPS="ec, zodl"`, so a `zodl`-only default would indeed have silently dropped the `ec` group. Fixed in 345c2bc: the `groups` input now defaults to `ec,zodl`, and `docs/CI.md` states the default and its provenance.

.github/workflows/firebase-app-distribution.yml
nesence-m: Agreed it deserves an explicit callout rather than living only in a resolved thread. Done in 345c2bc: a comment on the deploy step in the workflow states that the dedupe check was deliberately not ported and why (re-dispatch = new upload + re-notification; CI builds are not byte-reproducible, so a re-run is a genuinely new binary), `docs/CI.md` documents the same behavior change, and the PR description now lists it under a known-behavior-changes note.

PR #2412, opened 2026-08-10
Release 3.9.2

### LINKED ISSUE #2411: Release 3.9.2
## Release Checklist

- [x] Version bump
- [x] Changelog update
- [x] Fastlane files
- [ ] PR merged
- [ ] Build deployed to Google Play

noop-sk:
Closes #2411

### Fixed:

- Closing the error sheet during Keystone migration signing no longer discards already-signed rounds.

### Fixed:

- Transferring to a new phone via device-to-device setup no longer copies the app's local data, including the undecryptable encrypted preferences
  file, which previously forced a manual seed re-restore on the new device. Only the address book and metadata backups are carried over, matching
  cloud backup.
- Preference file creation no longer serializes behind one global lock: a slow or hung encrypted-preferences creation (e.g. blocked on the Keystore)
  no longer delays every other preference file, including the ones startup flows like the wallet secret and home state depend on. Preference-provider
  instances are now correctly cached instead of being rebuilt on every access, and steady-state preference reads/writes no longer take a lock at all.
- Opening the automatic migration progress screen no longer shows a blank screen while it loads. The migration schedule now appears immediately and a
  loading indicator is shown on screens that are still waiting for their content.
- Automatic migration no longer gets stuck if a scheduled transfer step fails transiently. The background driver now retries instead of silently
  stopping, which previously required backgrounding and re-foregrounding the app to recover.
- Fixed a back-navigation bounce through the Battery/Notification permission screens when returning from Review - pressing Back could land you
  several screens further back than expected.

### Added:

- We added Move to Ironwood, a guided migration that moves your Orchard funds into the new Ironwood shielded pool.
- We added two ways to migrate: a private option that splits your balance into smaller transfers sent automatically in the background over time, or an
  immediate single transfer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+### Fixed:
+
+- Closing the error sheet during Keystone migration signing no longer discards already-signed rounds.
+
 ## [3.9.1 (2361)] - 2026-08-08
 
 ### Fixed:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+## [3.9.2 (2370)] - 2026-08-10
+
 ### Fixed:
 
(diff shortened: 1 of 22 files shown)


zcash-harry (approved):
tACK
Lgtm after CI green

PR #2416, opened 2026-08-12
ci: extract SDK_COMMIT_PIN commit-pinned companion SDK build

noop-sk:
## Summary

Extracts the `SDK_COMMIT_PIN` CI capability from `main`'s PR #2406 onto the `maint/v3.9.x` line, cleanly separated from that PR's unrelated voting-module + swap refactor work, and from two other CI additions that happened to land in the same window (`auto-merge.yml`, `sync-fdroid-metadata.yml`) — neither of those is part of this feature and neither is included here.

`SDK_COMMIT_PIN` is an optional `gradle.properties` value: when set to a full 40-char commit sha, CI checks out that exact commit of `zcash/zcash-android-wallet-sdk`, builds it as a companion Gradle included build, and runs PR checks against it instead of the app's normal published Maven SDK snapshot. This lets an app PR be co-developed against unreleased SDK work without waiting on a Maven publish.

**Design update:** the branch gate now allows a live pin on `main` **and `maint/*`** branches (previously `main`-only), since maint APP lines legitimately need to co-develop against maint SDK lines. `release/*` and `review/*` remain hard-blocked — those must always build against a real published SDK version — enforced both in the PR-check gate and, independently, in `create-release.yml`, `deploy.yml`, and `release.yaml`, which each refuse to deploy/release if `SDK_COMMIT_PIN` is non-blank, regardless of branch.

**This PR now sets a live pin**, rather than leaving it blank: `SDK_COMMIT_PIN` is set to `91cff0936f45d1703ed7df39f347485894ac3103`, the current HEAD of `zcash/zcash-android-wallet-sdk`'s `maint/v3.0.x` branch — the SDK line `maint/v3.9.x` actually depends on (`ZCASH_SDK_VERSION=3.0.1-SNAPSHOT` in `gradle.properties`). This means this PR's own CI run exercises the real companion-SDK build path end to end (checkout, Rust/NDK setup, composite build), not just the infra with the pin left off as originally described.

## What's included

- New `.github/actions/setup-sdk-companion/action.yml` — composite action that reads `SDK_COMMIT_PIN`, enforces the branch gate (`main` or `maint/*`, else hard-fail), and (only if a pin is set) checks out the companion SDK commit, sets up its Rust/NDK toolchain, and wires `ORG_GRADLE_PROJECT_SDK_INCLUDED_BUILD_PATH`.
- `gradle.properties` — `SDK_COMMIT_PIN` property, set to `91cff0936f45d1703ed7df39f347485894ac3103` (SDK `maint/v3.0.x` HEAD as of this PR), with explanatory comment updated for the `maint/*`-allowed gate.
- `.github/workflows/pull-request.yml` — wires the companion-SDK setup step into lint/test/build jobs, bumps their timeouts to accommodate a cold Rust build when a pin is active, adds an app-shard pre-build step to avoid emulator contention, and adds an OOM-retry-once wrapper for the Store APK build.
- `.github/workflows/e2e-smoke.yml` — wires the companion-SDK setup step into the build job, bumps its timeout accordingly.
- `.github/workflows/create-release.yml`, `.github/workflows/deploy.yml`, `.github/workflows/release.yaml` — each adds a "Refuse to deploy with a live SDK_COMMIT_PIN" hard-fail step, unconditional on branch, so a pin can never silently leak into a real release/deploy.

## Why onto maint/v3.9.x

This capability landed on `main` bundled inside PR #2406, which did not merge into `maint/v3.9.x`. This PR brings the infra onto the maint line for consistency with `main`, and — per the updated design — actually activates it here so `maint/v3.9.x` can build/test against unreleased `maint/v3.0.x` SDK work, matching how `main` co-develops against SDK `main`.

## Test plan

- [x] `git diff --stat` confirms exactly the expected files changed: new `.github/actions/setup-sdk-companion/action.yml`; edits to `.github/workflows/{create-release.yml,deploy.yml,e2e-smoke.yml,pull-request.yml,release.yaml}` and `gradle.properties`.
- [x] All changed/added `.yml`/`.yaml` files pass `yaml.safe_load`.
- [x] Re-verified `zcash-android-wallet-sdk`'s `maint/v3.0.x` HEAD via `git ls-remote` immediately before pinning: `91cff0936f45d1703ed7df39f347485894ac3103`.
- [ ] CI passes on this PR with the pin live, targeting `maint/v3.9.x` — exercises the real companion-SDK build path (checkout + Rust/NDK + composite build), not just the fast no-op path.

### DIFF
--- a/.github/actions/setup-sdk-companion/action.yml
+++ b/.github/actions/setup-sdk-companion/action.yml
@@ -4,8 +4,9 @@ description: >-
   zcash/zcash-android-wallet-sdk, installs its Rust/NDK toolchain, and points Gradle's
   composite build at it via ORG_GRADLE_PROJECT_SDK_INCLUDED_BUILD_PATH — so a co-developed
   app PR can build/lint/test against unreleased SDK changes without waiting on a Maven
-  publish. No-ops (does nothing) whenever the pin is blank, which is the default, normal
-  case for every branch that isn't actively co-developing against an SDK PR.
+  publish. Supported on main, maint/* branches, and branches targeting either. No-ops (does
+  nothing) whenever the pin is blank, which is the default, normal case for every branch
+  that isn't actively co-developing against an SDK PR.
 outputs:
   active:
     description: "'true' if the companion SDK build was activated this run, else empty."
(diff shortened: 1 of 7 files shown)


nesence-m (approved):
utACK

PR #2408, opened 2026-08-10
Merge maint/v3.9.x into main

nesence-m:
Forward merge of the `maint/v3.9.x` maintenance line into `main`: fixes land on the oldest supported `maint/` line and are merged forward, never backward.

## What this forwards

`maint/v3.9.x` already contains `release/v3.9.0`, `release/v3.9.1` and `release/v3.9.2`, and through `maint/v3.8.x` also `release/v3.8.1`. All of those release→maint and maint→maint merges were verified already contained (`git merge-base --is-ancestor`) before starting, so `maint/v3.9.x` → `main` was the only merge outstanding. Before this PR, `maint/v3.9.x` was 299 commits ahead of `main` and `main` 42 ahead of `maint`.

## :rotating_light: Merge with a merge commit — do not squash

The template checklist below says to squash before review. **That does not apply to this PR.** Squashing would destroy the shared commit identity the maintenance-branch model depends on: `main` would no longer contain `maint/v3.9.x`'s 299 commits by sha, so every future forward merge would re-conflict on the same content. Use "Create a merge commit".

## Conflicts and how each was resolved

**`gradle.properties` — `JVM_TOOLCHAIN`**
Kept main's `21`. main raised it because the `zip321` dependency ships Java 21 bytecode (class file v65) that JVM unit tests must load; maint's `17` predates that and would break those tests.

**`ui-design-lib/src/main/res/ui/common/values/strings.xml` (3 hunks)**
Pure lint-attribute conflict — no content difference. Both sides mark the same coinholder-polling strings as intentionally untranslated: main via `translatable="false"` (commit "Translations set as non translatable"), maint via `tools:ignore="MissingTranslation"` (commit "Mark disabled coinholder-voting strings as intentionally untranslated"). Took main's, the later refinement of the same intent. Verified all three hunks carried identical string-name sets in identical order, and that the merged file still defines every one of the 1253 names on maint and the 1029 on main.

**`ui-lib/src/main/res/ui/balances/{values,values-es}/strings.xml` and `ui-lib/src/main/res/ui/error/{values,values-es}/strings.xml`** (modify/delete)
Accepted maint's deletion. maint's "MOB-1430: Rename Phase 3" consolidated these into ui-design-lib's common `strings.xml`; every string main still had in them resolves elsewhere in the merged tree, so keeping the files would only have duplicated resources.

**`ui-lib/src/main/java/co/electriccoin/zcash/di/UseCaseModule.kt`**
[...]

PR #2424, opened 2026-08-14
Repair maint/v3.9.x merge ancestry (PR #2423 landed as a squash)

nesence-m:
PR #2423 was auto-squash-merged by the auto-merge workflow (a `synchronize`-triggered race against the manual disable-auto call), breaking commit identity between `maint/v3.9.x` and `main` a second time — the first time was PR #2408.

This PR is an ancestry-only merge commit with **ZERO content change**: `0 files changed` is expected and correct (verified `git diff` against `main` is empty both before and after this commit). It exists only to record `merge/maint-v3.9.x-into-main`'s tip as an ancestor of `main` by sha, so `maint/v3.9.x` is fully contained again and the next cascade doesn't re-conflict on already-landed content.

**Must land via a merge commit — never squash.** Squashing this PR would defeat its entire purpose.

---


LukasKorba (approved):
utACK

PR #2427, opened 2026-08-17
MOB-1382: Fix reject button no-op on Keystone shield signing screen

noop-sk:
## Summary
- `CancelProposalFlowUseCase` had no branch for `ShieldTransactionProposal`, so rejecting a Keystone shield-signing request fell into the generic `else` case (`navigationRouter.backTo(Send::class)`), which silently no-ops because `Send` is never on the back stack for the auto-shield flow — the user gets stuck on the signing screen after tapping "Reject signature".
- Added a `ShieldTransactionProposal` branch that calls `navigationRouter.back()`, popping the signing screen back to whichever origin screen (shield-prompt dialog, Balances, or Home) is actually beneath it — same pattern already established for the swap/migration branches in this use case (see `a0b729da4`, `61aa84a5c`).

## Test plan
- [x] Added `shieldProposalNavigatesBackInsteadOfSend` to `CancelProposalFlowUseCaseTest`, confirmed it fails against pre-fix code and passes after
- [x] `:ui-lib:testZcashmainnetInternalDebugUnitTest` — 342 tests, 0 failures

Fixes MOB-1382: https://linear.app/zodl/issue/MOB-1382/keystone-device-cannot-reject-zodl-shield-transaction


nesence-m (approved):
utACK

PR #2430, opened 2026-08-17
docs(signing): publish APK certificate fingerprints for Obtainium/Verified Apps

### LINKED ISSUE #2394: Sign the APK

People that use [Obtainium](<https://github.com/ImranR98/Obtainium>) and ["Verified Apps"](<https://github.com/privacyguides/verified-apps-android>) would benefit if this app had a signature that we could verify.

I think this is how to do it? [https://developer.android.com/studio/publish/app-signing](<https://developer.android.com/studio/publish/app-signing>)

## Alternatives you've considered

Not filing and issue and just coping with an unverified app and unverified updates.

## Additional context

None.

---

## Resolution

PR #2430 opened: `docs(signing): publish APK certificate fingerprints for Obtainium/Verified Apps`

**Certificate SHA-256:** `412a1d2412a0be5ffbad9e7dd1eeedb9442dad8b9e9dffc334c07a0c3645ddfe`

Verified from 3.9.2-2370 release APK via `apksigner verify --print-certs`.

**Note on certificate DN:** shows `Zerocoin Electric Coin Company` — the original key created under ECC/Zashi. Not rotated as part of ZODL rebrand (rotation would require all F-Droid/direct APK users to reinstall). SHA-256 fingerprint is the canonical identifier.

**Changes in PR:**

* `README.md`: APK signing section with fingerprints + apksigner verification command + Obtainium setup instructions
* `apk-signatures.json`: machine-readable fingerprint file for automated tools
* `release.yaml`: print cert fingerprints in CI on every release for public auditability

y4ssi:
Fixes #2394.

Publishes the ZODL APK signing certificate fingerprints so users of Obtainium and privacyguides Verified Apps can verify APK authenticity.

**Certificate SHA-256:** `412a1d2412a0be5ffbad9e7dd1eeedb9442dad8b9e9dffc334c07a0c3645ddfe`

Verified from the 3.9.2-2370 release APK via `apksigner verify --print-certs`.

Note: the certificate DN shows `Zerocoin Electric Coin Company` — this is the original key created under the ECC/Zashi name. The key has not been rotated as part of the ZODL rebrand (rotating would require users to reinstall). The SHA-256 fingerprint is the canonical identifier.

**Changes:**
- `README.md`: APK signing section with fingerprints, `apksigner` verification command, and Obtainium setup instructions
- `apk-signatures.json`: machine-readable fingerprint file for automated tools
- `release.yaml`: print cert fingerprints in CI on every release for public auditability

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -181,15 +181,30 @@ jobs:
           gpg -u sysadmin@zodl.com --armor --digest-algo SHA256 --detach-sign app-zcashmainnet-foss-release.apk
           gpg -u sysadmin@zodl.com --armor --digest-algo SHA256 --detach-sign app-zcashmainnet-foss-release-full-externallibs.apk
           gpg -u sysadmin@zodl.com --armor --digest-algo SHA256 --detach-sign app-zcashmainnet-google-play-store-release-universal.apk
-          gpg --armor --export sysadmin@zodl.com > zodl-gpg-public-key.asc
+          GPG_FINGERPRINT=$(jq -r '.gpgKeyFingerprint' "$GITHUB_WORKSPACE/apk-signatures.json")
+          gpg --armor --export "$GPG_FINGERPRINT" > zodl-gpg-public-key.asc
+          test -s zodl-gpg-public-key.asc || { echo "::error::GPG key $GPG_FINGERPRINT (apk-signatures.json) is not in the CI keyring"; exit 1; }
 
-      - name: Print APK signing certificate fingerprints
+      # Gate, not just a printout: every APK must carry the certificate published
+      # in apk-signatures.json (and README.md), so a rotated or mis-set upload
+      # keystore cannot ship a release whose fingerprint contradicts the docs.
+      - name: Verify APK signing certificate fingerprints
+        shell: bash
         run: |
           cd artifacts/
+          : "${ANDROID_HOME:?ANDROID_HOME must point at the Android SDK}"
           APKSIGNER="$ANDROID_HOME/build-tools/$(ls "$ANDROID_HOME/build-tools" | sort -V | tail -n 1)/apksigner"
+          EXPECTED=$(jq -r '.signingCertificates[0].sha256' "$GITHUB_WORKSPACE/apk-signatures.json")
           for apk in *.apk; do
             echo "=== $apk signing certificate fingerprints ==="
-            "$APKSIGNER" verify --print-certs "$apk"
+            CERTS=$("$APKSIGNER" verify --print-certs "$apk")
+            echo "$CERTS"
+            { echo "### $apk"; echo '```'; echo "$CERTS"; echo '```'; } >> "$GITHUB_STEP_SUMMARY"
+            ACTUAL=$(echo "$CERTS" | sed -n 's/^Signer #1 certificate SHA-256 digest: //p')
+            if [ "$ACTUAL" != "$EXPECTED" ]; then
+              echo "::error::$apk is signed with certificate $ACTUAL, expected $EXPECTED (apk-signatures.json)"
+              exit 1
+            fi
           done
 
       - name: Upload to Release
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -181,12 +181,16 @@ jobs:
           gpg -u sysadmin@zodl.com --armor --digest-algo SHA256 --detach-sign app-zcashmainnet-foss-release.apk
           gpg -u sysadmin@zodl.com --armor --digest-algo SHA256 --detach-sign app-zcashmainnet-foss-release-full-externallibs.apk
           gpg -u sysadmin@zodl.com --armor --digest-algo SHA256 --detach-sign app-zcashmainnet-google-play-store-release-universal.apk
+          gpg --armor --export sysadmin@zodl.com > zodl-gpg-public-key.asc
 
       - name: Print APK signing certificate fingerprints
         run: |
           cd artifacts/
-          echo "=== APK signing certificate fingerprints ==="
-          apksigner verify --print-certs app-zcashmainnet-foss-release.apk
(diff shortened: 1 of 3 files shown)


LukasKorba (commented):
Second pass — all four of Dominik's findings are addressed. One doc fix before merge, plus two checks that are mine to do out of band.

**Verified independently**

- **GPG key** — fetched `033834DD49DECF9DBB9934BC6C93CA8E58E26AB1` from keys.openpgp.org and parsed the packets: primary fingerprint and signing subkey `1FE99324758F296718B457067F4BBBBA23F0617F` both match README:75,82-83 and `apk-signatures.json:19`. The key is served with **zero uid packets** and `by-email/sysadmin@zodl.com` 404s — so README:85-88 is accurate about the identity being unverifiable, not hedging.
- **APK SHA-256** — `412a1d24…3645ddfe` matches `AllowedAPKSigningKeys` in fdroiddata `metadata/co.electriccoin.zcash.foss.yml:927`, which F-Droid enforces against their reproducible build. Third-party corroboration outside our control, and it also independently backs the README:35-36 claim that f-droid.org ships the developer-signed APK.
- **Obtainium snippet** — checked against Obtainium's `source_provider.dart`: `apkFilterRegEx` is an unanchored `hasMatch` over candidates already filtered to apk/xapk/apkm/apks, so the `.asc` assets can't be selected, and the filter does not also match `-full-externallibs.apk`.
- **Table vs JSON** — byte-identical, every digest correct length and valid hex, DNs match. The choice of `apksigner` over `keytool` in the doc is the right one; `keytool` prints colon-separated uppercase and would have produced false mismatches against this table.

**Please fix before merge**

README:62-64,71 — the primary key-import path 404s for every release that exists today. `zodl-gpg-public-key.asc` is only produced by the new `release.yaml` step, so 3.10.2-2569 and 3.10.1-2512 don't carry it, and the "grab it from a newer release" fallback doesn't exist either. With the keyserver route closed by the missing uid, anyone following this doc between merge and the next release has no working way to import the key — while the text asserts present tense ("is attached to every release"). Either reword to future tense or land this alongside a release cut.

**Nits**

- README:38 — "may carry Google's app-signing certificate" is weaker than our own `docs/Sideloading.md:112-115`, which states flatly that Play re-signs on delivery. As written, a Play user who runs `apksigner` and sees a mismatch can't tell expected from attack, which is the exact confusion this doc exists to prevent. I'll confirm the cert in Play Console; then let's make it "will".
- `release.yaml` cert gate — `sed -n 's/^Signer #1 …'` combined with `jq '.signingCertificates[0]'` means an APK carrying an *additional* signer still passes, and the JSON array can never hold a second (rotated) key without being silently ignored. It fails closed on missing `ANDROID_HOME`, a missing JSON key, and an `apksigner` error, which is the important part — just noting the ceiling.
- README:52 — consider dropping MD5 from the table. `apksigner` prints it, so it isn't wrong, but a doc teaching verification shouldn't offer a non-collision-resistant digest as an option.
- No stated validity window or rotation story. The CI gate makes silent drift impossible, so what's left unanswered is only "how would a user know a rotation was legitimate".

**Mine, out of band** — SHA-256 is corroborated but SHA-1 and MD5 are not (F-Droid publishes only SHA-256, and they can't be derived from it), so I'll run `apksigner verify --print-certs` against the 3.10.2-2569 asset, and confirm the GPG fingerprint against `/release/gpg-signing-key` in Secrets Manager. keys.openpgp.org accepts anonymous uploads, so matching it proves consistency, not ownership.

LukasKorba (commented):
`182abad3f` closes it — and I'd rather have this wording than the version I asked for: saying outright that the current latest doesn't carry the asset and has no release-asset fallback yet is more useful to someone trying to verify today than a bare future tense would have been. No further code comments from me.

Holding the approve on my own out-of-band checks rather than on anything in the diff:

- `apksigner verify --print-certs` against the 3.10.2-2569 asset. SHA-256 is corroborated by F-Droid's `AllowedAPKSigningKeys`, but SHA-1 and MD5 aren't and can't be derived from it, so those two rows in the table are still unverified by anyone but you.
- The GPG key's *ownership* against `/release/gpg-signing-key` in Secrets Manager. My packet parse only proves the README matches what keys.openpgp.org serves, and that keyserver accepts anonymous uploads.
- Play Console for the "may carry" → "will carry" wording, as you said.

Will come back with all three.

LukasKorba (commented):
Revising what I asked for here — the two commands I listed earlier were the wrong ask, and one of them was a bad idea outright.

**Withdrawing the Secrets Manager check.** I suggested piping `/release/gpg-signing-key` through `gpg --import-options show-only`. That keeps the key out of the keyring but not out of the shell pipeline, history, or process list, and our own rule is that key material never lands in logs. Don't run it, and I shouldn't have proposed it.

The same fact is available with no secret handling at all: the release step this PR adds publishes `zodl-gpg-public-key.asc` as a release asset. Once the next release cuts, `gpg --show-keys` on that public file gives the fingerprint, and it comes from the same CI that does the signing. That's strictly better evidence than my version anyway — it verifies the key users will actually fetch, not the one sitting in the secret store.

**And the `apksigner` run is now mostly redundant**, because of the gate you added. CI runs `apksigner` on every release and fails the build if the certificate doesn't match `apk-signatures.json`, so SHA-256 is machine-enforced from here on — and independently corroborated by F-Droid's `AllowedAPKSigningKeys`.

That leaves SHA-1 and MD5 as the only values in the table nobody checks: the gate reads `.sha256`, F-Droid publishes only SHA-256, and neither can be derived from it. **So I'd rather you deleted both rows than verified them.** MD5 has no business in a verification doc regardless — it isn't collision-resistant, and offering it as an option next to SHA-256 invites someone to use it. Dropping both leaves the doc publishing only values that are enforced, which is a stronger position than publishing three and enforcing one.

If you'd rather keep SHA-1 for tooling that only speaks it, that's defensible — just say so in the doc and I'll confirm it against a release APK before we merge.

Everything else stands: the reworded key-import paragraph is good, and the remaining open item is mine — Play Console → App integrity, to settle whether README:38 says "may" or "will".

PR #2435, opened 2026-08-18
[MOB-1678] Use resilient voting config pin

noop-sk:
## Summary
- Points the bundled production `StaticVotingConfig.BUNDLED_PINNED_SOURCE` at the `voting.valargroup.dev` content-addressed gateway instead of directly at raw GitHub. The gateway reads GitHub normally and serves the automatically published Cloudflare copy during a GitHub outage, so a GitHub outage no longer blocks default-configured users from loading their voting trust anchor.
- The embedded SHA-256 checksum is still enforced exactly as before (verified live: the new gateway URL returns byte-identical content matching the pinned checksum) — the trust model is unchanged, only the transport gained a fallback path.
- Android has no separate bundled stage anchor, so this intentionally changes only the production pin (same as the iOS counterpart).

Android analog of the iOS hardening PR: zodl-ios#1996 (raised by Adam/Valargroup in #ext-zodl-valargroup, MOB-1678).

## Test plan
- [x] `ui-lib` unit tests (`StaticVotingConfigTest`, `VoteServerFailoverTest`, `VoteChainConfigVMTest`) pass
- [x] Live-fetched the new pinned URL and confirmed its SHA-256 matches the embedded checksum exactly
- [ ] On-device smoke test: default-configured wallet still loads the voting trust anchor and rounds list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+### Changed:
+
+- Coinholder Polling now loads its trusted configuration through the resilient voting config gateway, so a GitHub outage no longer blocks
+  configuration loading while the mirrored copy is available.
+
 ### Fixed:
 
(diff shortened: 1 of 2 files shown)


nesence-m (approved):
utACK

PR #2438, opened 2026-08-18
fix(fdroid): wait for GitLab pipeline + purge CF cache on publish

y4ssi:
## Problem

F-Droid users were getting 416 (invalid range) and hash-mismatch errors because:
1. The `publish_to_fdroid` job only fired the GitLab trigger and moved on — silent failures went unnoticed
2. Releases 3.9.0, 3.9.1, 3.9.2 all had their GitLab pipelines fail silently, leaving the index stuck at 3.8.1
3. CF cache served the stale APK even after manual fix

## Fix

- **Poll GitLab pipeline** until success/failure (max 15 min, 10s intervals) — fails the GH job if GitLab fails, making the problem immediately visible
- **Purge Cloudflare cache** after successful publish for APK + index-v2.json + entry.jar — no more stale caches

## Test plan
- [ ] Run workflow dispatch for next release, verify GitLab polling appears in logs
- [ ] Verify CF purge step runs after GitLab success

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -203,40 +203,92 @@ jobs:
   # release tag, so the right APK version gets published. All the AWS Secrets
   # Manager / keystore / fdroidserver / git-push work now lives in GitLab CI.
   #
-  # Best-effort: a failed trigger must not fail the release.
+  # Triggers the GitLab Pages F-Droid pipeline, waits for it to finish,
+  # and purges the Cloudflare cache on success so clients get fresh APK+index.
   publish_to_fdroid:
     needs: [release]
     runs-on: ubuntu-latest
     permissions:
       contents: read
     steps:
-      - name: Trigger GitLab Pages F-Droid publish
+      - name: Configure AWS credentials (OIDC)
+        uses: aws-actions/configure-aws-credentials@d979d5b3a71173a29b74b5b88418bfda9437d885 # v6.1.1
+        with:
+          role-to-assume: ${{ secrets.AWS_RELEASE_ROLE_ARN }}
+          aws-region: us-east-1
+
+      - name: Trigger GitLab Pages F-Droid publish and wait
         env:
           GITLAB_FDROID_TRIGGER_TOKEN: ${{ secrets.GITLAB_FDROID_TRIGGER_TOKEN }}
-          # Project id + target ref live in repo variables so they can be updated
-          # in one place if the GitLab project is recreated or the pipeline moves
-          # to an environment branch (review: noop-sk). Default to the known
-          # values if the vars are unset, so the step still works pre-config.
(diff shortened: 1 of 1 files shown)


zcash-harry (changes requested):
## Review: two likely-blocking issues, plus a correctness/security gap

### 🔴 Likely blocking: AWS OIDC step will fail

The workflow has a top-level `permissions: id-token: write` (line 30), but `publish_to_fdroid` declares its own job-level `permissions: contents: read` — job-level `permissions:` fully replaces the workflow default, it doesn't merge with it. Since this job's block doesn't list `id-token: write`, the new `Configure AWS credentials (OIDC)` step should fail outright with an OIDC permission error. Needs:

```yaml
permissions:
  contents: read
  id-token: write
```

### 🔴 Likely blocking: GitLab trigger token reused for API polling

The new polling loop sends `GITLAB_FDROID_TRIGGER_TOKEN` as a `PRIVATE-TOKEN` header to `GET /pipelines/:id`. GitLab **pipeline trigger tokens** are a distinct credential type from personal/project access tokens — scoped only for `POST .../trigger/pipeline`, not valid as `PRIVATE-TOKEN` for general API reads. If that's genuinely what this secret is (the name strongly suggests so, and the pre-PR code only ever used it for the trigger POST), every poll request 401s, `.status` never resolves, and every release times out after 15 minutes and fails — even when the GitLab pipeline itself succeeds. Worth confirming the actual token type before merging; if it's trigger-only, polling needs a separate read-scoped token (or a callback from GitLab CI instead of polling from this side).

### 🟠 Correctness — undermines this PR's own goal

The Cloudflare purge step never checks for failure. `ZONE_ID` isn't validated (a bad token or renamed zone yields `zones/null/purge_cache`), and the final purge result is only `echo`'d, never asserted — the step exits 0 regardless. Given the PR's premise is "silent failures went unnoticed," this reintroduces that same failure mode for the cache-purge half: a broken CF token or zone lookup silently no-ops while the job reports success.

### 🟡 Security — unmasked secret in logs

`CF_ZONE` (actually the CF API token) comes from `aws secretsmanager get-secret-value`, not the GitHub `secrets.*` context, so GitHub's automatic log-redaction doesn't cover it. Suggest `echo "::add-mask::${CF_ZONE}"` right after fetching it.

### 🟡 Naming/clarity

`CF_ZONE` actually holds the API token, not a zone — confusing next to the real `ZONE_ID` a few lines later.

### ⚪ Minor

`PIPELINE_ID` is written to `$GITHUB_ENV` but nothing reads it — the trigger and poll loop are both in the same `run:` block.

---

**What's solid:** the AWS action is pinned by commit SHA, the polling loop's status handling is sensibly exhaustive with a graceful fallback on transient API hiccups, the 15-minute budget math checks out, and the trigger-call failure handling is correctly stricter than before.

Holding on the two 🔴 items — both look like they'd fail on the very next real release.

---
*Sent using Claude*

zcash-harry (approved):
utACK

PR #2440, opened 2026-08-18
fix(fdroid): read CF zone ID from SM instead of API lookup

y4ssi:
Follow-up to #2438.

The CF zone ID for zodl.com is now stored in SM at `/infra/cloudflare/foss-fdroid-zone-id` and read in the workflow — same pattern as the API token. This removes the zone ID from the workflow source and the API lookup call on every release.

**No new secrets in GitHub Actions** — everything comes from SM via OIDC.


zcash-harry (approved):
utACK

PR #2444, opened 2026-08-19
MOB-1740: App Theme in Settings with Pure Black (OLED) dark look

nesence-m:
## What

Final MOB-1740 UX per the Q3 designs ([Figma node 7011-6045](https://www.figma.com/design/1aeq8gleYh9Yr1l33TwELR/PR-App-Designs-Q3--26?node-id=7011-6045&m=dev)):

- **App Theme** is now a top-level Settings entry (palette icon, between Beta: Coinholder Polling and Advanced Settings) instead of living in Advanced Settings.
- The **Theme screen** matches the final design: logo + palette header, back-arrow top bar, System / Light / Dark option cards, footer Save changes. The OLED checkbox is gone.
- Tapping **System** or **Dark** opens a new **"When your device is dark"** bottom sheet picking **Classic Dark** vs **Pure Black**; its Save changes persists the appearance + dark look and pops back to the Settings list. Tapping **Light** selects it locally and the screen's own Save changes persists it.
- Pure Black keeps the OLED palette work from the earlier commits (OledZashiColors, forced -night resource resolution, themed in-app browser chrome).

## How

- `ThemeSettingsVM/View` reworked (no checkbox; System/Dark forward to the sheet route); new `screen/theme/darklook/` package with `ThemeDarkLookArgs/State/VM/View` registered as `dialogComposable`, mirroring the SwapSlippage sheet pattern.
- `SetAppearanceModeUseCase` now ends with `navigationRouter.backTo(MoreArgs::class)`.
- Strings updated to the final Zodl copy (EN + ES); new `ic_settings_theme` palette drawable.

## Testing

- `ThemeSettingsVMTest` (7) + new `ThemeDarkLookVMTest` (5) — all green.
- ktlint clean; detekt has only the two pre-existing findings on the branch (verified unrelated via stash).
- Manually verified on an emulator: entry position/icon/copy, Dark → sheet → Pure Black → Save lands on Settings with pure black applied, persistence on re-entry, Light path.


LukasKorba (commented):
Blocked on one item: the fix round introduced a functional regression that makes the core flow of the feature unreachable. Finding #1 was implemented literally, but the bot's premise was wrong for this screen, and applying it broke Save.

## Blocking

**You can no longer switch to System or Dark unless you also change the dark look.**

`ThemeDarkLookVM.kt:61` — `isEnabled = oledEnabled != originalOledEnabled`.

The VM only knows the OLED boolean (its constructor takes `args`, `navigationRouter`, `isOledEnabledStorageProvider`, `setAppearanceMode` — no `AppearanceModeStorageProvider`). But its Save is the **only** path that persists the appearance mode for SYSTEM and DARK:

- `ThemeSettingsVM.kt:76-82` — `onModeClick` forwards everything that isn't LIGHT to the sheet and deliberately doesn't touch `selectedMode`.
- `ThemeDarkLookVM.kt:71-74` — `onSaveClick` calls `setAppearanceMode(args.mode, …)`, which is what writes `appearance_mode`.
- `SetAppearanceModeUseCase` has exactly two call sites in the tree, and that's the one that matters here.

Repro from stored `LIGHT` + `is_oled_enabled=false` (also the never-chosen default):

1. Tap **Dark** → sheet opens with Classic Dark preselected, so `selectedOledEnabled == originalOledEnabled == false`.
2. `isEnabled = false != false` → **Save disabled**.
3. No other control on the sheet. Dismiss is `onBack()` → `navigationRouter.back()`, nothing persisted.

So Dark-with-Classic-Dark is unreachable, and the only way to reach Dark at all is via Pure Black. Same for SYSTEM ↔ DARK in either direction while the look stays put, and for LIGHT → SYSTEM. This worked at `cdc2782` (the parent), where `ButtonState` had no `isEnabled` and defaulted to `true`.

The invariant this sheet actually needs is `(oledEnabled != originalOledEnabled) || (args.mode != storedAppearanceMode)`, which means injecting `AppearanceModeStorageProvider` — or keep it always-enabled and document why. The bot's "mirror `ThemeSettingsVM`" reasoning doesn't transfer: the Theme screen's Save commits one value, the sheet's commits two.

**And the tests now pin the broken behaviour.** `ThemeDarkLookVMTest.kt:81-91` asserts it as correct. `:94-115` seeds `stored = LIGHT, args = DARK` but clicks Pure Black *before* Save, so it never touches the mode-changed-look-unchanged path — and dropping `assertTrue(state.saveButton.isEnabled)` from `assertSingleSelection` (`:130-141`) removed the one assertion that would have caught this.

## On the other findings

Card order (`ThemeSettingsVM.kt:52`) and the test fakes (both files now `MutableStateFlow`-backed) are addressed. Two notes: nothing pins the card order — `ThemeSettingsVMTest.optionFor:167-170` uses `.first { it.mode == mode }`, order-insensitive — and `ThemeSettingsView.kt:222` still uses `AppearanceMode.entries` in the preview. The `observe()` fake change is cosmetic since nothing exercises it.

You were right to push back on the `ic_settings_theme.xml` finding — both call sites tint (`SettingsListItemLeadingIcon.kt:26-31`, `ThemeSettingsView.kt:171-176`) and `ColorFilter.tint` overrides the vector's stroke. Same for the `ContextThemeWrapper` one: `onApplyThemeResource` only runs on `getTheme()` and the object is rebuilt when the key changes, so that `IllegalStateException` path isn't reachable. Both bot false alarms.

## Should-fix

**Flash of the wrong theme on cold start.** `MainActivity.kt:171-177` reads the theme from `OldHomeViewModel`, whose flows seed with `SYSTEM`/`false` (`OldHomeViewModel.kt:39-49,52-60`); the backing read is `Dispatchers.IO.limitedParallelism(1)` so it can't land before first composition. The splash gate is `SecretState.LOADING` (`MainActivity.kt:159`), unrelated, and `SPLASH_SCREEN_DELAY = 0.seconds`. Dark device + Light chosen → dark frames then light, and vice versa. It's a full repaint, not a subtle tint, because the flip swaps `LocalContext`/`LocalConfiguration` (`ZcashTheme.kt:122-125`) and the whole tree re-resolves. Gating `setKeepOnScreenCondition` on the first appearance-mode emission closes it. (The read itself is correctly off the main thread — that part's clean.)

## Nits

- `ThemeSettingsVM.kt:86-89` — `require` inside `viewModelScope.launch` is a production crash, not an assertion. If the precondition ever breaks (someone adds a mode that doesn't forward to the sheet — the exact case the bot worried about) the `IllegalArgumentException` escapes to the default handler. Early return + log documents the invariant without shipping a crash.
- `ThemeSettingsScreen.kt:14` — `BackHandler { state?.onBack() }` is unconditionally enabled, so a back press during the null window is swallowed. `enabled = state != null`.
- `ThemeOption.kt:82-86` — plain `clickable`, no `Role.RadioButton`/`selectableGroup`, so TalkBack announces three generic buttons with no selected state. Not a regression (matches `exchangerate.settings.Option`) but `MigrationModeOption` in this same PR does it properly.
- Same seed-then-correct shape as the cold-start flash on both theme screens (`ThemeSettingsVM.kt:29-47`, `ThemeDarkLookVM.kt:27-45`) — one frame can render System/Classic-Dark checked regardless of what's stored.
- `ui/theme_settings/drawable/ic_theme_settings.xml` and its `-night` twin are added and referenced by nothing. Leftovers from the earlier moon-icon commit; the res folder is still needed for the strings.
- `ThemeSettingsView.kt:171-176` — 40dp icon whose art occupies the inner 20 units, in a 40dp `Box` with another `padding(10.dp)`; net glyph ≈10dp vs ≈20dp in the Settings row. Worth an eyeball against Figma.

## Checked and clear

I went looking for M3 tonal-elevation grey leaking into pure black and it isn't there: fully token-driven, `BlankBgScaffold` uses `Surfaces.bgPrimary`, the sheet container `Surfaces.bgSecondary`, all three `tonalElevation` usages are `0.dp`, and there's no `surfaceContainer*`/`surfaceVariant`/`surfaceTint` anywhere — so `OledColorPalette` copying only `surface`/`background` is fine. System bars are right (`ZcashTheme.kt:147-169`, `DefaultOledScrim` on the nav bar). No hardcoded colors in the new Compose code, light mode untouched. `splash_screen_background` stays `#231F20` under OLED, but that's pre-existing and out of scope.

## Scope

All defensible, none silent, but flagging what touches existing screens: three `drawable-night` assets (`ic_tor_settings`, two transaction-filter icons) had `#231F20` fills removed so they don't show as charcoal squares on pure black — the tor one redraws the disc at r=32 instead of stroking a ring over r=34, claimed pixel-identical on classic dark. Worth a designer's eye on **classic** dark, not just OLED. `OledZashiColors.kt` raises `strokeSecondary` Shark.900 → Shark.800, OLED-only. And `MigrationSetupScreen.kt:309-375` swaps M3 `RadioButton` for the new shared `ZashiRadioIndicator` — geometry replicates M3's and `Role.RadioButton` survives, but `ZashiRadioButton.kt:174-178` documents `checkedContent` as needing to be opaque while `MigrationRadioCircle` is a stroked ring; it only works because the geometry matches exactly. Tighten that KDoc.

ui-design-lib/src/main/java/co/electriccoin/zcash/ui/design/theme/ZcashTheme.kt
```diff
@@ -38,49 +48,115 @@ import co.electriccoin.zcash.ui.design.theme.typography.ZashiTypographyInternal
...
+ * @param appearanceMode The user's chosen light/dark appearance. Defaults to the value provided by an enclosing
+ * [ZcashTheme] so that nested, always-dark screens inherit the user's choice.
+ * @param isOledEnabled Whether pure black should be used whenever the theme resolves to dark. Independent of
+ * [appearanceMode] - it applies under [AppearanceMode.DARK] just as much as under [AppearanceMode.SYSTEM]
+ * resolving to dark. Defaults to the value provided by an enclosing [ZcashTheme].
  */
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun ZcashTheme(
     forceDarkMode: Boolean = false,
```
nesence-m: remove this flag completely - we can force appearanceMode param as AppearanceMode.Dark instead

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/viewmodel/OldHomeViewModel.kt
```diff
@@ -27,6 +34,65 @@ class OldHomeViewModel(
      */
     val isHideBalances: StateFlow<Boolean?> = booleanStateFlow(StandardPreferenceKeys.IS_HIDE_BALANCES)
 
+    /**
+     * Both stored theme values as one emission, so [appearanceMode], [isOledEnabled] and [isThemeResolved]
+     * all derive from a single upstream read and can never disagree about whether it has landed yet. Null
+     * until that read completes.
+     */
+    private val theme: StateFlow<ThemeAppearance?> =
```
nesence-m: lets not use OldHomeViewModel at all but instead lets make ThemeVM

ui-design-lib/src/main/java/co/electriccoin/zcash/ui/design/theme/ZcashTheme.kt
nesence-m: Done in 373ae74a1 - `forceDarkMode` is gone from `ZcashTheme`; the always-dark screens and dark previews now pass `appearanceMode = AppearanceMode.DARK`, and the former `forceDarkMode = false` call sites just drop the argument.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/viewmodel/OldHomeViewModel.kt
nesence-m: Done in 64e099a2c - the theme state moved to a new `ThemeVM` next to `ThemeSettingsVM`/`ThemeDarkLookVM` (registered in `viewModelModule`, read by `MainActivity`), and `OldHomeViewModel` is back to its pre-PR shape; covered by `ThemeVMTest`.

LukasKorba (commented):
The Save regression is properly fixed — that one's closed. Two things before this merges, and a note on scope.

## The regression — fixed, and the tests are better than I asked for

`ThemeDarkLookVM.kt:27` injects `AppearanceModeStorageProvider`, `:48` reads `storedMode`, and `:80` is the invariant. I walked all six cases including the genuine no-op:

| stored mode / oled | args.mode | selection | Save |
|---|---|---|---|
| LIGHT / false | DARK | Classic Dark | **enabled** — the original repro, now reachable |
| LIGHT / false | SYSTEM | Classic Dark | enabled |
| DARK / false | SYSTEM | unchanged | enabled |
| SYSTEM / false | DARK | unchanged | enabled |
| DARK / true | DARK | Classic Dark | enabled |
| DARK / false | DARK | Classic Dark | **disabled** — correct, genuine no-op |

And you avoided the seed-then-correct race I was worried about: `selectedOledEnabled` starts `null` (`:35`) and `state` maps null→null (`:39`), so `isSaveEnabled` can't run before the stored read, and both stored values are assigned at `:48-49` before the update at `:50` in the same launch. Save is never briefly wrong in either direction.

Making `isSaveEnabled` a required parameter of `assertSingleSelection` (`:247`) so every caller has to state it is stronger than the assertion I asked you to restore. `saveEnabledWhenModeDiffersEvenThoughLookIsUnchanged` (`:123`) asserts with no click first, which is exactly the missing case. Only gap left is stored=DARK→args=SYSTEM, which only has its mirror tested.

All seven nits addressed. Cold-start gate at `MainActivity.kt:167` is sound — `isThemeResolved` derives from the same combined upstream, `SharingStarted.Eagerly`, and both prefs emit unconditionally, so it does complete.

## Before merge

**1. `componentActivity()` unwraps exactly one level and this PR can stack two wrappers.** `KoinActivityViewModel.kt:44-50` is `context is ContextWrapper && context.baseContext is ComponentActivity`, else throw. `ScreenshotTest.kt:128` sets `configurationOverrideFlow` on every test → `Override.kt:32-44` wraps `LocalContext` → `ContextThemeWrapper(Activity)`; `MainActivity.kt:177-181` then renders `ZcashTheme` inside that, and `ZcashTheme.kt:109-120` wraps again → `ContextThemeWrapper(ContextThemeWrapper(Activity))`. `AndroidHome.kt:19`'s `koinActivityViewModel<ZashiTopAppBarVM>()` then sees a base context that isn't a `ComponentActivity` and throws.

The KDoc at `ZcashTheme.kt:94-100` argues the pattern is safe *because* it "matches the same safe pattern Override.kt already uses" — that holds for one wrap and is falsified by stacking with the very component it cites. Please fix the reasoning along with the code: either a `while` unwrap loop in `componentActivity()`, or don't re-wrap when the ambient context is already a wrapper.

This is the note you skipped in the earlier round as "informational, non-blocking". The finding as originally written was about `applyOverrideConfiguration`, and I agreed with you that that specific path wasn't reachable — but the double-wrap underneath it is a different and real defect, so the dismissal took the wrong half.

To be straight about what I did and didn't verify: I confirmed the unwrap logic and the two wrap sites by reading head, and `test_android_modules_emulator (app)` is red. I could **not** confirm from the check annotations that this is what's failing it — that same job is also red on #2497, which doesn't touch theming, and both show only generic runner annotations. So please confirm against the job log which of the two it is before you spend time on it.

**2. The head commit doesn't match your own description.** You said the state was gated on the first theme emission in `OldHomeViewModel`; head is `64e099a`, four commits past the block, and `373ae74`+`64e099a` (14:53, after your comment) relocate the theme state into a new `ThemeVM` across 24 files in the design lib. Same thing I flagged on #2494 an hour ago: I'm reviewing a diff whose changelog describes something else. Not asking you to revert it — just say what you pushed, and I'll read the right thing the first time.

Two soft edges in that new code:

- `MainActivity.kt:167` — `||` short-circuits, so `themeVM` isn't constructed while `secretState` is `LOADING`; the theme read is serialized *after* the wallet read instead of running alongside it. Reorder the operands and the gate costs nothing.
- `ThemeVM.kt:29-42` — no `.catch` or timeout. If `preferenceHolder()` throws, the `stateIn` coroutine dies and **the splash never lifts**. `ThemeVMTest.kt:74` pins that shape as intended. The failure mode I reported was "wrong theme, app opens"; the fix turns it into "app never opens", which is worse. Worth a timeout that falls back to SYSTEM.

## Nit

`ThemeOption.kt:124,141` — `ThemeRadioIndicator`'s `modifier` is applied inside both content lambdas rather than on the root. Inert today since no caller passes one, wrong if anyone does.

`Keystone` is a Maestro `home.syncComplete` timeout, unrelated to this.

LukasKorba (commented):
All four done. This is clean from my side.

**The unwrap loop terminates.** `KoinActivityViewModel.kt:46-53` — the `as?` yields null the moment the chain hits a non-`ContextWrapper`, which is the root of every real chain (`ContextImpl` under `Application`/`Activity`), and `ContextWrapper(null).baseContext` is null too, so both "chain ends without a `ComponentActivity`" cases throw with a better message than before. The only non-terminating input is a self-referential wrapper, which needs someone to deliberately build `ContextWrapper(null)` and attach it to itself; the framework never produces one. Unreachable rather than unguarded — fine.

**And you fixed both ends, which I didn't insist on.** `ZcashTheme.kt:119-153` skips the wrap when `resolvedConfiguration == currentConfiguration`, and I checked that doesn't break forced `-night`: `ConfigurationOverride.newConfiguration` (`Override.kt:55-64`) rewrites only the `UI_MODE_NIGHT_MASK` bits, so equality holds exactly when the ambient night bit already matches the resolved theme — precisely the case where no wrap is needed. Whenever the chosen appearance diverges, `:136-152` still wraps. I also looked for the case where the config compares equal but the context's own config disagrees; both providers in the repo supply `LocalConfiguration` and `LocalContext` together, so that path doesn't exist. KDoc at `:97-118` now names the chain walk as the safety net and the wrap-skip as an optimisation, which is the correction that mattered.

**The splash fallback is placed correctly**, and I checked the thing that would have made it useless: `.catch` at `ThemeVM.kt:57-59` does see the real failure, because `BaseNullableStorageProvider.observe()` (`StorageProvider.kt:55`) invokes `preferenceHolder()` *inside* the flow builder, so a throwing `create()` surfaces as a flow exception rather than a synchronous throw escaping the `channelFlow`. The fake models it the same way, so the test is faithful. Timeout and catch both emit `DEFAULT_THEME_APPEARANCE`, `isThemeResolved` goes true on both paths, gate releases. And you updated `themeStaysUnresolvedWhileEitherStoredValueIsMissing` to assert at virtual time zero instead of leaving it pinning a permanent hang — that's the part people usually forget.

One deliberate trade worth naming in the KDoc: after `.catch` emits, the combined flow completes, so if only one store throws the other's healthy value is discarded for the session with no retry — `ThemeVMTest.kt:128-140` pins exactly that (stored DARK + throwing OLED store → SYSTEM). Right call for a splash gate, just make sure it's a decision on the record rather than a surprise later.

`||` reorder at `MainActivity.kt:171` with the rationale at `:153-155` so it doesn't get tidied back, and `ThemeOption.kt:119` puts the modifier on the root. Both good.

**CI:** don't chase it. The emulator job's log has zero `ClassCastException`, zero `componentActivity`, zero `ComposeTimeoutException` — it's `Failed to find ColorBuffer` and `device offline` mid-run, which `pull-request.yml:300-309` documents as CPU/memory contention, and the same job is red right now on MOB-1340 and mob-1832. Since the double-wrap would have manifested exactly here, the absence of that signature is positive evidence the fix works, not just absence of evidence. `test_kotlin_modules` is green, which is where `ThemeVMTest` actually runs.

Small correction to your writeup: you called the e2e red "the Maestro `home.syncComplete` timeout", but that flow shows `COMPLETED` in the log — phase 2 died at `Maestro CLI not found` before any flow ran, and Keystone failed on the backup-phrase assertion. Same conclusion, different cause.

LukasKorba (dismissed):
Approving.

Worth recording what this took, since the interesting part wasn't the theme work: the Save-button finding from the bot was implemented literally and made Dark unreachable, and the `ContextThemeWrapper` note was dismissed as informational when the reachable half of it was a real double-wrap crash. Both are now fixed properly, and the second one you fixed at both ends rather than just the one I asked for.

Deliberately accepted: after `ThemeVM`'s `.catch` emits, the combined flow completes, so a single throwing store discards the other's healthy value for the session with no retry. Right call for a splash gate — just on the record.

Not blocking: the three `drawable-night` assets this PR edits change existing classic-dark screens, and the tor icon redraws its disc rather than stroking a ring. Claimed pixel-identical, worth a designer's eye on classic dark at some point, not before merge.

LukasKorba (approved):
Re-approving. I verified the rebase rather than taking the byte-identical claim on trust, and it holds — with a bonus.

Comparing the PR's own diff before and after (`main...00c80b45b` vs `maint/v3.10.x...a87d084c`), all 65 files match exactly, additions and deletions included. The rebase added nothing. It *dropped* two files:

- `feature-migration/.../FakeOrchardMigrationSdk.kt` (+3)
- `ui-lib/.../AutomaticServerRepository.kt` (+2-2)

Those are exactly the two bits of scope creep I flagged in the second round — the SDK-pin compile fix and the detekt `ReturnCount` refactor in a shared repository, neither related to theming. They're gone because `maint/v3.10.x` already carries them. So the rebase left the theme change untouched and shed the unrelated pieces. Cleaner than what I approved.

Keeping the pre-rebase history at `backup/mob-1740-pure-black-theme-main` was the right instinct — it's what let me do this comparison at all, since `00c80b45b` stays reachable.

One thing to be deliberate about rather than something to fix: the base moved from `main` to `maint/v3.10.x`, so this ships in 3.11.0. My review judged it as a feature landing on `main`; on a maint line the three `drawable-night` assets that change existing classic-dark screens are the part worth a second look, since patch-line users get a visual change they didn't ask for. Nothing blocking, and the KDoc note on the single-store-failure trade is in — thanks for that.

PR #2457, opened 2026-08-20
[MOB-1751] Parse and interpret cross-chain payment URIs

zcash-harry:
## Summary

- parse Bitcoin, Litecoin, EIP-681, and Solana payment URIs through one Android SDK parser API
- map typed SDK results into the existing asset-selection, Pay, and Swap flows
- reject Solana transaction requests, unsupported EVM methods, unsupported schemes, and unknown assets
- limit native-token aliases to supported app assets and actual cross-chain exceptions

## Dependencies

- https://github.com/zodl-inc/zodl-android-wallet-sdk/pull/7 (retargeted from zcash/zcash-android-wallet-sdk#2196, now closed -- SDK development moved to zodl-inc/zodl-android-wallet-sdk)
- https://github.com/zcash/librustzcash/pull/2989 (merged)
- CI is pinned to SDK commit fbacf6f65bdd309aaa6e34ba7a7158b2204fabde on zodl-inc/zodl-android-wallet-sdk

## Testing

- ktlint, detekt, app compilation, and the full ui-lib unit test suite passed against the local SDK checkout
- SDK quick CI (ktlint/detekt) and managed-device JNI instrumentation passed

AI disclosure: Implemented with OpenAI Codex assistance.


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
```diff
@@ -45,11 +53,30 @@ private sealed interface ScanAddressPipelineResult {
...
+    val asset: SwapAsset?,
+    val isPaymentRequest: Boolean
 )
+
+fun ScanResult.resolve(assets: Collection<SwapAsset>, currentAsset: SwapAsset?): ResolvedScanResult {
+    val requestedAsset = request?.resolveAsset(assets, currentAsset)
+    return ResolvedScanResult(
+        address = address,
+        amount = request?.resolvedAmount(requestedAsset) ?: amount.takeIf { request == null },
+        asset = requestedAsset ?: currentAsset.takeIf { request == null },
```
noop-sk: `resolve()` only falls back to `currentAsset`/the typed amount when `request == null` (this line, and the same pattern for `amount` on the line above: `request?.resolvedAmount(...) ?: amount.takeIf { request == null }`). Once `CrossPayRequestParser` recognizes *any* payment-request URI, `currentAsset` is discarded outright whenever `resolveAsset()` can't find a match — e.g. an unsupported chain, curated assets not yet loaded, or an ambiguous match (see the other comments on `CrossPayRequest.kt`).

Concretely: user has BNB already selected on Pay/Swap, scans a valid ERC-20 payment link for a token `resolveAsset()` can't resolve → the previously-good BNB selection silently becomes `null`, and (via `isPaymentRequest`, consumed in `PayVM.kt`/`SwapVM.kt`) the typed amount is cleared too. Tapping Request/Swap Quote then silently no-ops (`asset ?: return@launch`) with no error shown to the user. This should only overwrite `currentAsset`/the amount when the request actually resolved to a concrete asset, not merely when it was recognized as a payment request.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
```diff
@@ -0,0 +1,243 @@
...
+        // reference.chain is an explicit, already-trusted non-EVM chain (e.g. Solana's literal
+        // "sol") from the parser -- it doesn't need the same EVM-chain validation as the
+        // current-derived fallback used when chainId is absent (every ERC-20 request: EIP-681
+        // has no `chain` string, only a numeric chainId).
+        val chain = reference.chainId?.let(::evmChain) ?: reference.chain ?: evmChainIfKnown(current?.chainTicker)
+        if (reference.chainId != null && chain == null) return emptyList()
+        if (chain == null) return emptyList()
+        return assets.filter {
+            it.chainTicker.equals(chain, true) &&
+                it.contractAddress?.equals(reference.address, true) == true
```
noop-sk: This `equals(reference.address, true)` comparison is applied to both EVM contract addresses (hex, safely case-insensitive) and Solana SPL mint addresses (base58, case-sensitive — differing only by case is a *different* address) via the same code path, since `SolanaTransfer` requests with an `splToken` are also routed through `contractAssets()` (see `CrossPayRequestParser.toCrossPayRequest()`). If the curated asset list ever contains an SPL token whose stored mint address differs only in casing from a distinct token's mint, `resolveAsset` would treat them as equal and silently resolve to the wrong SPL token for a scanned `solana:...?spl-token=` request.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
```diff
@@ -0,0 +1,243 @@
...
+                null
+            }
+        }
+
+    private fun BigDecimal.asDisplayAmount() = CrossPayRequest.Amount(this, isAtomic = false)
+
+    private fun String?.toAtomicAmount(): CrossPayRequest.Amount? =
+        this?.let {
+            CrossPayRequest.Amount(
+                value = BigInteger(removePrefix("0x"), HEX_RADIX).toBigDecimal(),
```
noop-sk: `BigInteger(removePrefix("0x"), HEX_RADIX)` can throw `NumberFormatException` for a malformed/empty `valueHex` (or one using an uppercase `0X` prefix, since `removePrefix` is case-sensitive). `CrossPayRequestParser.parse()` only catches `InvalidPaymentUriException`, so this exception propagates further and is silently swallowed by the outer `runCatching` in `ScanGenericAddressVM.onScanned`. The result: `state` never updates to `VALID`/`INVALID`, `hasBeenScannedSuccessfully` stays `false`, and the scan screen is left stuck with zero feedback to the user.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
```diff
@@ -0,0 +1,243 @@
...
+        assets: Collection<SwapAsset>,
+        reference: AssetReference.Contract,
+        current: SwapAsset?
+    ): List<SwapAsset> {
+        // reference.chain is an explicit, already-trusted non-EVM chain (e.g. Solana's literal
+        // "sol") from the parser -- it doesn't need the same EVM-chain validation as the
+        // current-derived fallback used when chainId is absent (every ERC-20 request: EIP-681
+        // has no `chain` string, only a numeric chainId).
+        val chain = reference.chainId?.let(::evmChain) ?: reference.chain ?: evmChainIfKnown(current?.chainTicker)
+        if (reference.chainId != null && chain == null) return emptyList()
```
noop-sk: Unlike `evmNativeAssets()` above (which checks `reference.chainId != null && chain == null` *before* falling back to `current`), here the fallback to `evmChainIfKnown(current?.chainTicker)` is folded into the same expression as the explicit-chain resolution, so the null-check on the next line can't see whether the *original* chainId lookup failed.

Concretely: user has `usdcOnArbitrum` selected as `current` (chainTicker="arb"). Scans an EIP-681 ERC-20 URI for an unsupported chain (e.g. Fantom, id "250") whose contract address happens to equal the same USDC contract — `evmChain("250")` returns `null`, falls through to `evmChainIfKnown("arb")` = `"arb"`, the guard doesn't fire, and the resolver matches `usdcOnArbitrum` on the wrong chain instead of rejecting — exactly the bug class this PR explicitly fixed for `EvmNative` (see the comment above) but missed for `Contract`. No test in `CrossPayRequestTest.kt` covers this `Contract` case, only the `EvmNative` variant.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
```diff
@@ -0,0 +1,243 @@
...
+                "8453" to "base",
+                "42161" to "arb",
+                "43114" to "avax"
+            )
+        val NATIVE_TOKENS =
+            mapOf(
+                "arb" to setOf("eth"),
+                "base" to setOf("eth"),
+                "bsc" to setOf("bnb"),
+                "op" to setOf("eth", "op"),
```
noop-sk: `NATIVE_TOKENS["op"] = setOf("eth", "op")` makes a plain native transfer on Optimism ambiguous whenever the curated asset list contains both ETH (chainTicker="op", tokenTicker="eth") and the OP governance token (chainTicker="op", tokenTicker="op"). With `current == null`, a native EIP-681 transfer on Optimism (chainId "10") resolves via `nativeAssets(assets, "op")`, which matches *both* assets. Two candidates with nothing to disambiguate → `resolveAsset()` returns `null` for a request that's actually unambiguous (a native transfer is always the gas token, ETH), incorrectly showing "no asset resolved" to the user.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/SwapVM.kt
```diff
@@ -217,12 +218,16 @@ internal class SwapVM(
             if (result != null) {
                 navigationRouter.back()
                 internalState.update {
+                    val resolved = result.resolve(getCuratedSwapAssetsUseCase().data.orEmpty(), it.swapAsset)
                     it.copy(
                         selectedContact = null,
-                        addressText = result.address,
+                        addressText = resolved.address,
+                        swapAsset = resolved.asset,
```
noop-sk: The QR-scanner button is unconditional across both swap directions (`SwapVMMapper.kt` wires `qrScannerButton` outside the `when (state.swapDirection)` branches). In `SWAP_INTO_ZEC` mode the visible address field is the *refund* address (`SwapVMMapper.kt` routes `onAddressClick` to `SwapRefundAddressInfoArgs` only for this direction, and this VM uses `refundAddress = address` further down), not a payment recipient. But this line unconditionally overwrites `swapAsset` from the scan result regardless of direction — so scanning a QR of your own refund address (e.g. your BTC address, entirely unrelated to the swap) silently changes the swap's *source* token selection (e.g. USDC-on-Base → BTC) out from under the user with no confirmation.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
```diff
@@ -185,15 +186,24 @@ internal class PayVM(
             if (result != null) {
                 navigationRouter.back()
                 internalState.update {
+                    val resolved = result.resolve(getCuratedSwapAssetsUseCase().data.orEmpty(), it.asset)
```
noop-sk: `getCuratedSwapAssetsUseCase()` (which re-runs the curated-assets filter, an O(n·m) scan) is called synchronously inside the `internalState.update {}` CAS lambda. `update {}` is a compare-and-swap retry loop — under any concurrent writer to `internalState` (e.g. `onAddressChange` firing from user typing while the scan result comes back), this filtering work re-runs on every retry, and it duplicates a curated-assets computation the VM already produces reactively elsewhere via `getCuratedSwapAssetsUseCase.observe()`. Consider computing `resolved` once, outside the `update {}` block, before entering the atomic update. (Same pattern in `SwapVM.kt`'s `onQrCodeScannerClick`.)

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
```diff
@@ -0,0 +1,243 @@
...
+        if (reference.chainId != null && chain == null) return emptyList()
+        // Only fall back to the currently-selected asset's chain when it is actually an EVM
+        // chain -- a SOL/BTC/LTC asset selected as `current` must not silently match a native
+        // EVM request just because nativeTokens' permissive default matches any chain against
+        // itself.
+        val resolvedChain = chain ?: evmChainIfKnown(current?.chainTicker)
+        return resolvedChain?.let { nativeAssets(assets, it) }.orEmpty()
+    }
+
+    private fun contractAssets(
```
noop-sk, line 101: `evmNativeAssets()` (above) and `contractAssets()` independently implement the same three-way "resolve target EVM chain: explicit chainId, else explicit non-EVM chain, else current-asset fallback" shape, with `contractAssets` additionally needing a second `if (chain == null) return emptyList()` because it folds in `reference.chain` as a third source. This is copy-pasted branching with a slight variation rather than a shared helper — a future chain-family addition risks re-copying the same branch a third time and drifting, as the ordering bug flagged in the other comment on this method already shows one copy getting it right and the other wrong.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
```diff
@@ -0,0 +1,243 @@
...
+                is AssetReference.EvmNative -> evmNativeAssets(assets, reference, current)
+                is AssetReference.Contract -> contractAssets(assets, reference, current)
+            }
+
+        return current?.takeIf(candidates::contains) ?: candidates.singleOrNull()
+    }
+
+    private fun nativeAssets(assets: Collection<SwapAsset>, chain: String) =
+        assets.filter {
+            it.chainTicker.equals(chain, true) && it.tokenTicker.lowercase() in nativeTokens(chain)
```
noop-sk: `it.chainTicker.equals(chain, true)` reimplements the case-insensitive chain-ticker comparison that `SwapAsset.isSame()` already centralizes elsewhere in the codebase (used by `GetCuratedSwapAssetsUseCase`/`FilterSwapAssetsUseCase`). Not a functional bug here (the token-side check differs — set membership vs. a single ticker — so `isSame` doesn't drop in directly), but this is now a third independent copy of the same case-insensitive chain-comparison idiom (the other two are in `contractAssets()` below), and a shared `SwapAsset.matchesChain(chain: String)` helper would remove the duplication risk.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
```diff
@@ -45,11 +53,30 @@ private sealed interface ScanAddressPipelineResult {
     data class Scanned(
         val address: String,
         val amount: BigDecimal?,
+        val request: CrossPayRequest?,
         override val args: ScanGenericAddressArgs
     ) : ScanAddressPipelineResult
 }
 
 data class ScanResult(
```
noop-sk, line 61: `ScanResult.address`/`amount` are redundant once `request` is present: `resolve()` (below) ignores `amount` whenever `request != null` and only reads `ScanResult.address` because it happens to equal `request.address` by construction. These two fields exist purely to satisfy the non-payment-request code paths (`onZip321SingleAddressScanned`/`onZip321Scanned`, which never set `request`) — harmless today, but a maintenance trap if a future edit changes one extraction path (e.g. `ScanGenericAddressVM.onAddressScanned`) without the other.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
```diff
@@ -45,11 +53,30 @@ private sealed interface ScanAddressPipelineResult {
...
-    val amount: BigDecimal?
+    val amount: BigDecimal?,
+    val request: CrossPayRequest? = null
+)
+
+data class ResolvedScanResult(
+    val address: String,
+    val amount: BigDecimal?,
+    val asset: SwapAsset?,
+    val isPaymentRequest: Boolean
```
noop-sk: `isPaymentRequest: Boolean` here (and `Amount.isAtomic: Boolean` in `CrossPayRequest.kt`) is a boolean flag driving screen-state branching in `PayVM.kt`/`SwapVM.kt` (`resolved.isPaymentRequest -> NumberTextFieldInnerState()`), where this codebase's convention favors a sealed/state model over boolean flags for exactly this kind of state. Could be modeled as e.g. `sealed interface ResolvedScanResult { data class PaymentRequest(...) ; data class PlainAddress(...) }` instead.

noop-sk (commented):
Automated review pass (Claude Code) on the app-side of the cross-chain payment URI parser. Filed as individual inline comments below; summary: the most significant issue is that `ScanResult.resolve()` silently wipes the previously-selected asset and typed amount whenever a recognized payment-request URI fails to resolve to a concrete asset (unsupported chain, assets not yet loaded, ambiguous match), with no error shown to the user — plus a chain-fallback ordering bug in `contractAssets()` that mirrors a bug this PR explicitly fixed elsewhere, a case-insensitive address-match bug for Solana mint addresses, and an uncaught `NumberFormatException` path that can freeze the scan screen.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
noop-sk: Still open — this is the most significant issue from the original pass. `resolve()` is unchanged: `asset = requestedAsset ?: currentAsset.takeIf { request == null }`. A recognized-but-unresolvable payment request (unsupported chain, assets not loaded, ambiguous match) still silently discards the previously-selected asset and clears the typed amount, with no error surfaced to the user.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
noop-sk: Fixed — a new `addressesMatch(a, b, chain)` helper branches case-sensitive for `chain == "sol"` and case-insensitive otherwise, used in `contractAssets()`.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
noop-sk: Fixed — `toAtomicAmount()` now wraps the `BigInteger(...)` parse in `runCatching { }.getOrNull()`, so a malformed `valueHex` no longer hangs the scan screen. Residual: `removePrefix("0x")` is still case-sensitive for an uppercase `0X` prefix, but that case now silently resolves to "no amount" instead of propagating — the reported stuck-screen failure mode is gone.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
noop-sk: Fixed — both `evmNativeAssets()` and `contractAssets()` now go through a shared `resolveTargetChain(chainId, explicitChain, current)`, and an explicit `chainId` short-circuits with no fallback to `current`, matching the invariant this comment flagged as missing.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
noop-sk: Fixed — `NATIVE_TOKENS["op"]` is now `setOf("eth")` only, with a comment explaining OP is deliberately excluded to avoid the ambiguity.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/SwapVM.kt
noop-sk: Fixed — a new `isRefundAddressScan` gate preserves `swapAsset`/`amountTextState` in `SWAP_INTO_ZEC` mode. Residual (efficiency, not correctness): `result.resolve(...)` is still called unconditionally before the gate, so the resolution work itself runs and is discarded on refund-address scans.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
noop-sk: Partially fixed — `resolved` is now computed once outside the `update {}` CAS block, so the retry-loop duplication is gone. Still calls the synchronous `getCuratedSwapAssetsUseCase()` rather than reusing the reactive `.observe()`-derived list already held in state, so the redundant computation itself remains (once per call, not once per retry).

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
noop-sk: Fixed — consolidated into the shared `resolveTargetChain()` referenced in the comment above; both call sites now use one implementation.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
noop-sk: Fixed as suggested — a `SwapAsset.matchesChain(chain: String)` extension now exists and is reused at both call sites in this file. Minor caveat: it still doesn't reuse the pre-existing `SwapAsset.isSame()` centralized elsewhere in the codebase, though as noted originally that's not a functional bug.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
noop-sk: Unchanged — `ScanResult.address`/`amount` are still present alongside `request`. Low-priority maintenance note, not re-verified as blocking.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
noop-sk: Still open — `isPaymentRequest: Boolean` is unchanged, still a raw boolean flag rather than a sealed/state model.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/NearSwapDataSource.kt
```diff
@@ -63,6 +63,7 @@ class NearSwapDataSource(
             tokenIcon = tokenIconProvider.getIcon(dto.symbol),
             usdPrice = dto.price,
             assetId = dto.assetId,
+            contractAddress = dto.contractAddress,
```
noop-sk, line 72: Now that `contractAddress` is populated here, `getSupportedTokens()`'s dedup key a few lines above — `distinctBy { Triple(it.symbol, it.blockchain, it.decimals) }` — was not updated to include it, so two NEAR-returned tokens sharing symbol+chain+decimals but different contract addresses (e.g. native vs. bridged USDC on the same chain) collapse into one — the API order decides which survives. Since `contractAddress` is now load-bearing for `CrossPayRequest.contractAssets()` resolution, the silently-dropped variant becomes permanently unresolvable via a scanned payment URI even though the backend genuinely supports it. Suggest including `contractAddress` in the dedup key.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
```diff
@@ -184,16 +185,31 @@ internal class PayVM(
...
+                // computation the VM already produces reactively elsewhere via
+                // getCuratedSwapAssetsUseCase.observe(). The `else -> it.amount` fallback below
+                // still has to read state inside the lambda, since it depends on whatever the
+                // latest retry's state actually is.
+                val resolved = result.resolve(getCuratedSwapAssetsUseCase().data.orEmpty(), internalState.value.asset)
                 internalState.update {
+                    val amount =
+                        when {
+                            resolved.amount != null -> NumberTextFieldInnerState.fromAmount(resolved.amount)
+                            resolved.isPaymentRequest -> NumberTextFieldInnerState()
```
noop-sk: A recognized payment-request URI with no amount always clears the typed amount, even when the resolved asset is identical to what was already selected (e.g. rescanning the same `ethereum:0xRecipient` address with no `?value=`). Concretely: user has ETH-mainnet selected with "1.5" typed, rescans a bare address-only EIP-681 URI — `resolved.asset` resolves back to the same ETH-mainnet asset, but `resolved.amount == null` and `resolved.isPaymentRequest == true`, so the `when` hits `NumberTextFieldInnerState()` and silently discards the "1.5" even though nothing about the asset changed. Same pattern in SwapVM.kt:237.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/SwapVM.kt
```diff
@@ -216,13 +217,25 @@ internal class SwapVM(
...
+                        addressText = resolved.address,
+                        swapAsset = if (isRefundAddressScan) it.swapAsset else resolved.asset,
                         amountTextState =
-                            if (result.amount != null) {
-                                NumberTextFieldInnerState.fromAmount(result.amount)
+                            if (isRefundAddressScan) {
+                                it.amountTextState
+                            } else if (resolved.amount != null) {
+                                NumberTextFieldInnerState.fromAmount(resolved.amount)
+                            } else if (resolved.isPaymentRequest) {
```
noop-sk: Same amount-wipe issue as PayVM.kt:199 — a recognized-but-amountless payment request clears `amountTextState` even when `resolved.asset` is unchanged from the current selection.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/scan/ScanGenericAddressVM.kt
```diff
@@ -39,11 +40,13 @@ internal class ScanGenericAddressVM(
         }
 
     private suspend fun onAddressScanned(result: String) {
+        val request = CrossPayRequestParser.parse(result)
```
noop-sk: `CrossPayRequestParser.parse(result)` is a new suspend, Rust-backed call whose failures (other than `InvalidPaymentUriException`, which it catches internally) propagate up into the bare `runCatching { }` in `onScanned` (line 31), which has no `.onFailure` handling. If the SDK's JNI parser throws anything else (e.g. a native-library load failure), `state` never transitions to `VALID`/`INVALID` and `hasBeenScannedSuccessfully` stays false — the scan screen silently stops responding, the same failure class already fixed once for Keystone PCZT scanning (#2441).

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/scan/ScanGenericAddressVM.kt
```diff
@@ -39,11 +40,13 @@ internal class ScanGenericAddressVM(
...
         state.update { ScanValidationState.VALID }
         navigateToScanAddress.onScanned(
-            address = result,
-            amount = null,
-            args = args
+            address = request?.address ?: result,
```
noop-sk: `request?.address ?: result` collapses two distinct cases into the same fallback: "not a recognized URI scheme at all" and "a recognized scheme the parser explicitly rejects as out of scope" (per `CrossPayRequestParser`'s own doc comment — Solana Pay transaction-request links and unrecognized EIP-681 methods). A scanned Solana Pay tx-request QR (`solana:https://merchant.example/pay?...`) is a real-world format the parser recognizes and deliberately rejects, but here the entire URI string gets accepted as a literal recipient address and `state` is set `VALID` (line 44) with no error shown.

noop-sk (commented):
Follow-up pass: independent re-review of the commits added since the first review. These are new findings, not related to the threads above.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
```diff
@@ -0,0 +1,281 @@
...
+
+    fun resolveAsset(assets: Collection<SwapAsset>, current: SwapAsset?): SwapAsset? {
+        val candidates =
+            when (val reference = assetReference) {
+                is AssetReference.Native -> nativeAssets(assets, reference.chain)
+                is AssetReference.EvmNative -> evmNativeAssets(assets, reference, current)
+                is AssetReference.Contract -> contractAssets(assets, reference, current)
+            }
+
+        return current?.takeIf(candidates::contains) ?: candidates.singleOrNull()
```
noop-sk: `current?.takeIf(candidates::contains)` relies on `SwapAsset` data-class equality, which now also compares the newly-added `contractAddress` field. If `internalState.value.asset` ("current") was captured from an earlier `SwapAssetsData` snapshot where a token's `contractAddress` was null, and a later-refreshed curated-assets snapshot has the same token with `contractAddress` now populated, `candidates.contains(current)` fails by equality even though it is conceptually the same asset — breaking the "no chain id -> keep current selection" fallback that this PR's own regression tests (e.g. `missingChainIdFallsBackToCurrentSelection`) assume holds.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
zcash-harry: Fixed — `resolve()` now falls back to `currentAsset` whenever nothing concrete resolves, regardless of whether `request` was recognized at all, and the fallback amount handling was reworked to match. A new `ResolvedScanResult.isResolvedPaymentRequest` flag (renamed from `isPaymentRequest`) is true only when the request was both recognized *and* resolved. On the "no error shown to user" half: I looked for an existing toast/snackbar mechanism at the Pay/Swap screen level to surface this and there isn't one — introducing new UI-error machinery felt out of scope for this fix, so I left the state-corruption bug fixed but did not add user-facing messaging. Flagging as a legitimate follow-up if that's wanted.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
zcash-harry: Fixed — resolveAsset() now compares `current` against `candidates` by `assetId` instead of full data-class equality, so an older snapshot missing `contractAddress` no longer fails the "keep current selection" fallback.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
zcash-harry: Fixed — both PayVM.kt and SwapVM.kt now only clear the typed amount when the resolved asset (compared by assetId) actually differs from what's already selected. An amount-less rescan of the same asset leaves the typed amount alone.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/SwapVM.kt
zcash-harry: Fixed alongside the PayVM.kt instance — see the reply there.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/scan/ScanGenericAddressVM.kt
zcash-harry: Fixed — the outer runCatching in onScanned now has an .onFailure that sets state to INVALID, matching the #2441 precedent, so a parser exception (e.g. a native-library fault) no longer leaves the scan screen unresponsive.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/scan/ScanGenericAddressVM.kt
zcash-harry: Fixed — CrossPayRequestParser.parse() now returns a CrossPayParseResult sealed type (Request / NotAPaymentRequest / Rejected) instead of a bare nullable CrossPayRequest. ScanGenericAddressVM now treats Rejected (Solana Pay tx-request links, unrecognised EIP-681 methods) as an invalid scan rather than falling through to treating the raw URI as a literal address.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/NearSwapDataSource.kt
zcash-harry: Fixed — getSupportedTokens()'s dedup key now includes contractAddress, so native/bridged variants sharing symbol+chain+decimals no longer collapse into one.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/swap/SwapVM.kt
zcash-harry: Fixed — resolve() is now only called when the scan isn't a refund-address scan, so the work is skipped entirely rather than computed and discarded.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
zcash-harry: This one's actually the documented intended usage: GetCuratedSwapAssetsUseCase's own KDoc says `invoke()` is for exactly this kind of one-off read, with `observe()` reserved for reactive UI. curate() itself is a cheap single filter pass, and internalState doesn't otherwise hold a cached curated-assets snapshot to reuse without adding new state just for this. Leaving as-is unless you see a reason to deviate from the documented pattern.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/CrossPayRequest.kt
zcash-harry: Agreed this is a real minor duplication (matchesChain vs. isSame), but the semantics genuinely differ (chain-only vs. token+chain), so folding them together would either overload isSame's signature or introduce a redundant token check at the call sites. Leaving as noted -- not a functional bug, and the two idioms are shallow enough that I don't think a shared abstraction pays for itself here.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
zcash-harry: Agreed, still a maintenance trap as originally noted. Left as-is for now since it's low-priority and non-blocking -- happy to clean it up in a follow-up if you'd rather not carry it.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/NavigateToScanGenericAddressUseCase.kt
zcash-harry: Agreed this would be a cleaner model. Given the boolean is only consumed at two call sites and both are correct today, I've left it as a boolean rather than doing the sealed-type refactor in this PR -- flagging as a good follow-up if we want to invest in it.

noop-sk (commented):
Independently re-verified all 17 findings from the two prior review rounds against the current code — including the ones with no visible reply from you, since your Slack message just said "fixed" without specifying which. Good news: all 17 are genuinely closed, no cherry-picking needed.

**Confirmed fixed (15/17), including the ones with no prior visible reply:**
- The asset-wipe bug in `resolve()` (the original "most significant issue") — genuinely gone, `asset = requestedAsset ?: currentAsset` unconditional now. Your decline to add user-facing error UI is reasonable — confirmed there's genuinely no snackbar/toast/errorState mechanism at the Pay/Swap screen level to hook into.
- Solana case-sensitive address matching, the `NumberFormatException` freeze risk, the chain-fallback ordering bug (now unified via `resolveTargetChain()`), the Optimism native-token ambiguity, the refund-address QR-scan overwrite, the redundant chain-comparison duplication, and the `SwapAsset` equality/`contractAddress` regression — all confirmed fixed with the exact mechanisms you described.
- `NearSwapDataSource.kt`'s dedup key now includes `contractAddress` — confirmed, no more silent NEAR-token collapsing.
- Both amount-wipe-on-rescan issues in `PayVM.kt`/`SwapVM.kt` — confirmed fixed, now only clears when the resolved asset actually differs.
- **The most serious open item, `ScanGenericAddressVM.kt`'s uncaught-exception scan-screen freeze (the one mirroring the already-fixed #2441 bug class)** — confirmed fixed: `onScanned` now has `runCatching{}.onFailure { state.update { INVALID } }`, matching the #2441 precedent.
- The Solana-Pay-tx-request-silently-accepted-as-address bug — confirmed fixed: `CrossPayRequestParser.parse()` now returns a sealed `CrossPayParseResult` with a distinct `Rejected` case, correctly routed to `INVALID` instead of falling through to a literal-address accept.
- Grepped the whole diff for the old `isPaymentRequest` name after the rename to `isResolvedPaymentRequest` — zero stale references, clean rename across all 3 call sites.

**Declined, both reasonable (2/17):**
- `ScanResult.address`/`amount` redundant fields — left as low-priority style debt, fine.
- `isPaymentRequest` → renamed to `isResolvedPaymentRequest` but still a raw Boolean, not the sealed model originally suggested — reasonable scope call for this PR.

No new bugs introduced by any of the fixes.

**Not approving yet, for two reasons unrelated to the code review itself:**
1. This is still marked draft, and you said you're doing your own human pass — that should gate the formal approval regardless of my findings.
2. `test_kotlin_modules` is failing for real (not flake): `ui-lib:compileZcashmainnetStoreDebugKotlin` can't resolve `PaymentUriParser`/`Eip681PaymentRequest`/etc. Root cause: `gradle.properties`' `SDK_COMMIT_PIN` points at `f23d5bd21368adc4bd339c57d6bc8eaaa71134c5`, which I confirmed **no longer exists** in `zodl-inc/zodl-android-wallet-sdk` — it's an unreachable commit, almost certainly orphaned by SDK PR #7's squash/rebase from the original #2196. SDK #7 is now approved (head `fbacf6f6`) — once merged, this pin needs repointing to the merged commit before this PR's CI can go fully green. The `static_analysis_detekt` and emulator-test failures are unrelated (a pre-existing `main` baseline issue and infra flake respectively).

Everything else checks out — ready to approve once it's out of draft and the pin is repointed.

`https://github.com/zodl-inc/zodl-android/pull/2457`

noop-sk (approved):
Re-verified after the new commit (head `ea877b2f`). Both blockers from my last COMMENT are resolved:

**SDK_COMMIT_PIN — genuinely fixed.** Now points at `fbacf6f6...`, confirmed reachable in `zodl-inc/zodl-android-wallet-sdk` (SDK #7's approved head). Root cause of the earlier compile break was actually two-fold: the stale sha *and* the companion-SDK checkout action pointing at the wrong repo (`zcash/zcash-android-wallet-sdk` instead of the `zodl-inc` fork) — both fixed. Confirmed the diff between the old and new head is scoped entirely to CI config (`gradle.properties`, `setup-sdk-companion/action.yml`, `pull-request.yml`) — none of the previously-reviewed logic (`CrossPayRequest.kt`, `NavigateToScanGenericAddressUseCase.kt`, `PayVM.kt`, `SwapVM.kt`, `ScanGenericAddressVM.kt`, `NearSwapDataSource.kt`) was touched, so all 17 previously-verified findings stand as fixed with no re-check needed.

The two remaining red checks are both confirmed unrelated to this PR:
- `static_analysis_detekt` — fails on `AutomaticServerRepository.kt:88` (`ReturnCount`), a file not in this PR's diff at all; pre-existing baseline debt on `main` from an unrelated commit (`cdb2c5b4f6`, 2026-08-19). Will block every PR until fixed separately, not specific to this one.
- `test_android_modules_emulator (app)` — confirmed via the actual log: `adb: device offline` → emulator connection lost mid-run (4/12 tests done, 0 failures up to that point) → `AdbCommandRejectedException`/`DeviceException: No connected devices!`. Classic emulator-infra flake, same signature as the PR #2472 flake diagnosed earlier — not an assertion failure, not caused by this PR's code.

No new logic changes to review, no new findings.

`https://github.com/zodl-inc/zodl-android/pull/2457`

PR #2458, opened 2026-08-21
MOB-1750: handle sub-0.01 ZEC Orchard residue for restored/self-migrated wallets

noop-sk:
## Summary
- New home-banner + Migration Complete screen variant ("You've moved to Ironwood") for a
  small unlocked Orchard residue (0.0001-0.01 ZEC) not tied to an unseen in-app migration
  celebration -- covers a fresh wallet restore and an already-acknowledged completion with
  leftover dust. The original "Migration complete" celebration screen is unchanged and keeps
  precedence whenever it hasn't been seen yet, per the ticket.
- `MigrationCompleteArgs` changed from `data object` to `data class(isResidueOnly)`, threaded
  through the VM into `MigrationCompleteState` so the screen renders the right header/summary
  card/recommendation copy without touching the celebration variant's own rendering path.
- Lowers the shared `MIGRATION_DUST_THRESHOLD_ZATOSHI` fallback to 0.0001 ZEC to match the
  companion SDK PR.
- 7 new EN + ES strings matching the ticket's copy verbatim (ES is my own draft translation,
  not vendor-reviewed -- flagging for a native-speaker pass before release).
- Companion SDK PR: zcash/zcash-android-wallet-sdk#2192 (bugfix/MOB-1750 -> maint/v3.1.x).

Linear: https://linear.app/zodl/issue/MOB-1750

Screenshot 2026-08-21 at 13 59 15
Screenshot 2026-08-21 at 13 12 42
Screenshot 2026-08-21 at 13 12 36

## Test plan
- [x] `./gradlew ktlint detektAll` clean (fixed one new `CyclomaticComplexMethod` finding by
      extracting `SummaryCard`, no baseline edits)
- [x] `feature-migration`/`ui-lib` unit test suites pass, including new
      `isResidueOnlyArgPropagatesThroughToState`/`isResidueOnlyDefaultsToFalseForTheOriginalCelebrationVariant`
      coverage for the VM->state wiring
- [x] Full debug APK (`assembleZcashtestnetInternalDebug`) built and installed on an emulator
      against a real testnet wallet with a genuine 0.00545 ZEC Orchard residue -- confirmed the
      new banner/screen render correctly with live data, and (via a temporary, fully-reverted
      local override) confirmed the original celebration screen still renders correctly too

### DIFF
--- a/feature-migration/src/main/java/co/electriccoin/zcash/ui/common/usecase/MigrationHomeMessageSourceImpl.kt
+++ b/feature-migration/src/main/java/co/electriccoin/zcash/ui/common/usecase/MigrationHomeMessageSourceImpl.kt
@@ -41,6 +41,14 @@ import co.electriccoin.zcash.ui.design.R as DesignR
 
 private const val PERCENT_MULTIPLIER = 100
 
+/** The banner phase + title/subtitle override for one [MigrationMessageState] — replaces an
+ * anonymous [Triple] so each field is named at the call site instead of positional. */
+private data class BannerCopy(
+    val phase: MigrationBannerPhase,
+    val title: String?,
+    val subtitle: String?,
+)
+
 /**
  * The home-banner source, fully LIVE off the engine — no plan cache anywhere (see
  * `spec/2026-07-30-plan-cache-elimination-proposal.md`): counts come from the engine's transfer
@@ -137,18 +145,18 @@ class MigrationHomeMessageSourceImpl(
             }
         // Spec §6.2/§6.3 — takes priority over the ordinary phases below: a plan needing
         // re-confirmation is more actionable than its last-known progress/completion state.
-        val (phase, title, subtitle) =
+        val bannerCopy =
             when (data.attentionKind) {
                 MigrationAttentionKind.PLAN_UPDATE -> {
-                    Triple(MigrationBannerPhase.ATTENTION, "Update migration plan", "Tap to review the details")
+                    BannerCopy(MigrationBannerPhase.ATTENTION, "Update migration plan", "Tap to review the details")
                 }
 
                 MigrationAttentionKind.TRANSFER_EXPIRED -> {
                     val range = data.attentionRangeText
-                    Triple(
-                        MigrationBannerPhase.ATTENTION,
-                        if (range != null) "Transfer $range expired" else "A transfer expired",
-                        "Tap to review the details",
+                    BannerCopy(
+                        phase = MigrationBannerPhase.ATTENTION,
+                        title = if (range != null) "Transfer $range expired" else "A transfer expired",
+                        subtitle = "Tap to review the details",
                     )
                 }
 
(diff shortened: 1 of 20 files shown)


feature-migration/src/main/java/co/electriccoin/zcash/ui/common/usecase/MigrationHomeMessageSourceImpl.kt
```diff
@@ -148,6 +154,18 @@ class MigrationHomeMessageSourceImpl(
 
                 null -> {
                     when {
+                        // MOB-1750: a small leftover Orchard balance not tied to an unseen in-app
+                        // migration celebration gets its own "X ZEC left in Orchard" copy instead of
+                        // the "Migration complete" celebration title.
+                        data.isComplete && data.isResidueOnly -> {
+                            val amount = stringRes(Zatoshi(data.residualBalanceZatoshi)).getString(context)
+                            Triple(
```
nesence-m: data classes are preferred over Triple

nesence-m (changes requested):

feature-migration/src/main/java/co/electriccoin/zcash/ui/screen/migration/complete/MigrationCompleteScreen.kt
```diff
@@ -269,6 +267,54 @@ fun MigrationCompleteView(state: MigrationCompleteState) {
...
+                .background(ZashiColors.Surfaces.bgSecondary)
+                .padding(20.dp),
+        verticalArrangement = Arrangement.spacedBy(10.dp),
+    ) {
+        if (state.isResidueOnly) {
+            // MOB-1750: reduced summary — no Transfers/Duration rows, since a residue not tied
+            // to an unseen in-app celebration has no round-trip campaign to report on (whether
+            // it's genuinely 0 sent-in-app, or an already-seen completion's leftover residue).
+            SummaryRow(
+                label = stringRes(DesignR.string.migrationComplete_inIronwoodLabel).getValue(),
```
nesence-m, line 286: **"In Ironwood" reads 0 on restore**

For a fresh wallet restore — one of the two cases this PR targets — `state.totalTransferred` comes from the engine's persisted summary (`getMigrationSummary()?.totalMigratedZatoshi ?: 0L` in `MigrationCompleteVM`), which is null for a wallet migrated on another device. The residue card then shows "In Ironwood: 0 ZEC" under the "You've moved to Ironwood" title while the wallet's real Ironwood balance is nonzero. Consider showing the live Ironwood balance in the residue variant, or hiding the row when the summary is null.

feature-migration/src/main/java/co/electriccoin/zcash/ui/common/usecase/MigrationHomeMessageSourceImpl.kt
```diff
@@ -148,6 +154,18 @@ class MigrationHomeMessageSourceImpl(
 
                 null -> {
                     when {
+                        // MOB-1750: a small leftover Orchard balance not tied to an unseen in-app
+                        // migration celebration gets its own "X ZEC left in Orchard" copy instead of
+                        // the "Migration complete" celebration title.
+                        data.isComplete && data.isResidueOnly -> {
+                            val amount = stringRes(Zatoshi(data.residualBalanceZatoshi)).getString(context)
```
nesence-m, line 179: **Eagerly resolved banner strings**

Injecting `Context` to resolve these with `.getString(context)` at the source layer works, but the repo convention keeps display text as `StringResource` resolved in composables — the root cause is `MigrationMessageState`'s pre-existing raw-`String` title/progressLabel fields, so this PR isn't the place to fix it. Worth a follow-up migrating those fields to `StringResource`, which would also drop the `Context` dependency from this class and keep the banner correct across locale changes.

nesence-m (commented):
The structure checks out: the celebration branch (`Complete && !hasSeenComplete`) still precedes the residue branch as the description claims, the Koin `singleOf`/`viewModelOf` DSL absorbs the new `Context`/`args` constructor params without module edits, and no `MigrationCompleteArgs` call site was missed by the `data object → data class` change.

One coordination point at the PR level: `migrationMessageFor` takes the dust threshold from the live SDK (`migrationDustThresholdZatoshi()`), with the app constant only a `getOrDefault` fallback — and this PR doesn't bump the SDK dependency. Until a v3.1.x SDK release containing zcash/zcash-android-wallet-sdk#2192 is picked up here, the residue banner still gates at 0.001 ZEC while ui-lib callers of the updated constant (e.g. `shouldShowIronwoodAnnouncement`) already gate at 0.0001 — the two disagree in the 10k–100k zatoshi gap. Fine if both ship in the same v3.10.x release; could you note the intended landing order?

Two comments inline.

feature-migration/src/main/java/co/electriccoin/zcash/ui/common/usecase/MigrationHomeMessageSourceImpl.kt
noop-sk: Fixed in b85fe1747 — extracted a private `BannerCopy(phase, title, subtitle)` data class, replacing the anonymous `Triple`.

feature-migration/src/main/java/co/electriccoin/zcash/ui/screen/migration/complete/MigrationCompleteScreen.kt
noop-sk: Fixed in aad653ff — added `GetIronwoodBalanceUseCase` (same live-synchronizer source the Balance Breakdown sheet already uses) and the residue variant's total now reads from that instead of `getMigrationSummary()`. Confirmed live on a real testnet wallet: verified the engine's `orchard_ironwood_migration_transactions`/`crossing_values` tables were genuinely empty (0 rows) despite real prior migration activity — root cause is `proposeImmediateMigration()` ("Migrate anyway") being documented to never persist to that store, so `getMigrationSummary()` can never account for transfers sent that way. Added a regression test (`residueVariantReadsTotalTransferredFromLiveIronwoodBalanceNotEngineSummary`) pinning that the engine summary is never even queried for this variant.

feature-migration/src/main/java/co/electriccoin/zcash/ui/common/usecase/MigrationHomeMessageSourceImpl.kt
noop-sk: Agreed, leaving as-is here per your note — this class already had the same eager-Context pattern for its other phases (ATTENTION, IN_PROGRESS, READY_TO_SEND) before this PR, so MOB-1750 is just following existing precedent rather than introducing it. Filed as a follow-up: migrate `MigrationMessageState.title`/`progressLabel` from raw `String` to `StringResource` so resolution happens in the composable and this class can drop its `Context` dependency entirely.

nesence-m (dismissed):
utACK

nesence-m (approved):

PR #2459, opened 2026-08-21
MOB-1723: show the Insufficient Funds sheet for swap proposal failures on every engine

nesence-m:
Fixes the app half of [MOB-1723](https://linear.app/zodl/issue/MOB-1723/android-insufficient-balance-calculation-appears-off-not-accounting): five users on v3.9.2/v3.9.3 reported `Insufficient balance (have X, need Y including fee)` surfacing on the technical error screen during swaps instead of the friendly Insufficient Funds sheet.

Companion SDK PR (must merge first): [zcash/zcash-android-wallet-sdk#2194 MOB-1723: report insufficient funds and multi-step PCZT proposals as typed exceptions](https://github.com/zcash/zcash-android-wallet-sdk/pull/2194)

## Root causes & what changed

**1. Error mapping only worked on one engine.** `getOrThrow` recognized insufficient funds by string-matching the root cause of `ProposalFromParametersException` — which the Slipstream engine never produced (it delegated to `RustBackend` bare, so a raw `RuntimeException` escaped to the generic catch → technical error screen). The SDK now throws a typed `TransactionEncoderException.InsufficientFundsException` from **both** engines, and `getOrThrow` maps purely by type; all message matching is deleted.

**2. TEX-on-Keystone check moved to its true home.** The app's propose-time `validate()` (`KeystoneAccount` + `AddressType.Tex`) is replaced by mapping the SDK's new `PcztException.MultiStepProposalUnsupportedException` (structural step-count rejection of ZIP-320/TEX proposals) to `TexUnsupportedOnKSException` at PCZT creation. Every Keystone flow creates the PCZT inside the same `try` that catches it, so user-visible behavior is unchanged.

**3. Balance was validated only at typing time.** `requestExactInput` now re-checks `canSpend` as its first statement — before the shielded-address rotation and the provider quote round-trip — and forwards to the Insufficient Funds sheet on failure. The typing-time checks in the swap and pay mappers are unified on the same `WalletAccount.canSpend` primitive via `InternalState.canSpend`. Post-quote drift (and exact-output, where the ZEC cost is only known after the quote) is handled by the Rust check arriving as the typed exception.

Out of scope (tracked separately as the fee-headroom part of MOB-1723): the strict check still allows amount == spendable, which then fails on fees — but it now fails onto the friendly sheet.

## SDK pin

The companion SDK PR has **merged**; `SDK_COMMIT_PIN` points at the merged sha on `origin/maint/v3.1.x` (`4b39e71f`, the merge commit of #2194), per the pin comment's contract. Release branches additionally need the pin cleared and a published SDK release (v3.1.x point release) — the standing rule.

## Tests

- New `ProposalDataSourceTest` (6): typed insufficient-funds mapping, PCZT TEX mapping, pass-through and no-double-wrap cases.
- `RequestSwapQuoteUseCaseTest`: +3 (pre-quote block for Zashi and Keystone with no address rotation and no quote request; flex-input performs no balance check), fixtures extended with `canSpend` stubs.
- Full `ui-lib` suite 433/433 green against the pinned SDK; detekt + ktlint clean; whole-app compile verified.

---

> **Note**

# Author

> **Note**

# Reviewer

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
@@ -18,7 +18,6 @@ import cash.z.ecc.android.sdk.model.proposeSend
 import cash.z.ecc.android.sdk.type.AddressType
 import co.electriccoin.lightwallet.client.model.LightWalletEndpoint
 import co.electriccoin.zcash.spackle.Twig
-import co.electriccoin.zcash.ui.common.model.KeystoneAccount
 import co.electriccoin.zcash.ui.common.model.NetworkDimension
 import co.electriccoin.zcash.ui.common.model.SubmitResult
 import co.electriccoin.zcash.ui.common.model.SwapQuote
@@ -35,36 +34,20 @@ import org.zecdev.zip321.parser.ParserContext
 import java.math.BigDecimal
 
 interface ProposalDataSource {
-    @Throws(
-        TransactionProposalNotCreatedException::class,
-        InsufficientFundsException::class,
-        TexUnsupportedOnKSException::class
-    )
+    @Throws(TransactionProposalNotCreatedException::class, InsufficientFundsException::class)
     suspend fun createProposal(account: WalletAccount, send: ZecSend): RegularTransactionProposal
 
-    @Throws(
-        TransactionProposalNotCreatedException::class,
-        InsufficientFundsException::class,
-        TexUnsupportedOnKSException::class
-    )
+    @Throws(TransactionProposalNotCreatedException::class, InsufficientFundsException::class)
     suspend fun createZip321Proposal(account: WalletAccount, zip321Uri: String): Zip321TransactionProposal
 
-    @Throws(
-        TransactionProposalNotCreatedException::class,
-        InsufficientFundsException::class,
-        TexUnsupportedOnKSException::class
-    )
+    @Throws(TransactionProposalNotCreatedException::class, InsufficientFundsException::class)
     suspend fun createExactInputProposal(
         account: WalletAccount,
         send: ZecSend,
         quote: SwapQuote,
     ): ExactInputSwapTransactionProposal
 
-    @Throws(
-        TransactionProposalNotCreatedException::class,
-        InsufficientFundsException::class,
-        TexUnsupportedOnKSException::class
-    )
+    @Throws(TransactionProposalNotCreatedException::class, InsufficientFundsException::class)
     suspend fun createExactOutputProposal(
         account: WalletAccount,
         send: ZecSend,
@@ -74,7 +57,7 @@ interface ProposalDataSource {
     @Throws(TransactionProposalNotCreatedException::class, InsufficientFundsException::class)
     suspend fun createShieldProposal(account: WalletAccount): ShieldTransactionProposal
 
-    @Throws(PcztException.CreatePcztFromProposalException::class)
+    @Throws(PcztException.CreatePcztFromProposalException::class, TexUnsupportedOnKSException::class)
     suspend fun createPcztFromProposal(account: WalletAccount, proposal: Proposal): Pczt
 
     @Throws(PcztException.AddProofsToPcztException::class)
@@ -92,7 +75,9 @@ class TransactionProposalNotCreatedException(
     cause: Exception
 ) : Exception(cause)
 
-class InsufficientFundsException : Exception()
+class InsufficientFundsException(
+    cause: Throwable? = null
+) : Exception(cause)
 
 class TexUnsupportedOnKSException : Exception("TEX addresses are unsupported on Keystone")
 
@@ -106,7 +91,6 @@ class ProposalDataSourceImpl(
     override suspend fun createProposal(account: WalletAccount, send: ZecSend): RegularTransactionProposal =
         withContext(Dispatchers.IO) {
             getOrThrow { synchronizer ->
-                validate(account, synchronizer, send.destination.address)
                 RegularTransactionProposal(
                     destination = send.destination,
                     amount = send.amount,
@@ -140,8 +124,6 @@ class ProposalDataSourceImpl(
                         )
                     }
 
-                validate(account, synchronizer, payment.recipientAddress.value)
-
                 if (payment.nonNegativeAmount == null) {
                     throw TransactionProposalNotCreatedException(IllegalArgumentException("Null amount"))
                 }
@@ -171,7 +153,6 @@ class ProposalDataSourceImpl(
     ): ExactInputSwapTransactionProposal =
         withContext(Dispatchers.IO) {
             getOrThrow { synchronizer ->
-                validate(account, synchronizer, send.destination.address)
                 ExactInputSwapTransactionProposal(
                     destination = send.destination,
                     amount = send.amount,
@@ -189,7 +170,6 @@ class ProposalDataSourceImpl(
     ): ExactOutputSwapTransactionProposal =
         withContext(Dispatchers.IO) {
             getOrThrow { synchronizer ->
-                validate(account, synchronizer, send.destination.address)
                 ExactOutputSwapTransactionProposal(
                     destination = send.destination,
                     amount = send.amount,
@@ -221,11 +201,14 @@ class ProposalDataSourceImpl(
 
     override suspend fun createPcztFromProposal(account: WalletAccount, proposal: Proposal): Pczt =
         withContext(Dispatchers.IO) {
-            val synchronizer = synchronizerProvider.getSynchronizer()
-            synchronizer.createPcztFromProposal(
-                accountUuid = account.sdkAccount.accountUuid,
-                proposal = proposal
-            )
+            try {
+                synchronizerProvider.getSynchronizer().createPcztFromProposal(
+                    accountUuid = account.sdkAccount.accountUuid,
+                    proposal = proposal
+                )
+            } catch (_: PcztException.MultiStepProposalUnsupportedException) {
+                throw TexUnsupportedOnKSException()
+            }
         }
 
     override suspend fun addProofsToPczt(pczt: Pczt): Pczt =
@@ -257,20 +240,6 @@ class ProposalDataSourceImpl(
                 .getSynchronizer()
                 .redactPcztForSigner(pczt)
         }
-
-    /**
-     * @throws TexUnsupportedOnKSException if the address is a TEX address and the account is a Keystone account
-     */
-    @Throws(TexUnsupportedOnKSException::class)
-    private suspend fun validate(
-        account: WalletAccount,
-        synchronizer: Synchronizer,
-        address: String
-    ) {
-        if (account is KeystoneAccount && synchronizer.validateAddress(address) == AddressType.Tex) {
-            throw TexUnsupportedOnKSException()
-        }
-    }
 
     @Suppress("CyclomaticComplexMethod", "TooGenericExceptionCaught")
     private suspend fun submitTransactionInternal(
@@ -355,17 +324,8 @@ class ProposalDataSourceImpl(
         try {
             val synchronizer = synchronizerProvider.getSynchronizer()
             block(synchronizer)
-        } catch (e: TransactionEncoderException.ProposalFromParametersException) {
-            val message = e.rootCause.message ?: ""
-            if (message.contains("Insufficient balance", true) ||
-                message.contains("The transaction requires an additional change output of", true)
-            ) {
-                throw InsufficientFundsException()
-            } else {
-                throw TransactionProposalNotCreatedException(e)
-            }
-        } catch (e: TexUnsupportedOnKSException) {
-            throw e
+        } catch (e: TransactionEncoderException.InsufficientFundsException) {
+            throw InsufficientFundsException(e)
         } catch (e: TransactionProposalNotCreatedException) {
             throw e
         } catch (e: Exception) {
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
@@ -57,7 +57,16 @@ interface ProposalDataSource {
     @Throws(TransactionProposalNotCreatedException::class, InsufficientFundsException::class)
     suspend fun createShieldProposal(account: WalletAccount): ShieldTransactionProposal
 
-    @Throws(PcztException.CreatePcztFromProposalException::class, TexUnsupportedOnKSException::class)
+    /**
+     * Propagates the SDK's [PcztException.MultiStepProposalUnsupportedException] untouched: whether a
+     * multi-step rejection means "TEX is unsupported on Keystone" depends on what the proposal was
+     * for, which only the caller knows (see
+     * [co.electriccoin.zcash.ui.common.repository.KeystoneProposalRepository.createPCZTFromProposal]).
+     */
+    @Throws(
+        PcztException.CreatePcztFromProposalException::class,
+        PcztException.MultiStepProposalUnsupportedException::class
+    )
     suspend fun createPcztFromProposal(account: WalletAccount, proposal: Proposal): Pczt
 
     @Throws(PcztException.AddProofsToPcztException::class)
@@ -201,14 +210,10 @@ class ProposalDataSourceImpl(
 
     override suspend fun createPcztFromProposal(account: WalletAccount, proposal: Proposal): Pczt =
         withContext(Dispatchers.IO) {
-            try {
-                synchronizerProvider.getSynchronizer().createPcztFromProposal(
-                    accountUuid = account.sdkAccount.accountUuid,
-                    proposal = proposal
-                )
-            } catch (_: PcztException.MultiStepProposalUnsupportedException) {
-                throw TexUnsupportedOnKSException()
-            }
+            synchronizerProvider.getSynchronizer().createPcztFromProposal(
+                accountUuid = account.sdkAccount.accountUuid,
+                proposal = proposal
+            )
         }
 
     override suspend fun addProofsToPczt(pczt: Pczt): Pczt =
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RequestSwapQuoteUseCase.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RequestSwapQuoteUseCase.kt
@@ -1,5 +1,6 @@
 package co.electriccoin.zcash.ui.common.usecase
 
+import cash.z.ecc.android.sdk.ext.convertZecToZatoshi
 import cash.z.ecc.android.sdk.model.Memo
 import cash.z.ecc.android.sdk.model.WalletAddress
 import cash.z.ecc.android.sdk.model.Zatoshi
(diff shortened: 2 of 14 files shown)


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RequestSwapQuoteUseCase.kt
```diff
@@ -47,6 +48,11 @@ class RequestSwapQuoteUseCase(
         slippage: BigDecimal,
         canNavigateToSwapQuote: () -> Boolean
     ) {
+        if (!accountDataSource.getSelectedAccount().canSpend(amount.convertZecToZatoshi())) {
```
noop-sk: **Pre-check has three gaps versus the equivalent post-failure path**

1. This early return navigates to `InsufficientFundsArgs` without clearing `swapRepository`/`keystoneProposalRepository`/`zashiProposalRepository`, unlike the `InsufficientFundsException` catch further down in this file (lines 137-139) which clears all three. If a user got a quote/proposal once, then edits the amount to something now unaffordable and retries, this short-circuit leaves the previous quote/proposal stashed as stale live state in the repositories.
2. `accountDataSource.getSelectedAccount()` is read here and again independently inside `requestNextShieldedAddress()`/`createProposal()`. If the user switches the selected account between tapping "Get quote" and the async response, this check validates account A while the refund address and proposal end up built against a different live read — defeating the point of the pre-check for the account actually used.
3. This uses `cash.z.ecc.android.sdk.ext.convertZecToZatoshi`, which throws `IllegalArgumentException` on a negative `BigDecimal`, while every other ZEC→Zatoshi conversion in this same swap-from-ZEC path (`SwapVMMapper.kt`, `ExactOutputVMMapper.kt`) uses the local `co.electriccoin.zcash.ui.screen.swap.convertZecToZatoshi`, which clamps into `[0, 21_000_000]` and never throws. Worth switching to the local one for consistency and to avoid an uncaught throw if `amount` is ever negative or out of range.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
```diff
@@ -221,11 +201,14 @@ class ProposalDataSourceImpl(
...
-            synchronizer.createPcztFromProposal(
-                accountUuid = account.sdkAccount.accountUuid,
-                proposal = proposal
-            )
+            try {
+                synchronizerProvider.getSynchronizer().createPcztFromProposal(
+                    accountUuid = account.sdkAccount.accountUuid,
+                    proposal = proposal
+                )
+            } catch (_: PcztException.MultiStepProposalUnsupportedException) {
```
noop-sk: **Remap is too broad for a shared chokepoint**

`createPcztFromProposal` unconditionally remaps every `PcztException.MultiStepProposalUnsupportedException` to `TexUnsupportedOnKSException`, but this method is also called by the Keystone migration-sweep flows (`MigrationCompleteVM.kt`, `MigrationReviewVM.kt`) and by `ShieldFundsUseCase`, none of which have anything to do with TEX. Given the app's own Keystone batch cap (96 actions/round), a migration sweep or shield proposal touching many notes/UTXOs is a realistic way to need more than one PCZT step for reasons unrelated to TEX. A Keystone user shielding funds with no TEX address anywhere in the flow could end up seeing an error dialog reading "TEX addresses are unsupported on Keystone." Consider keeping the TEX-specific remap scoped to the swap/pay call sites rather than this shared method.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
```diff
@@ -317,6 +317,13 @@ internal interface InternalState {
 
     val totalSpendableBalance: Zatoshi
         get() = account?.spendableShieldedBalance ?: Zatoshi(0)
+
+    /**
+     * The single spendability primitive the whole app validates against, so the typing-time check and
+     * the proposal-time check in the SDK can never disagree. Re-evaluated whenever the selected
+     * account emits a new balance.
+     */
+    fun canSpend(amount: Zatoshi): Boolean = account?.canSpend(amount) ?: false
```
noop-sk: **Duplicated across `PayVM` and `SwapVM`**

`fun canSpend(amount: Zatoshi): Boolean = account?.canSpend(amount) ?: false` is added verbatim here and in `SwapVM.InternalState` (`SwapVM.kt:384`) — both interfaces already independently duplicate the adjacent `totalSpendableBalance` property the same way. A `WalletAccount?` extension (or a shared account-view-state contract) would keep the null-fallback semantics in one place instead of two copies that nothing enforces stay in sync.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/ExactOutputVMMapper.kt
```diff
@@ -103,7 +103,7 @@ internal class ExactOutputVMMapper {
         return stringResByDynamicCurrencyNumber(state.getZec() ?: BigDecimal(0), CURRENCY_TICKER).withStyle(
             StyledStringStyle(
                 color =
-                    if (zatoshi != null && state.totalSpendableBalance < zatoshi) {
+                    if (zatoshi != null && !state.canSpend(zatoshi)) {
```
noop-sk: **Same check recomputed at 4 call sites**

`zatoshi != null && !state.canSpend(zatoshi)` (and the `BigDecimal`-heavy `state.getZatoshi()` it depends on) is recomputed independently here and at lines 140, 184 and 226 instead of once as a derived property on `ExactOutputInternalState`. Beyond the repeated work per recomposition, a future change to the insufficient-funds rule needs editing all four near-identical sites, risking drift between them.

noop-sk (changes requested):
Reviewed the app half of MOB-1723 (draft, pending SDK #2194). The typed-exception mapping direction is right, but a few things are worth a look before this comes out of draft. Left inline comments on the specific lines; three more notes here since they land outside this diff's changed lines but are directly relevant to what this PR changes:

- `RequestSwapQuoteUseCase.requestExactOutput` (line 72, the Pay/exact-output flow) has no pre-quote `canSpend` re-check, unlike the sibling `requestExactInput` this PR adds it to (line 51) — even though `ExactOutputVMMapper` has the identical typing-time `canSpend` check and thus the identical balance-drop-after-typing race the CHANGELOG describes closing. Worth adding the same early check there for consistency.
- The `catch (_: TexUnsupportedOnKSException)` block in `requestQuote()` (line 132) is missing `return@withContext`, unlike the `InsufficientFundsException` and generic `Exception` catches right below it. It's pre-existing code, but this PR's `ProposalDataSource` change (see inline comment) makes this exception reachable from more call paths, and as written the code falls through after clearing the proposal repositories to `if (canNavigateToSwapQuote()) navigationRouter.forward(SwapQuoteArgs)`, double-navigating onto a swap-quote screen backed by state that was just cleared.
- `WalletAccount.canSpend()` (`ui-lib/.../common/model/WalletAccount.kt:79`, unchanged by this PR) is a fee-blind raw comparison (`spendableShieldedBalance >= amount`), which the PR description itself calls out: for a balance sitting exactly at that boundary, the new pre-quote check still passes and the app still pays for the full async quote round trip before the SDK's fee-aware post-quote check catches it — the round-trip-avoidance benefit silently doesn't apply to the boundary case the PR names as its known remaining gap.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RequestSwapQuoteUseCase.kt
nesence-m: Fixed in 9cc1043f3 (points 1 and 3):

1. The pre-check block now clears the quote and both proposal repositories exactly like the proposal-time catch, and `assertBlockedBeforeTheQuote` in the test verifies the clears.
3. Switched to the local clamping `co.electriccoin.zcash.ui.screen.swap.convertZecToZatoshi`, consistent with the rest of the swap-from-ZEC path — no uncaught throw on a negative/out-of-range amount.

On point 2 (account switched mid-flight): fully closing that race would mean threading an account snapshot through `requestNextShieldedAddress()` and both proposal repositories' public APIs, which all do live reads today (also on the paths that had no pre-check before this PR). Instead I documented the intended layering in KDoc: the pre-check is a best-effort fast path against a snapshot; the authoritative check is the SDK's typed insufficient-funds failure at proposal time, which is evaluated against the very account the proposal is built for and lands on the same Insufficient Funds sheet. So even when the race hits, the user still ends up on the right sheet rather than a wrong send. Happy to do the snapshot-threading refactor as a follow-up if you think it's worth the API churn.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
nesence-m: Fixed in 9cc1043f3. `ProposalDataSource.createPcztFromProposal` no longer remaps anything — the raw `MultiStepProposalUnsupportedException` propagates, and the remap moved to `KeystoneProposalRepository.createPCZTFromProposal`, which already holds the current `TransactionProposal`. It maps to `TexUnsupportedOnKSException` only when the proposal is a `SendTransactionProposal` whose destination is `WalletAddress.Tex` — the one flow where a multi-step proposal is the expected shape. Shielding and migration-sweep proposals (no TEX destination, or no destination at all) rethrow the raw exception and land on the generic error path. Covered by three new `KeystoneProposalRepositoryTest` cases (TEX payment, non-TEX payment, migration sweep) plus the updated datasource pass-through test.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
nesence-m: Fixed in 9cc1043f3 — both `canSpend` and the adjacent `totalSpendableBalance` now live as nullable `WalletAccount?` extensions next to the model, and both `InternalState`s delegate to them, so the null-fallback semantics exist in exactly one place.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/ExactOutputVMMapper.kt
nesence-m: Fixed in 9cc1043f3 — the check is now a single `isInsufficientFunds` property on `ExactOutputInternalState`, computed once per state instance; all four sites render from it.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RequestSwapQuoteUseCase.kt
noop-sk: Verified against head `33544bc`: (1a) repo-clearing is fixed — the early return now clears all three repositories before navigating. (1c) is fixed — now uses the local clamped `convertZecToZatoshi`. (1b) the account-switch race — I understand the call: the SDK's own check is authoritative regardless, so this is a documented best-effort fast path, not a correctness gap. Fine as accepted.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/ProposalDataSource.kt
noop-sk: Verified against head `33544bc`: fixed, and better than the narrow fix I'd have expected — the remap moved out of the shared `ProposalDataSource.createPcztFromProposal` entirely and into `KeystoneProposalRepository`, scoped by an actual `paysTexAddress()` check. Confirmed `ShieldTransactionProposal`/`MigrationSweepTransactionProposal` aren't `SendTransactionProposal`, so shield/migration-sweep proposals now correctly rethrow the raw exception instead of getting the TEX-specific message. Nice fix.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/PayVM.kt
noop-sk: Verified against head `33544bc`: fixed — consolidated into a `WalletAccount?.canSpend()` extension, both VMs now delegate to it.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/pay/ExactOutputVMMapper.kt
noop-sk: Verified against head `33544bc`: fixed — `isInsufficientFunds` is now a single derived property on `ExactOutputInternalState`, consumed at all 4 former call sites. Confirmed logically equivalent to the old inline check.

noop-sk (changes requested):
Re-reviewed the fix commit (`9cc1043f3`). All 4 inline threads verified — replied on each. 3 of 4 fully fixed, 1 (the account-switch race) reasonably accepted as a documented best-effort fast path since the SDK's own check is authoritative. The TEX-remap fix (#2) in particular is a better fix than what I originally asked for.

Two of the three body-only notes from the first pass are still genuinely open and weren't part of the '4 comments addressed' — flagging them again since they weren't inline-commentable at the time and still aren't (the touched lines remain unchanged in this diff):

1. `RequestSwapQuoteUseCase.requestExactOutput` (currently around line 82) still has no pre-quote `canSpend` check, unlike `requestExactInput`. Since `ExactOutputVMMapper` has the identical typing-time check, this flow still pays for the full async quote round-trip on a balance that's already known to be insufficient.
2. The `catch (_: TexUnsupportedOnKSException)` block in `requestQuote()` (currently around line 142) is still missing `return@withContext`, unlike the two catches right below it. This is now more reachable than when I first flagged it, since the TEX-remap fix (#2) means this exception now only fires for genuine TEX-on-Keystone cases rather than any multi-step proposal — but the double-navigation-after-clearing-state bug on that path is unchanged.

The third body-only note (fee-blind `WalletAccount.canSpend()` boundary case) is unchanged as expected — the PR description already tracks it as a known, separate follow-up, so no action needed there.

Given #1 and #2 above are real and unaddressed, I'll leave this as requesting changes, though the PR is close.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RequestSwapQuoteUseCase.kt
```diff
@@ -63,13 +79,28 @@ class RequestSwapQuoteUseCase(
...
+     * lands on the same Insufficient Funds sheet.
+     */
     suspend fun requestExactOutput(
         amount: BigDecimal,
         address: String,
         selectedAsset: SwapAsset,
         slippage: BigDecimal,
         canNavigateToSwapQuote: () -> Boolean
     ) {
+        if (!accountDataSource.getSelectedAccount().canSpend(amount.convertZecToZatoshi())) {
```
noop-sk: **Currency-unit mismatch — `amount` here isn't ZEC**

`requestExactInput`'s `amount` genuinely is a ZEC quantity (confirmed at its call site, `SwapVM.onRequestSwapQuoteClick` for SWAP_FROM_ZEC), so `amount.convertZecToZatoshi()` is correct there. But `requestExactOutput`'s `amount` comes from `PayVM.onRequestSwapQuoteClick` as `state.getOriginTokenAmount()` — the raw amount typed in the **destination asset's** units (e.g. BTC/USDC), not ZEC. `ExactOutputVMMapper.ExactOutputInternalState.getZatoshi()` converts this exact same raw amount via a full USD cross-rate (`amountToken.multiply(asset.usdPrice).divide(zecAsset.usdPrice).convertZecToZatoshi()`) precisely because it isn't already ZEC.

So this guard currently calls `canSpend(amount.convertZecToZatoshi())` treating e.g. "0.01 BTC" as "0.01 ZEC" — depending on the destination asset's relative price this either false-blocks an affordable payment or lets an unaffordable one through to the full round trip. The new tests stub `canSpend` with `any()`, so they verify the branch was taken but not that the correct zatoshi value reached it — worth adding an assertion on the actual value passed. Needs the same USD cross-rate conversion `ExactOutputVMMapper.getZatoshi()` already does before this check.

noop-sk (changes requested):
Re-reviewed the latest fix commit. The `TexUnsupportedOnKSException` catch fix is solid — verified `TEXUnsupportedVM`/`TEXUnsupportedArgs`/`TEXUnsupportedState` read none of the three repositories, so clearing them before navigating there is safe, and the new regression tests (`verify(exactly = 0) { navigationRouter.forward(SwapQuoteArgs) }`) would have caught the original bug.

The `requestExactOutput` pre-check is structurally a correct mirror of `requestExactInput`, but that's exactly the problem — see inline comment for a currency-unit bug it introduces.

Also: this PR currently shows as not mergeable (`mergeable_state: dirty`) — looks like `maint/v3.10.x` has moved forward since the last merge-in commit and this needs a fresh merge/rebase before it can land, independent of the review findings.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/RequestSwapQuoteUseCase.kt
nesence-m: Addressed in two steps, both pushed:

1. 649e72cf3 fixed exactly this: the exact-output guard went through the same USD cross-rate conversion `ExactOutputVMMapper.getZatoshi()` uses (skipping the pre-check when either price wasn't loaded yet), and the tests were strengthened to assert the actual `Zatoshi` value reaching `canSpend()` instead of stubbing with `any()`.
2. c3eb8e09d then dropped the pre-quote `canSpend` fast path from **both** `requestExactInput` and `requestExactOutput` entirely. The SDK's typed insufficient-funds failure at proposal time is the authoritative check and lands on the same sheet, and the exact-output variant needed a price-dependent estimate just to approximate the answer — this thread being evidence the fast path wasn't worth its complexity for a round-trip optimization. The typing-time checks in the mappers and the typed-exception mapping are untouched, so the user-visible MOB-1723 fix is unchanged; CHANGELOG reworded to match.

Also merged `origin/maint/v3.10.x` in ffc8f7082, which clears the dirty mergeable state; `SDK_COMMIT_PIN` now rides the maint pin (0ef67b411), which contains SDK #2194. The three touched test classes (`RequestSwapQuoteUseCaseTest`, `ProposalDataSourceTest`, `KeystoneProposalRepositoryTest`) are green against that pin.

noop-sk (approved):
Re-verified both open findings from the last review round against the current diff (head `1823c3e7e`), not just the reply text:

1. **`requestExactOutput` cross-rate bug** — fixed, but not the way I expected: `649e72cf3` first patched it narrowly (routed through the same `ExactOutputVMMapper.getZatoshi()` cross-rate conversion), then `c3eb8e09d` removed the pre-quote `canSpend` fast path entirely from both `requestExactInput` and `requestExactOutput`. Confirmed in `RequestSwapQuoteUseCase.kt:43-87` — neither function has any balance pre-check now. Balance is enforced by the typing-time check in `ExactOutputVMMapper.kt:419` (correctly uses `getZatoshi()`) and by the SDK's typed `InsufficientFundsException` at proposal time. The unit-mismatched code path no longer exists, so the original bug can't fire.
2. **`TexUnsupportedOnKSException` missing `return@withContext`** — fixed. All three catch blocks in `requestQuote` (`RequestSwapQuoteUseCase.kt:126-144`) now end with `return@withContext`.

Rebase against `maint/v3.10.x` is done (no longer `mergeable: dirty`).

One cosmetic note, not blocking: the PR description still says "requestExactInput now re-checks canSpend as its first statement" — that describes the fast path removed in `c3eb8e09d`, so it's stale relative to the current diff. Worth a quick edit for anyone reading the description later, but doesn't affect the merged code.

`https://github.com/zodl-inc/zodl-android/pull/2459`

PR #2460, opened 2026-08-21
[#MOB-1690] Use monotonic clock for app-lock timeout

nesence-m:
Fixes Least Authority finding C ([MOB-1690](https://linear.app/zodl/issue/MOB-1690/c-application-lock-screen-can-be-bypassed-by-changing-system-clock)): the app-lock re-authentication timeout compared wall-clock timestamps (`System.currentTimeMillis`), and the check only ever clears the in-memory auth state — so setting the device clock backwards while the app was backgrounded made the delta negative and the lock screen never reappeared.

## Changes

- The backgrounded-at timestamp is now sampled from a new injectable `GetMonotonicTimeProvider` backed by `SystemClock.elapsedRealtime()` (monotonic, cannot be changed by the user) and held in-memory in `AuthenticationViewModel`; both the record and the check are plain synchronous calls (no coroutine/DataStore round-trip, which also closes a brief flash-of-content window on foregrounding).
- Removed the persisted `LATEST_APP_BACKGROUND_TIME_MILLIS` preference: all authentication state is in-memory in the activity-scoped ViewModel, so process death always re-locks regardless — the persisted value never provided protection, and dropping it avoids the elapsedRealtime-resets-at-boot ambiguity.
- Unit tests for the timeout path (previously uncovered): recent background keeps the unlocked state, >15 min background resets it, check without a recorded background time is a no-op.

## Testing

- `AuthenticationViewModelTest`: 9/9 green (6 existing + 3 new).
- `detektAll` and `ktlintFormat` clean.
- Manual: background the app, set the device clock back several hours, reopen — the lock still appears after 15 minutes of real elapsed time; a forward clock change no longer force-locks within the window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ### Fixed:
 
+- The app-lock re-authentication timeout is now measured with a monotonic clock instead of the system wall clock, so it can no longer be
+  bypassed by changing the device's date and time.
 - We polished several UI details: the swap quote review header now stands out from the sheet background, the Activity header turns frosted
   sooner so it stays readable while you scroll, the Terms of Service icon matches the Privacy Policy one, and the Ironwood migration screens
   scroll properly on small displays.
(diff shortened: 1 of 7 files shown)


noop-sk (approved):
Reviewed the monotonic-clock fix for the app-lock timeout — SystemClock.elapsedRealtime() correctly closes the wall-clock rollback bypass, all call sites and tests are consistent. No blocking issues. (One non-blocking note: the removed LATEST_APP_BACKGROUND_TIME_MILLIS preference key is left as dead data in DataStore on upgraded installs — not a functional bug today.)

noop-sk (approved):

PR #2461, opened 2026-08-21
MOB-1691: Only recreate encrypted preferences on provable corruption

nesence-m:
Fixes [MOB-1691 — D) AreKeysPresent Misclassifies Keychain Errors](https://linear.app/zodl/issue/MOB-1691/d-arekeyspresent-misclassifies-keychain-errors) (Least Authority audit, Android counterpart).

## Problem

The Least Authority finding is about iOS `areKeysPresent` collapsing every keychain error into "keys not present". The Android analog lives in `AndroidPreferenceProvider.newEncrypted`: **any** exception while opening `co.electriccoin.zcash.encrypted` (which holds the persisted wallet secret) was classified as device-to-device migration corruption, and the recovery was destructive — the Keystore master key and the prefs file were deleted. A transient Keystore fault (daemon not ready, provider hiccup, IO error) was therefore indistinguishable from genuine corruption and could irreversibly destroy the stored seed, while the plaintext `onboarding_state` still reported the wallet as present.

## Fix

The destructive recovery now runs only when the stored data is **provably** undecryptable:

- **Orphaned-file signature (real D2D case):** the encrypted prefs file exists while a *working* Keystore definitively reports the master key absent — checked before `MasterKey.Builder` recreates the alias. A Keystore that cannot even be queried never authorizes deletion.
- **Deterministic ciphertext/keyset failure:** `BadPaddingException`/`AEADBadTagException` (authentication failure), `CharConversionException` (Tink's malformed-hex keyset signature), or Tink's shaded `InvalidProtocolBufferException` (malformed keyset proto) anywhere in the cause chain.

Every other failure is logged via `Twig` and rethrown with the data kept intact, so a later retry (the factory cache retries creation on next access) can still succeed.

## Testing

- Existing `graceful_recovery_when_encrypted_prefs_are_corrupted` (garbage keyset) still passes — that case is deterministic corruption and still self-heals.
- New `graceful_recovery_when_master_key_is_lost_in_migration` simulates the real D2D scenario (encrypted prefs written, then the master key deleted from the Keystore) and verifies recovery.
- Full `:preference-impl-android-lib:connectedDebugAndroidTest` run green on a Pixel 9 Pro XL emulator (API 35).
- The "transient failure keeps data intact" branch is not device-testable (no way to fault the Keystore daemon from instrumentation), so it is covered by the classification logic itself.

# Author

# Reviewer

### DIFF
--- a/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
+++ b/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
@@ -3,6 +3,8 @@
 package co.electriccoin.zcash.preference
 
 import android.content.Context
+import androidx.security.crypto.EncryptedSharedPreferences
+import androidx.security.crypto.MasterKey
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.filters.SmallTest
 import co.electriccoin.zcash.preference.test.fixture.StringDefaultPreferenceFixture
@@ -13,6 +15,7 @@ import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 import java.io.File
+import java.security.KeyStore
 
 // Areas that are not covered yet:
 // 1. Test observer behavior
@@ -120,9 +123,48 @@ class EncryptedPreferenceProviderTest {
             assertFalse(restoredProvider.hasKey(StringDefaultPreferenceFixture.KEY))
         }
 
+    @Test
+    @SmallTest
+    fun graceful_recovery_when_master_key_is_lost_in_migration() =
+        runBlocking {
+            val context = ApplicationProvider.getApplicationContext<Context>()
+
+            // Faithful D2D simulation: the prefs file holds keysets encrypted with a Keystore
+            // master key, and that hardware-bound key does not exist on the target device.
+            // We reproduce this by writing through EncryptedSharedPreferences directly (bypassing
+            // the factory cache) and then deleting the master key from the Keystore.
+            val masterKey =
+                MasterKey
+                    .Builder(context)
+                    .apply { setKeyScheme(MasterKey.KeyScheme.AES256_GCM) }
+                    .build()
+            EncryptedSharedPreferences
+                .create(
+                    context,
+                    D2D_FILENAME,
+                    masterKey,
+                    EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+                    EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
+                ).edit()
+                .putString(StringDefaultPreferenceFixture.KEY.key, "secret")
+                .commit()
+
+            KeyStore.getInstance("AndroidKeyStore").apply {
+                load(null)
+                deleteEntry(MasterKey.DEFAULT_MASTER_KEY_ALIAS)
+            }
+
+            // Should NOT throw — the orphaned file is provably undecryptable, so it is recreated
+            val restoredProvider = AndroidPreferenceProvider.newEncrypted(context, D2D_FILENAME)
+
+            // Prefs are empty: user will re-enter seed phrase to recover wallet
+            assertFalse(restoredProvider.hasKey(StringDefaultPreferenceFixture.KEY))
+        }
+
     companion object {
         private const val FILENAME = "encrypted_preference_test"
         private const val RECOVERY_FILENAME = "encrypted_preference_recovery_test"
+        private const val D2D_FILENAME = "encrypted_preference_d2d_test"
 
         // Internal keyset key names used by EncryptedSharedPreferences to store Tink keysets
         private const val KEY_KEYSET = "__androidx_security_crypto_encrypted_prefs_key_keyset__"
--- a/preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
+++ b/preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
@@ -322,14 +322,15 @@ private fun androidKeyStore(): KeyStore =
  * True only for failures that are deterministic for the stored ciphertext or this device's
  * Keystore: an AEAD/padding authentication failure, a Tink keyset that no longer decodes
  * ([CharConversionException] is Tink's malformed-hex signature, InvalidProtocolBufferException its
- * malformed-proto one), Tink's Keystore self-test failure ([KeyStoreException] — thrown by
- * `AndroidKeystoreKmsClient` after its own internal retry when the Keystore provably cannot
- * round-trip data, a permanent condition on devices with a buggy hardware Keystore), or
- * [InvalidKeyException], the failure [createEncryptedSharedPreferences] actually throws for a
- * genuinely device-to-device-orphaned file when [isEncryptedFileOrphaned]'s Keystore query itself
- * failed twice (see [retryOnceOrDefault]) and so could not flag the orphan first — this is the
- * second line of defense for that case. Other Keystore and general IO failures are excluded
- * because they can be transient.
+ * malformed-proto one), Tink's Keystore self-test failure ([KeyStoreException] — in tink-android
+ * 1.8.0, `AndroidKeystoreKmsClient` throws it from exactly one place, `validateAead()`, when a
+ * post-key-creation AEAD encrypt/decrypt round-trip of a random message doesn't match; a
+ * permanent condition on devices with a buggy hardware Keystore), or [InvalidKeyException], the
+ * failure [createEncryptedSharedPreferences] actually throws for a genuinely
+ * device-to-device-orphaned file when [isEncryptedFileOrphaned]'s Keystore query itself failed
+ * twice (see [retryOnceOrDefault]) and so could not flag the orphan first — this is the second
+ * line of defense for that case. Other Keystore and general IO failures are excluded because they
+ * can be transient.
  */
 internal fun isUnrecoverableCorruption(exception: Exception): Boolean =
     generateSequence<Throwable>(exception) { it.cause }
--- a/preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
+++ b/preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
@@ -25,6 +25,7 @@ import kotlinx.coroutines.withContext
 import java.io.CharConversionException
 import java.io.File
 import java.security.KeyStore
+import java.security.KeyStoreException
 import javax.crypto.BadPaddingException
 
 /**
@@ -243,8 +244,13 @@ private class AndroidPreferenceFactoryImpl : AndroidPreferenceFactory {
     /**
      * The device-to-device migration signature: the encrypted preferences file exists, but a
      * working Keystore definitively reports the master key absent, so nothing can ever decrypt
-     * the file. A Keystore that cannot even be queried yields false — an unknown Keystore state
-     * must never authorize deleting the stored secrets.
+     * the file. The query is retried once so that a momentary Keystore hiccup does not disguise
+     * a genuinely orphaned file as healthy; a Keystore that still cannot be queried yields false —
+     * an unknown Keystore state must never authorize deleting the stored secrets.
+     *
+     * Must be evaluated before attempting [createEncryptedSharedPreferences]: a failed attempt has
+     * already recreated the master key alias via [MasterKey.Builder.build], so querying the
+     * Keystore afterwards would always report the alias present and never detect the orphan.
      */
     private fun isEncryptedFileOrphaned(
         context: Context,
@@ -253,27 +259,10 @@ private class AndroidPreferenceFactoryImpl : AndroidPreferenceFactory {
         if (!encryptedPreferencesFile(context, filename).exists()) {
             return false
         }
-        return runCatching {
-            val keyStore = KeyStore.getInstance(ANDROID_KEYSTORE)
-            keyStore.load(null)
-            !keyStore.containsAlias(MasterKey.DEFAULT_MASTER_KEY_ALIAS)
-        }.getOrDefault(false)
+        return retryOnceOrDefault(false) {
+            !androidKeyStore().containsAlias(MasterKey.DEFAULT_MASTER_KEY_ALIAS)
+        }
     }
-
-    /**
-     * True only for failures that are deterministic for the stored ciphertext: an AEAD/padding
-     * authentication failure, or a Tink keyset that no longer decodes ([CharConversionException]
-     * is Tink's malformed-hex signature, InvalidProtocolBufferException its malformed-proto one).
-     * Keystore and general IO failures are excluded because they can be transient.
-     */
-    private fun isUnrecoverableCorruption(exception: Exception): Boolean =
-        generateSequence<Throwable>(exception) { it.cause }
-            .take(CAUSE_CHAIN_LIMIT)
-            .any {
-                it is BadPaddingException ||
-                    it is CharConversionException ||
-                    it.javaClass.simpleName == "InvalidProtocolBufferException"
-            }
 
     private fun createEncryptedSharedPreferences(
         context: Context,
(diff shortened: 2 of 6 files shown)


preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
```diff
@@ -201,27 +204,77 @@ private class AndroidPreferenceFactoryImpl : AndroidPreferenceFactory {
...
+        }.getOrDefault(false)
+    }
+
+    /**
+     * True only for failures that are deterministic for the stored ciphertext: an AEAD/padding
+     * authentication failure, or a Tink keyset that no longer decodes ([CharConversionException]
+     * is Tink's malformed-hex signature, InvalidProtocolBufferException its malformed-proto one).
+     * Keystore and general IO failures are excluded because they can be transient.
+     */
+    private fun isUnrecoverableCorruption(exception: Exception): Boolean =
```
noop-sk: **Missing `KeyStoreException` in the allowlist**

Tink's `AndroidKeystoreKmsClient` throws `KeyStoreException` when its documented Keystore self-test (encrypt+decrypt+compare) fails — a real, non-transient failure mode on devices with a known-buggy hardware Keystore (Tink's own Javadoc calls this out). On such a device the master key alias still exists, so `isEncryptedFileOrphaned()` returns `false`, and this exception matches none of `BadPaddingException`/`CharConversionException`/`InvalidProtocolBufferException` either. The result: `newEncrypted()` rethrows on every single app launch, permanently locking that user out of their encrypted prefs (including the wallet seed) with no recovery path short of clearing app data — worse than the catch-all this PR replaces. Please add `KeyStoreException` to this check.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
```diff
@@ -201,27 +204,77 @@ private class AndroidPreferenceFactoryImpl : AndroidPreferenceFactory {
...
+     * must never authorize deleting the stored secrets.
+     */
+    private fun isEncryptedFileOrphaned(
+        context: Context,
+        filename: String
+    ): Boolean {
+        if (!encryptedPreferencesFile(context, filename).exists()) {
+            return false
+        }
+        return runCatching {
```
noop-sk: **Transient failure here silently defeats D2D recovery**

If this Keystore query throws once (e.g. "Keystore temporarily unavailable", a state Tink explicitly works around elsewhere), `getOrDefault(false)` reports "not orphaned" even in the genuine device-to-device migration case. The actual exception `createEncryptedSharedPreferences()` then throws for a missing key is `InvalidKeyException`, which also isn't covered by `isUnrecoverableCorruption()`'s allowlist — so a real D2D-orphaned file that hits this edge just rethrows forever instead of triggering the wipe-and-recreate. Either widen `isUnrecoverableCorruption()` to also catch `InvalidKeyException`, or retry this query once before defaulting to `false`.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
```diff
@@ -201,27 +204,77 @@ private class AndroidPreferenceFactoryImpl : AndroidPreferenceFactory {
...
      */
-    @Suppress("TooGenericExceptionCaught", "SwallowedException")
+    @Suppress("TooGenericExceptionCaught")
     override suspend fun newEncrypted(context: Context, filename: String): PreferenceProvider =
         encryptedCache.getOrCreate(filename) {
             val dispatcher = Dispatchers.IO.limitedParallelism(1)
 
             val sharedPreferences =
                 withContext(dispatcher) {
+                    val isOrphaned = isEncryptedFileOrphaned(context, filename)
```
noop-sk, line 223: **Unconditional Keystore round-trip on every cold start**

`isEncryptedFileOrphaned()` is computed before the `try` block even though `isOrphaned` is only read inside the `catch` branch. On the overwhelming majority of launches (no exception), this still pays for a `File.exists()` + `KeyStore.getInstance/load(null)/containsAlias` IPC round-trip whose result is never used. Move this call inside the `catch` block so it only runs on the actual failure path.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
```diff
@@ -245,7 +298,7 @@ private class AndroidPreferenceFactoryImpl : AndroidPreferenceFactory {
         filename: String
     ) {
         runCatching {
-            val keyStore = KeyStore.getInstance("AndroidKeyStore")
+            val keyStore = KeyStore.getInstance(ANDROID_KEYSTORE)
```
noop-sk: **Duplicated Keystore-open sequence**

`isEncryptedFileOrphaned()` (line 257) and this function independently repeat the same `KeyStore.getInstance(ANDROID_KEYSTORE)` + `load(null)` sequence. Worth factoring into a small `androidKeyStore(): KeyStore` helper — the constant extraction earlier in this same PR shows this duplication almost slipped through once already.

preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
```diff
@@ -120,9 +123,48 @@ class EncryptedPreferenceProviderTest {
             assertFalse(restoredProvider.hasKey(StringDefaultPreferenceFixture.KEY))
         }
 
+    @Test
+    @SmallTest
+    fun graceful_recovery_when_master_key_is_lost_in_migration() =
```
noop-sk, line 128: **No coverage for the exception types that fall through the allowlist**

This test (and the existing corruption test) don't exercise `KeyStoreException` from a Keystore self-test failure or `InvalidKeyException` from `isEncryptedFileOrphaned()`'s own query failing — the two gaps flagged above. The removed catch-all's implicit coverage of those cases isn't replaced here, so a future change to the allowlist or to `isEncryptedFileOrphaned()`'s fallback behavior could silently reintroduce them without a failing test.

noop-sk (changes requested):
Reviewed the narrower Keystore-failure classification replacing the old catch-all wipe. The D2D-orphan detection and the general design direction are right, but the exception allowlist in `isUnrecoverableCorruption()` misses a real, non-transient failure mode, which regresses recoverability for that case. Left inline comments on the specific lines; one more note here since it's outside this diff:

`WalletRepositoryImpl.init()`'s `scope.launch { migrateDecommissionedEndpointIfNeeded() }` (and the sync-bootstrap flow feeding the SDK `Synchronizer`) sit downstream of `AndroidPreferenceProvider.newEncrypted()` with nothing catching the now-rethrown exception. On a device that hits one of the unclassified failure types below, this becomes an uncaught throw from an unsupervised startup coroutine on every process start — a crash loop for core wallet functionality where the old code would have wiped-and-recovered. Worth tracing that call path before merging, even though the fix itself belongs in this PR's files.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
noop-sk: Verified against head `af8463306f`: fixed — `it is KeyStoreException` added to the allowlist (now line 335), with a dedicated unit test (`Keystore self-test failure is unrecoverable corruption`). One thing worth a second look before merge: this is a blanket `is KeyStoreException` match rather than something narrowed to Tink's specific self-test signature — if any other, genuinely transient Keystore condition on some device surfaces as a bare `KeyStoreException` (a fairly generic type), it would now trigger the same destructive wipe-and-recreate this PR exists to prevent. Given the stakes (wallet seed), worth double-checking how narrowly Tink actually scopes this exception before considering it fully closed — but not blocking on its own.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
noop-sk: Verified against head `af8463306f`: partially addressed. Instead of widening `isUnrecoverableCorruption()` to also catch `InvalidKeyException` (one of the two options I suggested), the fix retries the Keystore query once before defaulting to `false` (`retryOnceOrDefault`, lines 262-264/342-348). That narrows the window but doesn't close it: if the query still throws on both attempts (a sustained transient outage), `isOrphaned` still silently defaults to `false`, and since `InvalidKeyException` was never added to the allowlist either, there's no second line of defense — a genuine D2D case hitting that double-failure edge still rethrows forever. Since this is a narrower version of the original hole rather than a new one, I'd call it acceptable if intentional, but wanted to flag that it isn't fully closed.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
noop-sk: Verified against head `af8463306f`: not changed — this call is still unconditional before the `try` block. The new doc comment (lines 251-253) explains why: `MasterKey.Builder(context).build()` (inside `createEncryptedSharedPreferences`) recreates the Keystore alias as a side effect on failure, so querying `containsAlias` from inside `catch` would always see the alias present and never detect a real orphan. That's a legitimate constraint — moving the call as originally suggested would break correctness — but the actual concern (a Keystore IPC round-trip on every cold start, including the near-universal success path) is still there and doesn't look mitigated otherwise (no caching, no cheaper pre-check). Worth a conscious call on whether that's an acceptable cost, rather than closing this silently.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
noop-sk: Verified against head `af8463306f`: fixed — both call sites now go through a shared `androidKeyStore(): KeyStore` helper (lines 315-318).

preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
noop-sk: Verified against head `af8463306f`: partially addressed. A new `EncryptedPreferenceRecoveryTest.kt` covers `isUnrecoverableCorruption()`/`retryOnceOrDefault()` including a `KeyStoreException` case, and a new instrumented test (`graceful_recovery_when_master_key_is_lost_in_migration`) exercises the D2D scenario end-to-end — good additions. But since `InvalidKeyException` was never added to the allowlist (see the reply on the `isEncryptedFileOrphaned` thread), there's correspondingly no test for that case either — the original gap for that specific exception type is still untested because it's still unhandled.

noop-sk (changes requested):
Re-reviewed the fix commit. Replied inline on each of the 5 original threads with verification details. Summary: 2/5 fully fixed (KeyStoreException added to the allowlist; the duplicated Keystore-open sequence factored out), 3/5 only partially or not addressed:

- The `isEncryptedFileOrphaned()` transient-failure gap is narrowed (retry-once) but not closed, and `InvalidKeyException` was never added to `isUnrecoverableCorruption()` as a second line of defense.
- The unconditional pre-`try` Keystore round-trip on every cold start is unchanged — the new doc comment explains a real correctness constraint on *why* it can't simply move into the catch block, but the performance concern itself isn't otherwise mitigated.
- Test coverage for the `InvalidKeyException` case is correspondingly still missing, since that case is still unhandled.

None of this is a step backward from where the PR started, and the KeyStoreException fix meaningfully closes the highest-severity finding (permanent lockout). But I'd like a decision on the isEncryptedFileOrphaned defense-in-depth question and an explicit call on the cold-start round-trip cost before calling this done — see the inline replies for specifics.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
nesence-m: Added `InvalidKeyException` to the `isUnrecoverableCorruption()` allowlist as the second line of defense, with a dedicated test (`Keystore-orphaned file failure is unrecoverable corruption`). Fixed in 2fc86c722.

preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
nesence-m: Added coverage: `Keystore-orphaned file failure is unrecoverable corruption` in `EncryptedPreferenceRecoveryTest.kt`, pinning `InvalidKeyException` (including nested in the cause chain) as unrecoverable corruption. Fixed in 2fc86c722.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
nesence-m: Making the call explicit: accepting the cost. It's already minimized on two axes — the file-exists gate skips the Keystore IPC entirely until an encrypted file exists, and `encryptedCache.getOrCreate` makes it once per process per file, not per access. What remains is a single `containsAlias` binder round-trip per encrypted file per cold start, on `Dispatchers.IO`, off the main thread. Given the alias-recreation constraint documented in the KDoc rules out moving it into `catch`, we're treating that as an acceptable fixed cost rather than adding invalidation-sensitive caching.

preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt
nesence-m: Checked this against tink-android 1.8.0 sources (what security-crypto-ktx 1.1.0 actually resolves here — confirmed via `:preference-impl-android-lib:dependencies`). `AndroidKeystoreKmsClient` throws `KeyStoreException` from exactly one place: `validateAead()`, when the post-key-creation AEAD self-test (encrypt then decrypt a random 10-byte message, compare) doesn't round-trip. That's the only throw site in the class — a permanent, deterministic Keystore malfunction, not a generic failure category. One correction to the KDoc's characterization though: the retry logic I described as preceding it actually lives in a different method (`hasKey()`, which retries once on `NullPointerException` when the Keystore is momentarily unavailable) and isn't in the call path that leads to this particular `KeyStoreException` — I'll fix that wording. Net: keeping the blanket `is KeyStoreException` match is justified — it's narrow by construction in this Tink version, not by our checking.

noop-sk (dismissed):
Verified all three previously-open items are genuinely resolved: InvalidKeyException is now in the isUnrecoverableCorruption allowlist with a dedicated test covering both a bare instance and one nested in a GeneralSecurityException cause chain; WalletRepositoryImpl.init() now has a real try/catch around the launch body instead of just a log alongside a rethrow, closing the crash-loop risk; and the cold-start Keystore round-trip cost is a defensible, specifically-reasoned tradeoff (file-exists gate + per-process getOrCreate caching bound the actual cost to a single off-main-thread binder call). The Tink 1.8.0 citation for the KeyStoreException scoping is specific and checkable (AndroidKeystoreKmsClient.validateAead(), with a coherent self-correction about where the retry logic actually lives), not hand-waving. Also independently traced and confirmed Milan's cross-repo note that the SDK's WalletCoordinator has the same unsupervised-scope crash class (GlobalScope + Dispatchers.Main, no SupervisorJob) — correctly deferred as a follow-up rather than shoehorned in here. No new issues in either fix commit.

noop-sk (approved):
Re-approving after the retarget/rebase onto maint/v3.10.x (correct per our branching convention — this is a bugfix to already-released behavior, so maint is the right base, not main). Confirmed the 4 substantive commits are unchanged from what I already reviewed and approved; only the unrelated drive-by isServerCustom detekt commit was dropped (correctly, since that code is main-only). No re-review of content needed.

PR #2466, opened 2026-08-24
fix(send): re-read the spendable balance before proposing and flag in-progress sync

nullcopy:
Users saw "Error creating transaction proposal: Insufficient balance (have 0, need N including fee)" right after the Send form showed a sufficient spendable balance. The form gates its button on WalletAccount.spendableShieldedBalance, which comes from Synchronizer.walletBalances; the SDK only refreshes that at processor start and after each scanned batch. After updateChainTip the Rust wallet treats every non-stabilized note in the tip shard as unspendable until the queued range is scanned, so during that window (minutes over Tor, or when far behind the tip) the cached balance is stale and proposeSend fails.

- AccountDataSource.refreshSelectedAccount() forces SdkSynchronizer .refreshAllBalances() (which publishes the new summary synchronously) and rebuilds the selected account from the fresh walletBalances value instead of waiting on the async account pipeline. A failed refresh is logged and falls back to the last known account so it never blocks sending. The AccountBalance -> WalletAccount pool mapping is extracted into one shared helper (WalletAccount.withBalances) used by both the observers and the refresh path.
- CreateProposalUseCase runs that refresh and re-checks canSpend() before proposing for both Zashi and Keystone accounts; a shortfall clears the proposal repositories and routes to the Insufficient Funds sheet exactly like the SDK-reported InsufficientFundsException path.
- The Send form shows a small hint under the balance widget while the synchronizer is not SYNCED, noting the spendable balance may still change (new send_syncingBalanceHint string, en + es).

The underlying race is being addressed separately in librustzcash and the Android SDK; this keeps the app-side failure honest in the meantime.

Claude-Session: https://claude.ai/code/session_01QFC3F7xL54sfad9qfNYnUv

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
@@ -1,8 +1,10 @@
 package co.electriccoin.zcash.ui.common.datasource
 
 import android.content.Context
+import cash.z.ecc.android.sdk.SdkSynchronizer
 import cash.z.ecc.android.sdk.Synchronizer
 import cash.z.ecc.android.sdk.model.Account
+import cash.z.ecc.android.sdk.model.AccountBalance
 import cash.z.ecc.android.sdk.model.AccountImportSetup
 import cash.z.ecc.android.sdk.model.AccountPurpose
 import cash.z.ecc.android.sdk.model.AccountUuid
@@ -14,6 +16,7 @@ import cash.z.ecc.android.sdk.model.WalletBalance
 import cash.z.ecc.android.sdk.model.Zatoshi
 import cash.z.ecc.android.sdk.model.Zip32AccountIndex
 import cash.z.ecc.sdk.extension.ZERO
+import co.electriccoin.zcash.spackle.Twig
 import co.electriccoin.zcash.ui.R
 import co.electriccoin.zcash.ui.common.model.KeystoneAccount
 import co.electriccoin.zcash.ui.common.model.SaplingInfo
@@ -65,6 +68,21 @@ interface AccountDataSource {
     suspend fun getAllAccounts(): List<WalletAccount>
 
     suspend fun getSelectedAccount(): WalletAccount
+
+    /**
+     * Forces the SDK to re-read the wallet summary from the wallet database and returns the selected
+     * account with its balances taken from that fresh read.
+     *
+     * [selectedAccount] is only as fresh as [cash.z.ecc.android.sdk.Synchronizer.walletBalances], which
+     * the SDK refreshes at processor start and after each scanned batch. Between a chain-tip update and
+     * the scan of the queued range (minutes over Tor, or when far behind the tip) the Rust wallet treats
+     * every non-stabilized note in the tip shard as unspendable, so the displayed spendable balance can be
+     * stale in either direction. Call this right before an operation that must agree with the Rust
+     * wallet's current view (proposal creation) instead of trusting the cached account.
+     *
+     * A failed refresh never blocks the caller: it is logged and the last known account is returned.
+     */
+    suspend fun refreshSelectedAccount(): WalletAccount
 
     suspend fun getZashiAccount(): ZashiAccount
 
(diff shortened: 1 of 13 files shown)


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
```diff
@@ -66,6 +69,21 @@ interface AccountDataSource {
...
+     * [selectedAccount] is only as fresh as [cash.z.ecc.android.sdk.Synchronizer.walletBalances], which
+     * the SDK refreshes at processor start and after each scanned batch. Between a chain-tip update and
+     * the scan of the queued range (minutes over Tor, or when far behind the tip) the Rust wallet treats
+     * every non-stabilized note in the tip shard as unspendable, so the displayed spendable balance can be
+     * stale in either direction. Call this right before an operation that must agree with the Rust
+     * wallet's current view (proposal creation) instead of trusting the cached account.
+     *
+     * A failed refresh never blocks the caller: it is logged and the last known account is returned.
+     */
+    suspend fun refreshSelectedAccount(): WalletAccount
```
noop-sk, line 85: New abstract method with no default implementation. Three hand-rolled `FakeAccountDataSource` test fakes implement this interface but don't override it, so they'll fail to compile (`is not abstract and does not implement abstract member`):

- `feature-migration/src/test/java/co/electriccoin/zcash/ui/common/provider/MigrationTorPreferenceAccountScopingTest.kt`
- `feature-voting/src/test/java/co/electriccoin/zcash/ui/common/repository/VotingKeystoneRepositoryTest.kt`
- `feature-voting/src/test/java/co/electriccoin/zcash/ui/common/usecase/SkipRemainingKeystoneBundlesUseCaseTest.kt`

`:feature-migration:test` and `:feature-voting:test` won't compile until an override is added to each.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/send/SendViewModel.kt
```diff
@@ -40,9 +45,25 @@ class SendViewModel(
...
+     * may still move; the form surfaces this next to the balance.
+     */
+    val isSyncing: StateFlow<Boolean> =
+        walletSnapshotDataSource
+            .observe()
+            .map { it.isSyncing() }
+            .stateIn(
+                viewModelScope,
+                SharingStarted.WhileSubscribed(ANDROID_STATE_FLOW_TIMEOUT),
+                walletSnapshotDataSource.observe().value.isSyncing()
```
noop-sk, line 64: `stateIn`'s initial value reads `.value` on `walletSnapshotDataSource.observe()` before subscribing. If this VM is the first subscriber (cold start straight into Send, or the upstream `WhileSubscribed` window already lapsed), `.value` can still be the stale/initial snapshot even though the synchronizer is genuinely mid-sync, so `isSyncing` briefly reports `false` and the hint doesn't show until the first real emission arrives. Minor, but worth double-checking that `WalletSnapshotDataSource.observe()`'s initial value is never stale for this purpose.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/CreateProposalUseCase.kt
```diff
@@ -24,7 +24,18 @@ class CreateProposalUseCase(
     suspend operator fun invoke(zecSend: ZecSend, floor: Boolean) {
         val normalized = if (floor) zecSend.copy(amount = zecSend.amount.floor()) else zecSend
         try {
-            when (accountDataSource.getSelectedAccount()) {
+            // The Send form gates its button on the cached spendable balance, which can be stale while
+            // the SDK is between a chain-tip update and the scan of the queued range. Re-read it from
+            // the SDK first so an amount the wallet can no longer cover lands on the Insufficient Funds
+            // sheet instead of surfacing the SDK's "Insufficient balance (have 0, ...)" as a raw error.
+            val account = accountDataSource.refreshSelectedAccount()
```
noop-sk, line 31: This forces a full wallet-summary refresh (JNI/DB round trip) on every proposal creation, even when the synchronizer is already `SYNCED` and the cached balance can't be stale. Worth gating this behind an is-syncing check so the common fully-synced case doesn't pay the extra round trip on the send-button's critical path.

noop-sk (changes requested):
Traced this against John's Slack investigation (#wallet-feedback, 2026-08-24) — the fix matches recommendation #1 from that analysis (re-check spendability right before proposing, route a shortfall to the Insufficient Funds sheet) and is an honest mitigation rather than a fix for the underlying tip-shard scan race (which is tracked separately for the SDK/librustzcash side). Left three inline notes; the AccountDataSource one is a compile blocker that needs fixing before this can land.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
```diff
@@ -164,6 +182,29 @@ class AccountDataSourceImpl(
 
     override suspend fun getZashiAccount() = withContext(Dispatchers.IO) { zashiAccount.filterNotNull().first() }
 
+    @Suppress("TooGenericExceptionCaught")
+    override suspend fun refreshSelectedAccount(): WalletAccount =
+        withContext(Dispatchers.IO) {
+            val account = getSelectedAccount()
+            val synchronizer = synchronizerProvider.getSynchronizer()
+            if (synchronizer !is SdkSynchronizer) {
```
noop-sk, line 190: This is very likely dead code on the build the app actually ships. The pinned SDK commit (`SDK_COMMIT_PIN=26dab43d...`) has `IS_SLIPSTREAM_ENABLED=true` in its `gradle.properties`, which makes `sdk-incubator-lib` compile `EngineSynchronizerFactory` from `src/engineSlipstream/java` — so `Synchronizer.new()` returns a `SlipstreamSynchronizer`, not an `SdkSynchronizer`. `SlipstreamSynchronizer` is a sibling implementation of `CloseableSynchronizer` (`SdkSynchronizer`'s constructor is even `private`, so it can't be a supertype here). That means `synchronizer !is SdkSynchronizer` is `true` on the engine we actually build with, so this whole method just falls through to `return@withContext account` and `refreshAllBalances()` never runs — the fix this PR is meant to deliver doesn't fire at runtime.

Worth noting: the underlying UI-shows-spendable-but-propose-rejects race this PR targets was diagnosed against `CompactBlockProcessor`/`SdkSynchronizer` (the classic engine), but the app is already on the Slipstream engine, which has its own reactivity/masking mechanism for the same race — see `EngineMapping.kt`'s `maskIfStale`/`effectiveTipFresh` (while the tip is unverified, `spendableValue` is masked to 0 and folded into `valuePendingSpendability` at the engine level, refreshed every ~2s poll tick). It's worth first confirming whether the reported symptom still reproduces at all against the Slipstream engine before extending this fix to cover it — the `SdkSynchronizer` branch here may be dead code for the wrong reason (targeting an engine no longer in use), and this PR as written provides no coverage for the Slipstream path.

noop-sk (commented):
Follow-up after cross-checking against the SDK repo directly — this changes the severity assessment of the whole fix, not just a nit.

PR #2471, opened 2026-08-25
Release 3.10.1

### LINKED ISSUE #2470: Release 3.10.1
## Release Checklist

- [x] Version bump
- [x] Changelog update
- [x] Fastlane files
- [ ] PR merged
- [ ] Build deployed to Google Play

noop-sk:
Closes #2470

### Added:

- We added a home screen banner and follow-up screen for a small amount of ZEC that can be left in Orchard after migrating to Ironwood, so you can choose whether to lock it or move it.

### Fixed:

- We fixed migrating with a Keystone wallet, so migrations are now split into signing rounds correctly.

### Fixed:

- A temporary Android Keystore failure at startup is no longer mistaken for corrupted encrypted storage: the encrypted secret store is now
  recreated only when its data is provably undecryptable (for example after a device-to-device transfer, which cannot move hardware-bound
  keys), and any other failure keeps the stored data intact.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -210,6 +210,9 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
         with:
           fetch-depth: 0 # to fetch all commits
+      - name: Set up companion SDK build (if SDK_COMMIT_PIN is set)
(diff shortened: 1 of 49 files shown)


nesence-m (approved):

zcash-harry (commented):
tACK

PR #2472, opened 2026-08-26
MOB-1805: Add Coinholder Polling prompt to the wallet status widget

noop-sk:
## Summary
- Adds a Coinholder Polling (CHP) nudge to the Home message ("wallet status widget") banner, per [MOB-1805](https://linear.app/zodl/issue/MOB-1805/implement-a-coinholder-polling-prompt-in-the-wallet-status-widget).
- Ranked below Wallet Backup and above the other opt-in nudges (Tor, currency conversion, crash report); still outranked by the "always shown" states (Disconnected/Syncing/Restoring/Shielding).
- No dismiss affordance by design (matches Figma) — visibility is fully data-driven via a new `VotingHomeMessageSource` contract: voting enabled, an active session that hasn't passed its end time, and the account hasn't submitted votes for every proposal yet (with a durable-recovery fallback so it doesn't reappear for a user who already voted after an app restart).
- Tapping "Start" reuses the existing first-time-vs-returning-user navigation branch (extracted from `MoreVM` into a shared `NavigateToVotingUseCase`) for both Zodl and Keystone wallets.
- Icon and horizontal/vertical padding matched exactly against the Figma spec (`check-verified-02` outline icon, `24dp`/`14dp` padding) after a first pass shipped with the wrong icon/padding.

Branched from `main`, rebased onto `maint/v3.10.x` at the user's direction (this line is where the team is actively pushing coinholder-voting fixes right now).

Screenshot 2026-08-26 at 11 08 10

## Test plan
- [x] `./gradlew :zodl-android:ktlint` — clean
- [x] `./gradlew :zodl-android:detektAll` — clean (one pre-existing unrelated finding in `AutomaticServerRepository.kt` was fixed as a drive-by on `main`, but doesn't exist on this `maint/v3.10.x`-based branch since that function isn't present here — confirmed via the rebase conflict)
- [x] `compileZcashmainnetInternalDebugKotlin` for `ui-lib` and `feature-voting` — BUILD SUCCESSFUL
- [x] Manual verification on the Z-Droid emulator (mainnet account, live NU7 poll): banner appears with correct copy/icon/padding matching Figma, "Start" navigates into the voting flow
- [ ] No automated unit tests added yet for `VotingHomeMessageSourceImpl`/`NavigateToVotingUseCase` (flagged in review, not blocking)


nesence-m (commented):
Went through the whole home-message pipeline and voting wiring, not just the hunks. The core is solid — no blocking issues found — but two convention gaps are worth fixing before merge, and CI (build, unit tests, emulator + E2E jobs) was still running at review time.

**Worth fixing before merge:**

1. **No `CHANGELOG.md` entry.** This is a user-facing feature shipping in a 3.10.x point release, the `[Unreleased]` section is live, and the 3.10.0 section even carries the sibling "We brought back Coinholder Polling" entry — this one should join it per repo convention.

2. `ui-design-lib/src/main/res/ui/common/values/strings.xml:446` — **`home_message_coinholder_polling_subtitle` has no `values-es/` counterpart.** It's `tools:ignore="MissingTranslation"` and consistent with the voting feature's status quo (most `coinVote_*` keys are untranslated), but the concrete symptom is a mixed-language banner in the es locale: English title/subtitle with a Spanish "Iniciar" button, because the one reused string that IS translated is the CTA.

**Nits / hardening (non-blocking):**

3. `CoinholderPollingMessage.kt:70` — the CTA reuses `smartBanner_content_backup_button`; if backup copy ever changes, this button silently changes with it. A dedicated `home_message_coinholder_polling_button` key would decouple it and give the es translation a home.

4. `FeatureVotingImpls.kt:213` — round-end expiry is only evaluated on upstream emissions, so a round ending while the app sits on Home keeps the prompt until the next snapshot/account emission. Bounded in practice (refresh on app open; the polling screen handles closed rounds) — just noting.

5. `FeatureVotingImpls.kt:207,223` — `sessionsByRoundId.values.firstOrNull { ACTIVE }` is map-order-arbitrary if two rounds were ever simultaneously ACTIVE, and the id is re-derived via `toHex()` when the map key was already in hand. Works today (lowercase invariant matches `SubmitVotesUseCase`'s keying), but iterating `entries` and carrying the key would remove the casing assumption — note `VotingSessionScope` lookups are case-sensitive while `SubmitVotesUseCase` itself only trusts `ignoreCase = true` matches.

6. `FeatureVotingImpls.kt:228-236` — the persisted hide-predicate mirrors `VoteCoinholderPollingVM.refreshRecoveryVoteCounts` line-for-line (verified), but it hides only when `submittedCount == proposals.size`, while the VM treats any non-null count as VOTED. The two agree for any vote submitted through this app (the submit path requires drafts on every proposal), but a legacy/incomplete recovery snapshot with `submittedAt` set and fewer `proposalSelections` would show VOTED on the polling screen while Home keeps nagging until round end. Gating on `submittedAtEpochSeconds != null` alone would give exact parity.

7. No unit tests (acknowledged in the PR body) — `VotingHomeMessageSourceImpl` and `NavigateToVotingUseCase` are cleanly testable and `feature-voting/src/test/` already has the harness; the finding-6 predicate is exactly what a table-driven test would pin down.

**Verified clean, for the record:** no warm-up flash (recovery repo emits the real persisted value first; empty snapshot defaults to hidden); no permanent-suppression risk (visibility fully data-driven, cache in-memory, priority renumber consistent with `createMessage`'s when-order and never persisted); DI complete with the Koin smoke test covering it; navigation entirely through `NavigationRouter` and the `MoreVM` extraction is behavior-preserving; the drawable is tint-driven and dark-mode safe; `VOTING_ENABLED` kill switch honored; nothing main-only leaking into the maint base.

nesence-m (commented):
Re-verified the fix round (`d21f2bb6e`) against each original finding — real progress, but one item regressed, so not approving yet.

**Resolved:**
- **CHANGELOG** — fixed; `[Unreleased]/### Added:` entry in house style.
- **`firstOrNull` + `toHex()` re-derivation** — fixed exactly as suggested: `ActiveScope` now carries `entry.key` as the roundId, `entries` iteration, casing assumption gone with the import. (Selection between two simultaneously-ACTIVE rounds is still arbitrary, but the lookup key is now self-consistent, which was the substantive risk.)
- **Hide-predicate parity** — fixed; persisted gate is `submittedAtEpochSeconds != null`, and the in-memory OR-branch is the session-local analog of the same semantics.

**Remaining — and now slightly worse:**
- **The es strings.** `home_message_coinholder_polling_subtitle` still has no `values-es/` value, and the new dedicated `home_message_coinholder_polling_button` key (good idea in itself) shipped without one either — which is a **regression**: the old borrowed `smartBanner_content_backup_button` *did* have an es translation ("Iniciar", `values-es/strings.xml:358`), so on es devices the CTA flips from Spanish to English and the banner is now fully English. Both keys' `tools:ignore="MissingTranslation"` suppressed the lint that would have caught it. Trivial fix: add both es strings and drop the ignores.

**Silently skipped (no rationale anywhere — an explicit position either way is fine):**
- Round-end expiry still only evaluated on upstream emissions (`FeatureVotingImpls.kt`, unchanged).
- No unit tests for `VotingHomeMessageSourceImpl` / `NavigateToVotingUseCase`.

No other new defects in the delta — the `entries` rewrite is internally consistent and `distinctUntilChanged` over the string-carrying `ActiveScope` behaves the same or better. CI at `d21f2bb6e` was still running at verification time (static gates green, build/tests/lint in progress; nothing failed).

Happy to approve once the two es strings land and CI is green.

noop-sk (commented):
Approving. Confirming the es-translation point explicitly: no `values-es` entry for the new `home_message_coinholder_polling_button` (or `home_message_coinholder_polling_subtitle`) is intentional — Coinholder Polling stays English-only for now, matching the feature's existing `tools:ignore="MissingTranslation"` status quo. Not a regression to fix, a deliberate product decision.

CI note: `test_android_modules_emulator (app)` shows `FAILURE`, but it's a confirmed pre-existing flake — `co.electroniccoin.zcash.ui.screenshot.ScreenshotTest.takeScreenshotsForNewWalletAndRestOfAppLightArXB` failed with `AdbCommandRejectedException: device offline` (ADB losing the emulator mid-run), the same failure was already present on the prior commit (`faf28fc3f`) before this PR's fix commit landed, and it's unrelated to any file this PR touches. Everything else is green; `e2e-android (phase 2)` and `test_robo_release` were still in progress at review time.

Still open, non-blocking per earlier review: the `FeatureVotingImpls.kt:213` round-end-expiry-on-emissions note and missing unit tests — neither gates merge.

`https://github.com/zodl-inc/zodl-android/pull/2472`

nesence-m (approved):
Verified the es-translation position rather than taking it on faith, and it holds: at the head, EN has **156** `coinVote_*` keys while ES has **3**, all generic boilerplate (`coinVote_error_title`, `coinVote_common_retry`, `coinVote_delegationSigning_scanInstructions`) — the entire voting flow behind this banner is English on an es device, and the `tools:ignore="MissingTranslation"` markers make this a deliberate, lint-visible feature-level i18n policy rather than an oversight. On those facts the earlier "Iniciar" CTA was the inconsistency — one Spanish word fronting an all-English feature — so the regression finding is withdrawn as accepted feature policy. If/when CHP gets localized, both new keys are cleanly placed for it.

Everything else stands verified from the fix round: CHANGELOG entry, the `entries`/roundId rewrite, and the hide-predicate parity are in at `d21f2bb6e`, and no correctness or security finding remains open. The two skipped nits (round-end expiry only on upstream emissions; unit tests) are cosmetic/nice-to-have — the banner can linger briefly past round end but taps into a flow that handles an ended round.

CI note, for honesty: at review time `test_android_modules_emulator (app)` had failed and e2e phase 2 / robo were still running. That job is not in the branch ruleset's required checks and has failed on 7 of the last 8 Pull Request runs repo-wide across unrelated PRs — treating it as the known chronic breakage. If its logs turn out to show a CHP test once the run completes, ping me and I'll take another look.

[#2472 MOB-1805: Add Coinholder Polling prompt to the wallet status widget](https://github.com/zodl-inc/zodl-android/pull/2472)

PR #2473, opened 2026-08-26
MOB-1806/MOB-1809: Static voting config v2 with config mirror failover

nesence-m:
## Summary

Adopts Valar's static voting config **version 2** with full **mirror failover** on both configuration legs — the defense against ISPs blocking `voting.valargroup.dev` ([context thread](https://zodl.slack.com/archives/C0A8RHV7KEW/p1787694873005729?thread_ts=1787671580.949409&channel=C0A8RHV7KEW&message_ts=1787694873.005729)). Android counterpart of iOS [zodl-inc/zodl-ios#2012](https://github.com/zodl-inc/zodl-ios/pull/2012); implements [MOB-1806](https://linear.app/zodl/issue/MOB-1806/need-to-add-fallback-to-improve-server-side-issues) / [MOB-1809](https://linear.app/zodl/issue/MOB-1809/adopt-static-voting-config-v2-with-config-mirror-failover-android) (the hand-off carrying the authoritative spec — the `valargroup/token-holder-voting-config` README). Targets `maint/v3.10.x` (fix for the live NU7 CHP round; forward to `main` via the usual cascade). Three commits: v2 schema support → alignment with the iOS mirror-failover semantics → retry/timeout ownership + test hardening.

**Model** (`StaticVotingConfig`)
- Accepts `static_config_version` 1 **and** 2; the field is now required. v2 introduces the `dynamic_config_urls` array; v1's singular `dynamic_config_url` keeps working (users saved v1 custom sources during the NU7 incident). Unknown versions fail with a precise `Unsupported static_config_version N`, not a decode error.
- Every dynamic-config URL must be HTTPS (a static-config rotation can't downgrade voting to plaintext).
- The bundled trust anchor is now a **mirror list**: the canonical gateway pin plus a byte-identical `raw.githubusercontent.com` pin (same checksum `28fc9b63…` in path and query) — trust is carried by the checksum, not the origin.

**Static anchor walk** — deliberately wide: falls through to the next mirror on transport failure, any non-200, or checksum mismatch (each mirror is independently hash-gated, so fall-through is availability-only); a decode/validation failure *after* the hash matched is authoritative for every mirror and stops immediately. All-fail reports the canonical origin's error.

**Dynamic config walk** — README-normative (narrow): falls through only on transport failure or HTTP 5xx; 4xx and decode/validation failures are authoritative. First error reported; serving origin logged.

**Hardening**
- Config requests own their retry policy: they're exempted from the direct client's `HttpRequestRetry` (per-request `retry { noRetry() }`), so a dead mirror fails once in ≤15 s and the walk moves on — instead of the plugin internally retrying it ~5× (~90 s) first.
- 15 s per-attempt timeout on both config legs — Ktor-level where available, with an app-side `withTimeout` fallback on the Tor path (which previously had no bound at all).
- `zodl_cache_bust=<token>` query param on requests to exactly `raw.githubusercontent.com` (its CDN caches branch paths ~300 s and ignores request cache headers — matters during round rollover); the clean URL is what's logged and used in error messages (the token never leaks). The static-leg mirror is content-addressed and deliberately not busted.
- A custom source byte-equal (URL + checksum) to a bundled mirror uses the full bundled walk; settings' "Default" detection now requires URL **and** checksum to match a bundled mirror, so a same-URL/different-checksum custom source stays custom (fixes a pre-existing silent-collapse quirk).

No stored-data migration needed; v1 custom overrides keep working unchanged.

## Test plan

- Model: v1/v2 decode, singular/plural cross-rejection, missing version, non-HTTPS entries, precise unsupported-version message (wins over field errors), bundled-mirror invariants (both pins parse, same 64-hex checksum in path + query, checksum stripped from the fetched URL).
- Static walk: first-healthy-mirror short-circuit; fall-through on transport/non-200/hash-mismatch; authoritative stop on hash-matched-but-malformed bytes; first-error retention with distinguishable payloads; ordered requests over the real bundled two-URL list; empty source list.
- Dynamic walk: order respected; transport + 5xx fall through; **404 does not** fall through; decode/validate failures do not fall through; first-error retention; cache-bust applied only to exact-host `raw.githubusercontent.com` (subdomain/path near-misses stay clean), same token within a walk, fresh token across walks.
- Timeouts & retry: 15 s timeout capability asserted on config requests and absent from vote-server requests; app-side fallback converts a stalling attempt into mirror fall-through; with a real `HttpRequestRetry` plugin installed, a config request makes exactly one attempt while vote-server requests still retry.
- Memoization (same source → one fetch; source change / `invalidateConfigCache` → refetch), cancellation propagation, and `VoteChainConfigVM` default-collapse/duplicate detection.
- `./gradlew :feature-voting:test` — BUILD SUCCESSFUL, all variants (implementation + independent verification passes). `./gradlew :app:assembleZcashmainnetStoreDebug` — BUILD SUCCESSFUL. ktlint and `detektAll` fully clean on the `maint/v3.10.x` base.
- Reviewed by a multi-critic adversarial pass (spec compliance, Ktor correctness, test coverage, regressions); all confirmed findings addressed in the third commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 - Bottom sheets across the app now share one background color.
 - We rebuilt polling to work with your Ironwood-migrated funds.
 - We made poll loading more reliable during service outages.
+- Coinholder Polling now supports multiple dynamic configuration URLs and automatically falls back to the next one if the first is
+  unreachable, adding resilience against an ISP blocking the primary voting configuration host.
 
 ### Fixed:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,9 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 - Bottom sheets across the app now share one background color.
 - We rebuilt polling to work with your Ironwood-migrated funds.
 - We made poll loading more reliable during service outages.
-- Coinholder Polling now supports multiple dynamic configuration URLs and automatically falls back to the next one if the first is
-  unreachable, adding resilience against an ISP blocking the primary voting configuration host.
+- Coinholder Polling no longer becomes unavailable when the primary voting configuration service is unreachable or blocked: the app
+  verifies the same pinned configuration from a second independent mirror, walks the configuration's own mirror list, and config
+  requests now give up after 15 seconds instead of two minutes.
 
 ### Fixed:
 
--- a/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/model/voting/StaticVotingConfig.kt
+++ b/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/model/voting/StaticVotingConfig.kt
@@ -12,7 +12,7 @@ import java.util.Base64
 @Serializable
 data class StaticVotingConfig(
     @SerialName("static_config_version")
-    val staticConfigVersion: Int = STATIC_CONFIG_VERSION_V1,
+    val staticConfigVersion: Int,
     @SerialName("dynamic_config_url")
     val dynamicConfigURL: String? = null,
     @SerialName("dynamic_config_urls")
@@ -41,6 +41,7 @@ data class StaticVotingConfig(
                 if (dynamicConfigURL.isNullOrBlank()) {
                     throw VotingConfigException("dynamic_config_url must not be blank")
                 }
+                requireHttpsScheme(dynamicConfigURL, "dynamic_config_url")
             }
 
             STATIC_CONFIG_VERSION_V2 -> {
@@ -50,6 +51,7 @@ data class StaticVotingConfig(
                 if (dynamicConfigURLs.any(String::isBlank)) {
                     throw VotingConfigException("dynamic_config_urls must not contain blank entries")
                 }
+                dynamicConfigURLs.forEach { url -> requireHttpsScheme(url, "dynamic_config_urls") }
             }
         }
         if (trustedKeys.isEmpty()) {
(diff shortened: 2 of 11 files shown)


noop-sk (approved):
Reviewed with heightened supply-chain scrutiny, given a prior incident in this codebase involving an unpinned `BUNDLED_PINNED_SOURCE` repoint. Verdict: this is a fundamentally different (and safe) pattern.

**Mirror trust model:** both `BUNDLED_PINNED_SOURCE` (`voting.valargroup.dev`) and the new `BUNDLED_PINNED_SOURCE_MIRROR` (`raw.githubusercontent.com/valargroup/token-holder-voting-config`) carry the **identical** SHA-256 checksum, hardcoded in the app source (both in the path and the `?checksum=` param). `StaticVotingConfig.decodeAndVerify()` computes SHA-256 over the fetched bytes and throws `StaticVotingConfigHashMismatchException` on any mismatch — this is treated as retryable-to-next-mirror, but a hash-matched-yet-malformed payload is authoritative (non-retryable). This is a content-addressed, SRI-equivalent pattern: a compromised or malicious mirror cannot inject arbitrary content — it can only serve bytes matching the pinned hash (i.e. identical to the canonical source) or fail closed. Confirmed no `NEVER COMMIT` markers or unpinned trust-anchor override anywhere in the diff.

One layer down: the dynamic-config legs pulled from inside the verified static config aren't hash-pinned themselves, but the safety-critical `rounds` entries go through `retainingRoundsWithValidSignatures()` (Ed25519, keys sourced from the already-verified static config) — pre-existing, unmodified by this PR.

Timeout/retry correctness verified: 15s timeout applied via both Ktor per-request `timeout {}` and an app-side `withTimeout` fallback for the Tor client; `TimeoutCancellationException` is caught before the generic cancellation rethrow at every call site (avoids the classic trap); config requests use per-request `retry { noRetry() }` rather than disabling retry client-wide.

Test coverage is notably thorough for a security-sensitive change, including explicit negative tests for host-matching precision (`evil.raw.githubusercontent.com`, lookalike paths) proving exact-equality host checks rather than `contains`/`endsWith`.

No blocking correctness issues in the rest of the diff (mostly test additions — 2196 of 2768 lines).

`https://github.com/zodl-inc/zodl-android/pull/2473`

PR #2474, opened 2026-08-26
Fix ambiguous voting submission recovery

p0mvn:
## Summary
- retain incomplete voting round state instead of clearing recovery material
- reconcile spent-nullifier retries when the returned transaction hash confirms successfully
- apply the same idempotent recovery to delegation and vote-commitment submissions

## Motivation / Why
A voting transaction can be accepted while its HTTP response is lost, especially over a slow or interrupted Tor connection. A retry then reports that the nullifier is already spent even though the original transaction succeeded. Clearing an apparently incomplete round at that point can delete the private randomness needed to recover or continue the vote.

This change fails closed when existing round state is incomplete, preserving its database records for recovery. When a spent-nullifier response includes a transaction hash, it performs a bounded status lookup and treats a confirmed transaction as the original successful submission.

## Tests
- `:feature-voting:testZcashmainnetStoreDebugUnitTest` focused on `PrepareVotingRoundUseCaseTest`, `SubmitVotesUseCaseProgressTest`, and `KtorVotingApiProviderTest`
- verified against the published Android wallet SDK with no included-build replacement

## Known limitation
Spent-nullifier recovery still requires the server response to include a usable transaction hash. Lookup by governance nullifier or VAN commitment is outside this change.

# Author
- [ ] **Self-review** the code in GitHub's web interface
- [ ] Update manual tests as appropriate
- [ ] Check the code coverage report
- [ ] Update documentation as appropriate
- [ ] Run the app and try the changes


p0mvn:
The lints are further addressed in:
- https://github.com/zodl-inc/zodl-android/pull/2479
- https://github.com/zodl-inc/zodl-android/pull/2480

So, I am keeping this PR as is to avoid conflicts

PR #2475, opened 2026-08-26
Recover Base64-decoded voting leaf indexes

p0mvn:
## Summary
- recover delegation leaf indexes that compatibility clients mistakenly Base64-decode
- preserve strict failure handling for genuinely malformed event values
- add regression coverage for the observed `3753` → `߾w` transformation

## Motivation / Why
Mainnet voting submissions can fail after delegation confirmation when a numeric CometBFT event attribute is opportunistically Base64-decoded into non-ASCII text. Re-encoding that value restores the original decimal tree position and lets recovery continue.

## Tests
- `:feature-voting:testZcashmainnetFossDebugUnitTest --tests 'co.electriccoin.zcash.ui.common.usecase.SubmitVotesUseCaseProgressTest'`
- `:app:assembleZcashmainnetFossDebug`
- installed and launched the updated APK on a Pixel 10a

## Author checklist
- [ ] Self-review in GitHub
- [x] Added automated regression coverage
- [x] Ran the app on a physical device
- [ ] Checked code coverage report


p0mvn:
The lints are further addressed in:
- https://github.com/zodl-inc/zodl-android/pull/2479
- https://github.com/zodl-inc/zodl-android/pull/2480

So, I am keeping this PR as is to avoid conflicts

PR #2479, opened 2026-08-27
Voting submission recovery: review fixes

nuttycom:
Review fixes for #2474, one commit per confirmed finding from the review of that branch. Builds on `fix/voting-submission-recovery` and carries its commits; the last eight commits are the fixes.

- **Treat omission of a locked proposal as no conflict** — the unsubmitted branch of the conflict guard lacked the null check the submitted branch had, so a retry omitting a locked-but-unsubmitted proposal always raised `ConflictingProposalSelection`.
- **Lock proposal selections at first durable choice binding** — locking moves from whole-request/pre-broadcast to per-proposal, immediately before `buildVoteCommitment`; failures before that point no longer freeze never-broadcast choices for the round's lifetime.
- **Make the repository own the selection-lock invariant** — `storeProposalSelections` throws the typed error and every path that binds or reuses a choice (fresh bundle, cached-transaction reuse, proposal completion) revalidates through it; the divergent use-case duplicate check is gone. Includes a real-implementation repository test.
- **Report incomplete recovered-vote verification as retryable** — commitment-tree scan failures surface as the new typed `RecoveredVoteVerificationUnavailable` instead of raw `IllegalState/IllegalArgumentException`, and are no longer conflated with a verified mismatch.
- **Map typed voting errors to dedicated user messages** — `VotingErrorMapper` matches on the `VotingErrors` subtype (exhaustive `when`), with new strings for the conflict, mismatch, verification-unavailable, and round-not-prepared cases; substring matching remains only for foreign exceptions.
- **Resolve both recovered-vote leaves in one tree scan** — verification does a single paginated dual-target scan instead of two sequential full scans.
- **Decode commitment-tree responses via content negotiation** — the two new endpoints use `body<T>()` through the shared ContentNegotiation plugin like every sibling; the module-private `Json` instance is deleted.
- **Restore the detekt gate to green** — the 11 unbaselined issues are resolved structurally (function splits, scanner extracted to `CommitmentTreeScan.kt`); the baseline only loses the stale `isRoundRecoverable` entry and gains `LargeClass` for the progress test suite.

The review's plausible-but-unconfirmed findings (spent-nullifier poll budget, reconciler log-before-code ordering, missing `noCache()` on the tree GETs, non-nullable `tree` DTO field, unchecked recovered leaf index, early `singleShareMode` pinning, `FAIL_CLOSED` hard block) are intentionally not addressed here.


p0mvn (approved):
LGTM.

There is one edge case regression that I followed up with here: https://github.com/zodl-inc/zodl-android/pull/2480

Also, retested happy path manually.

PR #2481, opened 2026-08-27
Consolidate voting-submission-recovery hotfix + fix cached-delegation leaf-index gap

noop-sk:
## Summary

Consolidates the in-flight 3.10.2 voting hotfix stack (#2474, #2475, #2479, #2480 - all by @p0mvn, plus #2479 by @nuttycom) onto one branch and adds one more fix on top: the Base64-leaf-index recovery landed in #2475 only covered `delegateVoteVanPosition`, but the identical corruptible `leaf_index` attribute is also read by `probeCachedDelegationVanPosition` (the resumed-round fast-path probe), which still used a plain `toIntOrNull()`.

**Effect of the gap:** a resumed round with an already-confirmed but Base64-mangled cached delegation would be misread as unconfirmed, paying for a full rebuild-and-resubmit cycle instead of recovering directly - for exactly the population of users this hotfix exists to help, in a flow already flagged as slow.

## What's new here on top of the existing stack

- Extracted the recovery logic into a shared `String.recoverLeafIndexOrNull()` and used it at both call sites instead of just `delegateVoteVanPosition`.
- Documented and regression-tested the safety assumption behind `castVoteLeafPositions`' comma-delimited format not needing the same recovery (a comma is outside the Base64 alphabet, so an upstream opportunistic decode of that attribute is expected to fail closed rather than corrupt it).
- New tests: `recoverLeafIndexOrNull` edge cases (plain decimal, whitespace, Base64-recovery, non-recoverable non-ASCII, blank), a `castVoteLeafPositions` guard test, and an integration test (`resumedRoundRecoversCorruptedCachedDelegationLeafIndexWithoutResubmitting`) exercising the exact `probeCachedDelegationVanPosition` gap via the existing `RecoveryFixture`.

## Verification

- Red-green checked against both the new test and Roman's original `delegateVoteVanPositionRecoversLegacyBase64DecodedLeafIndex` test: reverting either call site back to plain `toIntOrNull()`, or breaking the shared `recoverLeafIndexOrNull()` itself, fails the relevant test(s); restoring the fix passes them again. Breaking the shared function fails **both** the original and new tests simultaneously, confirming they now share one fix point rather than two independent ones.
- ktlint + detekt clean.
- Full `feature-voting` unit test suite green.

## Supersedes

This branch contains everything from #2474 and #2475 (which already contain #2479 and #2480) plus the fix above. Recommend merging this instead of the individual PRs; will comment on those to point here.

## Follow-up needed before merge

@p0mvn - could you do a mainnet retest of the resumed-round/cached-delegation path specifically, the way you did for the other four recovery scenarios? The new coverage here is unit-test-only; you own the manual mainnet verification step for this fix family.


noop-sk (commented):
Verified the claimed `probeCachedDelegationVanPosition` leaf_index/Base64-recovery fix directly against the diff, and checked whether the #2474/#2475/#2479/#2480 consolidation lost any prior findings.

**Fix confirmed real (not just trusting the description):**
- `probeCachedDelegationVanPosition()` now ends with `?.attribute("leaf_index")?.recoverLeafIndexOrNull()` (was `?.toIntOrNull()`); `delegateVoteVanPosition()` refactored onto the same shared `String.recoverLeafIndexOrNull()` — plain `toIntOrNull()` first, falls back to Base64-re-encode-and-retry on any non-ASCII byte. Both functions now genuinely share the recovery path.
- Dedicated test: `resumedRoundRecoversCorruptedCachedDelegationLeafIndexWithoutResubmitting` (`SubmitVotesUseCaseProgressTest.kt`) — corrupted Base64-mangled `leaf_index`, asserts correct recovery with zero resubmissions. This exercises the probe function's specific gap, not just the sibling's existing coverage.
- New commit is tightly scoped: only `SubmitVotesUseCase.kt` (+28/-16) + tests.

**Consolidation check:** #2474/#2475 have no formal review threads to lose (only issue-comments — Roman's manual-testnet log, a design-rationale note, both consistent with this diff). #2479/#2480's one review (approved, with a follow-up note that became #2480) — nothing blocking left unaddressed. No leftover debug/revert scaffolding from the branch's interleaved experimental commits.

**Rest of diff:** no blocking issues. `castVoteLeafPositions()` intentionally does NOT get the same recovery treatment — reasoning (comma outside Base64 alphabet, so this corruption class can't occur there) is documented inline and backed by its own guard test. `VotingErrorMapper.toUserFriendlyMessage()` exhaustively matches the sealed `VotingErrors` type.

**CI:** all completed checks green (`test_kotlin_modules` incl. the new tests, ktlint, detekt, lint, gradle wrapper, secrets, properties, keystone-simulator build). Emulator tests + main build jobs were still in progress at review time — nothing red so far.

Base branch `maint/v3.10.x` correct per convention.

**Open item, out of code-review scope:** Roman's mainnet retest of the resumed-round/cached-delegation path is still outstanding per the PR body — a human/QA gate, not a code defect, shouldn't block this approval.

(Posting as COMMENT, not APPROVE — GitHub blocks self-approval and this PR is authored by our own account.)

`https://github.com/zodl-inc/zodl-android/pull/2481`

nesence-m (approved):

nesence-m (approved):
Reviewed at head 58a5fd9 — verified the claims independently rather than taking the description on faith, and they all check out.

**Consolidation is provably lossless.** `compare fix/voting-leaf-index-base64...fix/voting-submission-recovery-consolidated` shows ahead-by 1, behind-by 0 — the tip of Roman's stack (#2474 + #2475, containing the #2479/#2480 review-fix merges) is literally an ancestor of this branch, plus exactly one new commit. All five `[TO REVERT]` test-injection commits are paired with their reverts. Nothing was re-cherry-picked, so nothing could be dropped.

**The probe fix is contextually correct, not a mechanical mirror.** `probeCachedDelegationVanPosition` reads the same `delegate_vote` event's `leaf_index` attribute from the same `fetchTxConfirmation` path with the same corruption source as `delegateVoteVanPosition`, and the shared `recoverLeafIndexOrNull` helper preserves each caller's contract — the probe keeps null → re-run delegation, the delegate path keeps its throwing contract. Checked the recovery math too: a base64-mangled decimal string always trips the non-ASCII guard (digit chars map to 6-bit values ≥ 52, so the first decoded byte is ≥ 0xD0), and non-recoverable corruption fails closed via `toIntOrNull` — no silent-wrong-value path. The deliberate non-recovery in `castVoteLeafPositions` is sound (comma isn't in the Base64 alphabet) and guard-tested; these three are the only `leaf_index` readers in the repo.

**Red-green claim confirmed.** `resumedRoundRecoversCorruptedCachedDelegationLeafIndexWithoutResubmitting` traced against the fixture: without the fix, bundle 0 falls through to `submitDelegation` and both assertions fail (stored VAN position would be 1, not 3753; submission count 1, not 0). Roman's original test is retained unchanged.

Non-blocking notes:
- **CHANGELOG**: no `[Unreleased]` entry here or in #2474 for a user-visible recovery fix shipping in 3.10.2 — worth adding before the cut if release prep doesn't cover it.
- **CI**: `test_android_modules_emulator (app)` and `e2e-android (phase 1)` are red with infra signatures (device offline / adb daemon unreachable; the emulator job fails identically on source PR #2474, and this diff touches no screenshot code) — re-run rather than chase, but I didn't independently confirm the e2e one is pure flake.
- Nit: `recoverLeafIndexOrNull`'s KDoc says "can't be plain ASCII decimal" while the code checks any non-ASCII char (`> 0x7F`) — cosmetic wording.

Approval covers the code; Roman's mainnet retest of the resumed-round/cached-delegation path remains the QA gate before merge, as noted in the PR.

[#2481 Consolidate voting-submission-recovery hotfix + fix cached-delegation leaf-index gap](https://github.com/zodl-inc/zodl-android/pull/2481)

PR #2482, opened 2026-08-27
MOB-1814: Disable Coinholder Polling home prompt + skip voting network refresh when nothing pending

noop-sk:
## Summary

Two related fixes for Lukas's Slack pushback (`#wallet-team`, 2026-08-26) on the Coinholder Polling home-screen prompt added in #2472 (MOB-1805):

### 1. Disable the home-screen prompt (original scope)

Adds `COINHOLDER_HOME_PROMPT_ENABLED = false` next to the existing `VOTING_ENABLED` switch in `FeatureVotingImpls.kt`, short-circuiting `observeIsCoinholderPollingMessageVisible()` before any flow subscription. Kill switch, not a revert - the banner will likely ship again, redesigned, once eligibility/network-call concerns are addressed, and this keeps the UI/strings/`NavigateToVotingUseCase` extraction in place for that.

### 2. Skip the unconditional voting-session network refresh (added after further discussion)

Turns out disabling the prompt does **not** touch the actual network-call complaint. `VotingHomeMessageSourceImpl.observeIsCoinholderPollingMessageVisible()` (#2472's own code) never called the network itself - the `fetchServiceConfig()` + `fetchAllRounds()` Lukas objected to lives in `RefreshActiveVotingSessionUseCase`, called unconditionally from `VotingHomeHooksImpl.recoverPendingRouteIfNeeded()` on **every** home-screen open, for **every** user, regardless of whether they have any pending vote at all. This is pre-existing code from #2406/MOB-1678 (shipped in 3.9.x), untouched by #2472.

The check for "does this user actually have anything to recover" ran *after* the network call, not before it - so the refresh happened even for users who've never touched voting.

Fix: added `VotingRecoveryRepository.getRoundIdsWithPendingKeystoneRequest(accountUuid)`, a purely local lookup (reuses the same on-device index as the existing `getRoundIdsRequiringShareTracking`), and check it *first*. `refreshActiveVotingSession()` now only runs when a pending Keystone request actually exists locally. The existing safety check (round must still be listed active after refresh) is preserved, just moved after the local check.

## CHANGELOG

Removed the `[Unreleased]` bullet advertising the prompt, so 3.10.2's release notes don't promise a feature users won't see.

## Verification

- ktlint + detekt clean, full `feature-voting` unit test suite green.
- New `VotingHomeHooksImplTest` (previously no tests existed for `VotingHomeHooksImpl` at all): covers skip-when-nothing-pending, refresh-and-navigate-when-pending, and the safety check for a round that closed between the local check and the refresh.
- Red-green verified both fixes: reverting the prompt disable or the refresh-skip individually fails their respective new tests; restoring passes them again.
- Verified live on the `Z_-_Droid` emulator: an account with an active, eligible Coinholder Poll round no longer shows the home-screen prompt after this change, while other home messages (e.g. Currency Conversion) still render normally.

Ref: MOB-1814 (https://linear.app/zodl/issue/MOB-1814)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,6 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
-
-### Added:
-
-- We added a Wallet Status Widget prompt for Coinholder Polling, so you can jump straight into an active poll you haven't voted in yet.
-
 ## [3.10.1 (2512)] - 2026-08-25
 
 ### Added:
--- a/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/repository/VotingRecoveryRepository.kt
+++ b/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/repository/VotingRecoveryRepository.kt
@@ -267,6 +267,15 @@ interface VotingRecoveryRepository {
      * short-circuits when there are no unconfirmed shares.
      */
     suspend fun getRoundIdsRequiringShareTracking(accountUuid: String): List<String>
+
+    /**
+     * Enumerates round ids for the given account with a stored [VotingPendingKeystoneRequest] -
+     * a Keystone signature this device is still waiting on. Purely local (backed by the same
+     * on-device index as [getRoundIdsRequiringShareTracking]): callers can check "is there
+     * anything to recover at all" without first fetching the active session list from the
+     * network, and only pay for that fetch when this returns a non-empty list.
+     */
+    suspend fun getRoundIdsWithPendingKeystoneRequest(accountUuid: String): List<String>
 }
 
 class VotingRecoveryRepositoryImpl(
@@ -721,7 +730,16 @@ class VotingRecoveryRepositoryImpl(
         removeFromIndex(preferenceProvider, accountUuid, roundId)
     }
 
-    override suspend fun getRoundIdsRequiringShareTracking(accountUuid: String): List<String> {
+    override suspend fun getRoundIdsRequiringShareTracking(accountUuid: String): List<String> =
+        roundIdsMatching(accountUuid) { snapshot -> snapshot.submittedProposalIds.isNotEmpty() }
+
+    override suspend fun getRoundIdsWithPendingKeystoneRequest(accountUuid: String): List<String> =
+        roundIdsMatching(accountUuid) { snapshot -> snapshot.pendingKeystoneRequest != null }
+
+    private suspend fun roundIdsMatching(
+        accountUuid: String,
+        predicate: (VotingRecoverySnapshot) -> Boolean
+    ): List<String> {
         val preferenceProvider = encryptedPreferenceProvider()
         val scopedAccount = accountUuid.lowercase()
         val pairs = readIndex(preferenceProvider)
@@ -737,11 +755,7 @@ class VotingRecoveryRepositoryImpl(
                         removeFromIndex(preferenceProvider, scopedAccount, indexedRoundId)
                         return@mapNotNull null
                     }
-                if (snapshot.submittedProposalIds.isNotEmpty()) {
-                    snapshot.roundId
-                } else {
-                    null
-                }
+                snapshot.roundId.takeIf { predicate(snapshot) }
             }
     }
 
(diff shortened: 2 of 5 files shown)


noop-sk (commented):
Reviewed the kill switch for correctness — the key question was whether it also suppresses the unconditional network refresh Lukas flagged, or only the widget UI.

**Verified via the actual call sites (not just the description): only the UI is gated, and that's correctly disclosed, not hidden.**
- `FeatureVotingImpls.kt:206` — new `if (!VOTING_ENABLED || !COINHOLDER_HOME_PROMPT_ENABLED) return flowOf(false)` in `VotingHomeMessageSourceImpl.observeIsCoinholderPollingMessageVisible()` — this only controls widget rendering.
- `FeatureVotingImpls.kt:89-95` — `VotingHomeHooksImpl.recoverPendingRouteIfNeeded()` still calls `refreshActiveVotingSession()` unconditionally on every home open (gated only by the pre-existing `VOTING_ENABLED`), untouched by this PR. Confirmed via `HomeVM.kt:149` that this is a genuinely separate code path from the widget-visibility flow — the two were never coupled, so there's no hidden gap here, just an honestly-scoped fix matching the PR description's own disclosure that the refresh redesign is separate follow-up work.

Everything else checks out:
- Clean `const val` early-return, mirrors the existing `VOTING_ENABLED` pattern, doc comment correctly cross-references what it does and doesn't silence.
- Eligibility logic left in place but now dead — intentional, avoids redoing that work when the redesigned prompt ships; not confusing given the doc comment.
- CHANGELOG entry removed cleanly, no dangling artifacts.
- Zero overlap with the concurrent #2481 hotfix.
- CI: everything completed so far is green (some checks still in flight at review time), no tests existed for this banner before or after — acceptable for an emergency kill-switch PR.

No new findings, nothing blocking. (Posting as COMMENT, not APPROVE — GitHub blocks self-approval and this PR is authored by our own account.)

`https://github.com/zodl-inc/zodl-android/pull/2482`

nesence-m (approved):

nesence-m (commented):
Reviewed at head 22d1190. Independently verified the kill switch and the refresh skip — nothing blocking, but one doc fix is worth doing in this PR before merge.

**Verified working:**
- **Kill switch is complete.** The prompt's only production trigger is `GetHomeMessageUseCase` combining `observeIsCoinholderPollingMessageVisible()`; with the gate returning `flowOf(false)` the message is unreachable. `HomeMessageCacheRepository` is in-memory only (no persisted replay path), no deep link targets the prompt, and the More-screen voting entry stays intact (`VOTING_ENABLED` untouched). Hard-off compile-time default is the safe direction for the 3.10.2 cut.
- **Refresh skip introduces no regression.** `RefreshActiveVotingSessionUseCase` has exactly two production callers — this recovery hook and `VoteCoinholderPollingVM` (which fetches its own data on screen entry). All other `snapshot` consumers sit behind navigation that flows through the polling screen or through the recovery path itself, which still refreshes before reading the snapshot. The round-still-active safety check after refresh is preserved.
- **`VotingRecoveryRepository` refactor** (`roundIdsMatching` extraction) preserves `getRoundIdsRequiringShareTracking` semantics exactly, and the new `VotingHomeHooksImplTest` genuinely covers skip-when-nothing-pending, refresh-and-navigate-when-pending, and the round-closed-mid-flight case.

**Please fix before merge — stale KDoc on `COINHOLDER_HOME_PROMPT_ENABLED`** (`feature-voting/.../FeatureVotingImpls.kt:69-77`): it still claims this PR does **not** touch the `RefreshActiveVotingSessionUseCase` call in `recoverPendingRouteIfNeeded` — but lines 93-103 of the same file now gate exactly that call. And the "re-enable by flipping this back to `true`" promise is no longer accurate: `observeIsCoinholderPollingMessageVisible()` reads `votingApiRepository.snapshot`, whose only home-screen population path was the now-skipped unconditional refresh — flipping the constant alone leaves the snapshot empty on home open until the user visits the polling screen. The re-enabler will read this comment first; it should say the redesign needs to restore a data source.

**Nits (no action needed to merge):**
- Branch name `bugfix/MOB-1814` is missing the kebab-case description suffix per the team pattern (`bugfix/MOB-XXXX-<description>`).
- The refresh skip could arguably carry its own CHANGELOG "Fixed" bullet (fewer network calls on home open), though it's user-invisible.

**CI:** `test_android_modules_emulator (app)` failed on an emulator infrastructure flake (`ScreenshotTest.takeScreenshotsForRestoreWalletDarkEnUS` → device offline, "Expected 12 tests, received 4") in a module this diff doesn't touch — needs a re-run, not a code change. `e2e-android (phase 2)` and `test_robo_release` were still in progress at review time; everything else green including the new unit tests, detekt, ktlint, lint. Merge state currently UNSTABLE pending those.

Happy to approve as soon as the KDoc is fixed and CI settles.

PR #2483, opened 2026-08-27
Voting speed investigation (MOB-1808): fix redundant/duplicate CHP backend requests

noop-sk:
## Summary

Investigates and fixes why the Coinholder Polling (CHP) voting screen was slow and firing excessive/redundant backend requests — [MOB-1808](https://linear.app/zodl/issue/MOB-1808/voting-speed-investigation).

Root causes found and fixed, all confirmed live on-device via logcat request tracing (not speculation):

- **Duplicate `/rounds` fetch every auto-refresh tick.** `RefreshActiveVotingSessionUseCase` fetches config *and* the full round list, but `VoteCoinholderPollingVM.refreshVotingDataInternal()` was calling it purely to surface a config exception, right after `RefreshVotingRoundsUseCase` had already fetched the same round list. Extracted a new minimal `RefreshVotingServiceConfigUseCase` (config-fetch-and-store only) for the round-list VM to use instead. `RefreshActiveVotingSessionUseCase` itself is untouched — `VotingHomeHooksImpl`'s pending-Keystone-request recovery check genuinely needs its full round-list-populating behavior. Verified live: `/rounds` went from 2x/cycle to 1x/cycle.
- **Unnecessary periodic auto-refresh on the round list.** The screen polled every 5-30s in the background even when idle. Investigation showed `VoteTallyingVM` already runs its own dedicated, purpose-scoped polling for the one genuinely time-sensitive live transition (ACTIVE -> TALLYING/COMPLETED) a user might be waiting on — the round list's own polling was redundant for that case, and "return to the screen after voting" is already handled correctly by the existing `onScreenEntered()`/`ON_RESUME` reactive path. Removed the periodic auto-refresh entirely; the screen is now purely reactive to entry/return plus manual pull-to-refresh. Verified live: a 24s idle capture on the screen showed exactly 1 `/rounds` + 1 `/endorsed-rounds` request (from entry), then zero further backend activity.
- **TTL cache fallthrough bug** (found by an adversarial Fable-model review): `getResolvedConfigWithTtl()`'s expiry fallthrough called into a cache check with no actual age awareness, so an expired-but-source-matching config entry was silently re-served forever. Consolidated both config-cache paths into a single mutex-guarded `resolveConfigCached()` to close both the fallthrough and a latent race between two callers independently deciding "expired."
- **Share-delegation concurrency/cancellation issues** (also found by the same review): unbounded per-share fan-out sharing one Tor client (serialized, not actually concurrent) with a cancellation-swallowing bug that could fire extra requests after a batch already failed and mis-attribute failures to healthy servers. Fixed with per-share isolated Tor clients, a concurrency cap, correct `CancellationException` rethrow, and an app-side timeout fallback for the Ktor-no-op-under-Tor case. Same shared-client bug fixed in the PIR snapshot resolver's probe fan-out.
- **Service config cache TTL** bumped from 60s to 10min once the fallthrough bug above was fixed for real.

Two independent Fable-model adversarial reviews were run against this branch (once mid-branch, once against the final commits) — both came back with no blocking issues; all findings from both were fixed in follow-up commits already on the branch.

## Test plan

- [x] `:zodl-android:ktlint` and `:zodl-android:detektAll` both pass clean (3 genuinely new detekt findings from this branch's own commits were fixed for real, not baselined: `InstanceOfCheckForException`, `MaxLineLength`, `ForbiddenImport`, plus a `SwallowedException` introduced while fixing the first).
- [x] `feature-voting` unit test suite (187 tests) passes clean on committed code.
- [x] Live-verified on-device (mainnet DIM + testnet DIT builds) throughout: `zodl_an_apk_rust_check` passes (native lib/JNI/16KB-alignment), zero crashes, full vote-submission flow tested successfully multiple times.
- [x] Live logcat request tracing confirmed both the duplicate-`/rounds`-fetch fix (2x -> 1x per cycle) and the auto-refresh removal (zero background requests while idling on the screen).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+### Added:
+
+- Coinholder Polling now shows a pull-to-refresh indicator on the poll list, so you can manually refresh it on demand.
+
+### Changed:
+
+- Coinholder Polling's poll list no longer refreshes itself in the background while you're just browsing it; it now updates when you open the screen, return to it, or pull to refresh.
+- Vote submission is faster and more reliable over Tor: a slow or unresponsive vote-helper server no longer stalls the whole submission.
+
 ## [3.10.1 (2512)] - 2026-08-25
 
 ### Added:
--- a/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/PirSnapshotResolver.kt
+++ b/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/PirSnapshotResolver.kt
@@ -36,14 +36,22 @@ class HttpPirSnapshotResolver(
             throw PirSnapshotResolverException.NoEndpointsConfigured
         }
 
+        // MOB-1808: probes used to fan out concurrently but all through ONE shared/isolated Tor
+        // HttpClient (built once outside the loop). Confirmed live (via the identical pattern in
+        // VotingApiProvider.delegateShares) that the native Tor engine serializes concurrent
+        // requests on a single isolated client/stream regardless of Kotlin-level concurrency —
+        // so this fan-out was likely no faster than probing endpoints one at a time. Fix: give
+        // each concurrent probe its own freshly-built isolated Tor client instead.
         val outcomes =
-            httpClientProvider.create().use { client ->
-                coroutineScope {
-                    normalizedEndpoints
-                        .map { url ->
-                            async { probe(client, url, expectedSnapshotHeight) }
-                        }.awaitAll()
-                }
+            coroutineScope {
+                normalizedEndpoints
+                    .map { url ->
+                        async {
+                            httpClientProvider.create().use { client ->
+                                probe(client, url, expectedSnapshotHeight)
+                            }
+                        }
+                    }.awaitAll()
             }
 
         return outcomes
--- a/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/VotingApiProvider.kt
+++ b/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/VotingApiProvider.kt
@@ -63,7 +63,9 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.Semaphore
 import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.sync.withPermit
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 import org.json.JSONObject
@@ -157,11 +159,16 @@ class KtorVotingApiProvider(
     private val configRequestTimeoutMillis: Long = StaticVotingConfig.CONFIG_REQUEST_TIMEOUT_MS,
 ) : VotingApiProvider {
     private var cachedResolvedConfig: ResolvedVotingConfig? = null
+    private var cachedResolvedConfigFetchedAtMs: Long? = null
     private val configMutex = Mutex()
     private val serverHealthTracker = VotingServerHealthTracker()
     private val httpClientMutex = Mutex()
     private var cachedHttpClient: HttpClient? = null
     private var cachedHttpClientSupportsKtorTimeouts: Boolean? = null
+
+    // Bounds how many isolated Tor clients delegateShares() builds at once (MOB-1808) — see the
+    // comment at its call site for why an uncapped fan-out is a real client- and server-side risk.
+    private val shareDelegationSemaphore = Semaphore(MAX_CONCURRENT_SHARE_DELEGATIONS)
 
     override suspend fun validateConfigSource(source: PinnedConfigSource) {
         fetchStaticConfig(listOf(source))
@@ -170,11 +177,26 @@ class KtorVotingApiProvider(
     override suspend fun invalidateConfigCache() {
         configMutex.withLock {
             cachedResolvedConfig = null
+            cachedResolvedConfigFetchedAtMs = null
         }
     }
 
+    // MOB-1808: fetchServiceConfig() used to force-refresh unconditionally on every call. With
+    // the CHP round list open during an active round, at least 3 independent loops call this
+    // path — the 5s round-status auto-refresh, the 15s+ foreground share-rediscovery driver, and
+    // ad-hoc tally/tx-confirmation polling — each re-fetching the full static+dynamic config over
+    // Tor from scratch, never reusing what a sibling loop fetched moments earlier ("CHP is
+    // actually firing a lot of requests on its own from its screens" per the team's own Slack
+    // report). getResolvedConfig()'s cache was already source-aware (a config-source change
+    // still forces a real refetch) but forceRefresh=true bypassed it entirely regardless of age.
+    // Switching to TTL-bounded reuse (getResolvedConfigWithTtl) keeps every one of those loops
+    // fed from one shared fetch instead of each paying its own round trip, while still bounding
+    // how stale the served config can get. Callers that need a guaranteed-fresh fetch (e.g. a
+    // config-source change, or the submit flow, which already calls invalidateConfigCache()
+    // explicitly before it enters) go through that explicit invalidation instead of relying on
+    // this method to force one.
     override suspend fun fetchServiceConfig(): VotingServiceConfig =
-        getResolvedConfig(forceRefresh = true).serviceConfig
+        getResolvedConfigWithTtl().serviceConfig
 
     override suspend fun fetchActiveVotingSession(): VotingSession? =
         try {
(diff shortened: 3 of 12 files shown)


nesence-m (commented):
Reviewed at head 7f78d88 (picked up by sweep). Verified the core claims independently — no blocking correctness issues, but a few things worth addressing before merge.

**Verified sound:**
- The duplicate `/rounds` fix is real: `RefreshVotingRoundsUseCase` already does `fetchServiceConfig()` + `fetchAllRounds()` + endorsements, so the old follow-up `RefreshActiveVotingSessionUseCase` call was redundant, and the new `RefreshVotingServiceConfigUseCase` preserves the one non-redundant side effect (storing the `VotingConfigSnapshot`).
- The TTL cache does not weaken the recovery safety check from #2481: every consumer that decides round activity still live-fetches `/rounds`; only the config leg is TTL-cached, with `invalidateConfigCache()` still forced on screen entry / pull-to-refresh / source change. No overlap or conflict with #2482's refresh gating (`VotingHomeHooksImpl`/`FeatureVotingImpls` untouched, `mergeable: MERGEABLE`).
- Cancellation handling in `postShare` is correct for this Ktor version — checked against Ktor 3.4.0 source that `HttpRequestTimeoutException` extends `IOException`, not `CancellationException`, so the catch order behaves as intended and a Ktor timeout can't kill sibling shares.
- Concurrency primitives check out: `resolveConfigCached`'s mutex coalesces waiters onto the fresh entry, the shared-client cache is cleared before rebuild (no closed-client caching), `VotingServerHealthTracker` is mutex-guarded under the new 16-way fan-out, per-share clients closed via `.use`.
- `delegateShares` failure contract is unchanged in shape — batch retry could always re-post already-accepted shares (pre-existing server-idempotency assumption, not new risk).

**Worth addressing before merge:**
1. **Tests don't pin any of the new behavior.** The only test change (`VotingConfigWalkTest.kt:414-419`) *accommodates* the TTL cache rather than asserting it — it passes with or without the provider change. Nothing covers TTL reuse/expiry, the near-dead-serving fallthrough fix, the `CancellationException` rethrow, the semaphore cap, or the 2x→1x dedupe. For a PR whose core is a caching/concurrency primitive, a regression would be silent — at minimum a TTL-reuse + TTL-expiry test and a cancellation-rethrow test would pay for themselves.
2. **No CHANGELOG entry** despite user-visible changes (pull-to-refresh indicator added, background auto-refresh of the round list removed, submission much faster) — sibling #2482 updated `[Unreleased]` on this same base.
3. **The `configIssue` error-sheet path is now effectively unreachable** (`VoteCoinholderPollingVM.kt:544-561`): the probe inside `runCatching` runs right after `refreshVotingRounds()` already fetched the same config, so it's a guaranteed TTL-cache hit — deterministic config failures now surface as the generic rounds error instead of the dedicated sheet. Either drop the dead probe or reorder so config failures reach the sheet as designed; your call, but decide it explicitly.

**Nits / conscious trade-offs to confirm:**
- One-frame stale-rounds flash on config-source switch (`VoteCoinholderPollingVM.kt:175`) — the "already clears the repository first" comment is only asynchronously true.
- `PullToRefreshBox` wraps only the non-empty branch (`VoteCoinholderPollingView.kt:97`) — with auto-refresh gone, the empty state can only refresh via re-entry.
- Tor-preference flip closes the shared client under in-flight requests (`VotingApiProvider.kt:825`) — user-triggered, worst case one failed request.
- With no list poller, an idle list can show a stale ACTIVE card indefinitely; downstream re-fetch catches taps into expired rounds. Your stated rationale is coherent — just confirming the trade-off is accepted consciously.

**CI:** completed checks all green (build, detekt, ktlint, lint, `test_kotlin_modules`); `e2e-android`, `Keystone`, and the emulator job still in progress at review time; no other reviews yet.

Happy to approve once the CHANGELOG lands, the configIssue question is decided, and CI finishes — the test gap I'd strongly encourage but won't gate on by itself.

nesence-m (changes requested):
Re-verified the fix commit (7f78d88 → 2a73cfc) against each original point rather than the summary. All three are substantively closed — nice round — but the configIssue reorder introduces one real regression that I'd like fixed (or explicitly declined with rationale) before I approve.

**Original findings:**
1. **Tests — FIXED.** Traced all five new tests against production: each genuinely fails if the guarded behavior is reverted. `postShareTimeoutIsTreatedAsAFailureNotACancellation` is particularly well-built — the `assertFalse(exception is CancellationException)` is what saves it from passing vacuously (CE extends ISE). Residual gaps, none gating: (a) the TTL-**expiry** fallthrough is still unpinned — `invalidateConfigCacheForcesRefetchOnTheTtlPath`'s comment claims it's "the direct regression pin for the fallthrough bug", but invalidation nulls the cache and exercises the cache-miss branch; the old buggy code would also pass that test. Either correct the comment or add a `nowMillis` seam (you added a timeout seam for postShare, so the pattern exists). (b) postShare's plain-external-CE rethrow branch is untested. (c) The VM-level dedupe gap you flagged yourself — accepted as disclosed.
2. **CHANGELOG — FIXED.** Right section, accurate entries, and no conflict with #2482 (base's `[Unreleased]` is empty; PR is MERGEABLE).
3. **configIssue dead path — FIXED as asked.** The catch now sits on the first (real) config fetch, the second call survives only for `votingConfigRepository.store()`, and I verified your "only source of VotingConfigException" claim independently — including the subtle one: `VotingRoundAuthenticationException` does extend `VotingConfigException` but is swallowed inside `fetchAllRounds` via `authenticateVotingSessionOrNull`, so it can't leak through. `configIssue` genuinely reaches the sheet at round-open.

**New issue from the catch-move (the blocker):** the catch converts *every* config-leg failure — including plain offline/transient network failures, since `StaticVotingConfigFetchFailedException` and `DynamicVotingConfigTransientException` both extend `VotingConfigException` (`VotingApiProvider.kt:1036-1053`) — from an LCE error into LCE success. When the rounds repo is empty (first-ever screen entry, or after a config-source switch's hard clear), `roundsLceState` at `VoteCoinholderPollingVM.kt:181` yields `Success → emptyList()` and the user gets the `noRoundsSheet` ("there are no polls") instead of the error state + retry that 7f78d88 showed — and the configIssue sheet is unreachable because it only fires on tapping an ACTIVE card, of which there are none. Compounded by `PullToRefreshBox` wrapping only the non-empty branch, the offline user has no recovery path on that screen. Suggested shape: when the caught exception is set as `nextConfigIssue` and the repo snapshot has no rounds, rethrow (or surface the error/config sheet directly) instead of completing successfully.

**CI at 2a73cfc:** ktlint/detekt/lint/build/`test_kotlin_modules` all green (consistent with your 261-green claim). `test_android_modules_emulator (app)` failed on the restore-wallet screenshot test with emulator ColorBuffer errors — unrelated to voting, same flake signature as this morning's PRs, needs a re-run. e2e phases still in progress. `mergeStateStatus: BLOCKED` pending review + checks.

Happy to approve as soon as the empty-repo/offline path is handled (plus the test-comment correction from 1a, trivially) and CI settles.

nesence-m (approved):
Round 3 verified at head 4efd182 — both round-2 findings are genuinely fixed, and the fix commit contains nothing outside its stated scope.

**Offline/empty-repo blocker — FIXED.** The guard (`VoteCoinholderPollingVM.kt:601-606`) rethrows when a config-leg failure occurred AND the rounds repo is empty, exactly the suggested shape. I traced all four quadrants: offline+empty now reaches the LCE error state **with retry** (via `MutableLce.execute`'s standard error path — the same mechanism every other round-fetch failure uses); offline+cached still shows cached rounds with `configIssue` armed (the guard sits after arming, so nothing is lost); config-issue+cached preserves the round-1 configErrorSheet fix verbatim at round-open; and a genuinely-empty successful fetch still gets the `noRoundsSheet`. Ordering is safe (config fetched before rounds, so the empty check reads a consistent snapshot), and `CancellationException` is correctly rethrown by the LCE wrapper.

**TTL test — FIXED.** The `configCacheTtlMillis` seam is production-safe (default unchanged, Koin passes nothing), and the new expiry test genuinely pins the fallthrough bug: entry present, source-matching, aged past TTL — against the old code the second fetch re-serves cache and the request-count assertion fails. Real-time delay is ~40ms, deterministic in the correct direction. The old test's overclaiming comment is corrected in place.

**CI**: everything code-relevant is green, including `test_kotlin_modules` (the 262-test feature-voting suite). `test_android_modules_emulator (app)` failed again — logs are pending while the run completes, but this is the same job that failed with the identical restore-wallet/emulator-death signature on #2481, #2482, and this PR's previous head today, and this commit touches nothing it exercises; treating it as the known chronic breakage. If its logs turn out to show a voting test, I'll take another look. e2e/build/Keystone still in progress — they should finish green before merge (branch protection blocks it anyway).

Reminder for whoever merges second: this PR and #2484 will conflict both on the `withConfigRequestTimeoutFallback` → `withTorRequestTimeoutFallback` rename and on CHANGELOG (both insert at the same `[Unreleased]` anchor) — coordinate the order and rebase the survivor.

[#2483 Voting speed investigation (MOB-1808): fix redundant/duplicate CHP backend requests](https://github.com/zodl-inc/zodl-android/pull/2483)

PR #2484, opened 2026-08-27
MOB-1811: Bound vote-server failover attempts under Tor

noop-sk:
## Summary

Fixes [MOB-1811](https://linear.app/zodl/issue/MOB-1811/android-need-to-defer-health-checks-to-load-the-polls-faster-in-the) (Urgent) — the CHP round list and pre-vote screen could hang for a long, inconsistent time before showing content, reported in Slack as "waiting on a health check before displaying the proposal."

**Root cause:** `withVoteServerFailover()` (`VotingServerFailoverException.kt`) tries each of the ~10 configured vote servers one at a time with **no per-attempt timeout**. Under Tor (`installTimeouts = false`), Ktor's own `HttpTimeout` plugin is a no-op, so a server that *drops* (rather than refuses) the connection can hang one attempt indefinitely before the loop's catch even fires to try the next server. This is the single choke point (`executeWithVoteServerFailover` in `VotingApiProvider.kt`) behind both:
- the CHP round list refresh (`fetchAllRounds()`), and
- the pre-vote screen's `preparationGate` (via `PrepareVotingRoundUseCase` → `fetchServiceConfig()`/`fetchAllRounds()`), which shows only a shimmer until this resolves.

**Fix (scoped to the minimal, low-risk option — no server-health-based skipping or parallel/racing, per explicit decision):**
- Wrap every `executeWithVoteServerFailover()` attempt in the existing `withConfigRequestTimeoutFallback()` app-side timeout helper (already used for the config-fetch legs since MOB-1809), so a dropping server fails fast instead of hanging.
- Fix `withVoteServerFailover()`'s catch ordering: the resulting `TimeoutCancellationException` (which extends `CancellationException`) must be caught **before** the plain `CancellationException` branch, or a deliberate per-attempt timeout gets misread as genuine outer cancellation and aborts the whole failover walk instead of falling through to the next server.
- `shouldTryNextVoteServer()` gets the same special case so its default classification also treats a timeout as retryable, not as a reason to give up.

New constructor param `voteServerFailoverTimeoutMillis` (default 15s, matching the existing config-timeout convention) — overridable only for tests.

## Test plan

- [x] `:zodl-android:ktlint` / `:zodl-android:detektAll` clean.
- [x] New unit tests in `VoteServerFailoverTest.kt`: a `TimeoutCancellationException` from one server attempt now falls through to the next server instead of aborting the walk; a genuine outer `CancellationException` still propagates correctly (regression guard).
- [x] New integration test in `KtorVotingApiProviderTest.kt`: `fetchAllRounds()` against a vote server that hangs (`awaitCancellation()`, simulating a Tor connection being dropped) under `supportsKtorTimeouts = false` now fails fast via `VotingServerFailoverException` instead of hanging.
- [x] Full `feature-voting` unit test suite green (`VoteServerFailoverTest` 12/12, `KtorVotingApiProviderTest` 24/24, no failures module-wide).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+### Fixed:
+
+- Coinholder Polling's poll list and proposal loading no longer hang indefinitely when a vote server is unreachable over Tor; a slow or dropped connection now fails over to the next server within a bounded time.
+
 ## [3.10.1 (2512)] - 2026-08-25
 
 ### Added:
--- a/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/VotingApiProvider.kt
+++ b/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/VotingApiProvider.kt
@@ -154,6 +154,18 @@ class KtorVotingApiProvider(
      * timeout mechanisms share.
      */
     private val configRequestTimeoutMillis: Long = StaticVotingConfig.CONFIG_REQUEST_TIMEOUT_MS,
+    /**
+     * Per-attempt timeout for every [executeWithVoteServerFailover] call (MOB-1811), overridable
+     * only so tests can inject a short value. One of the ~10 configured vote servers can drop
+     * (not refuse) a Tor connection instead of failing it outright; under Tor
+     * ([HttpClientProvider]'s isolated client has `installTimeouts = false`) Ktor's own
+     * [io.ktor.client.plugins.HttpTimeout] is a no-op, so without this an attempt against that
+     * server can hang indefinitely instead of the failover loop moving on to the next one -
+     * exactly the "waiting for a health check before displaying the proposal" symptom reported
+     * against the round list and the pre-vote screen's [co.electriccoin.zcash.ui.common.usecase
+     * .PrepareVotingRoundUseCase] gate.
+     */
+    private val voteServerFailoverTimeoutMillis: Long = VOTE_SERVER_FAILOVER_TIMEOUT_MILLIS,
 ) : VotingApiProvider {
     private var cachedResolvedConfig: ResolvedVotingConfig? = null
     private val configMutex = Mutex()
@@ -659,12 +671,21 @@ class KtorVotingApiProvider(
         block: suspend HttpClient.(String) -> T
     ): T {
         val serverUrls = configuredVoteServerUrls()
-        return execute {
+        return executeWithKtorTimeoutSupport { supportsKtorTimeouts ->
             withVoteServerFailover(
                 path = path,
                 serverUrls = serverUrls,
                 shouldTryNext = shouldTryNext,
-                operation = { serverUrl -> block(serverUrl) }
+                operation = { serverUrl ->
+                    // MOB-1811: bounds each per-server attempt so one dropping (not refusing)
+                    // Tor connection can't stall the whole failover walk - see the constructor
+                    // param doc above and withConfigRequestTimeoutFallback's own TRAP doc for why
+                    // withVoteServerFailover's catch clauses must (and do) special-case the
+                    // TimeoutCancellationException this throws under Tor.
+                    withConfigRequestTimeoutFallback(supportsKtorTimeouts, voteServerFailoverTimeoutMillis) {
+                        block(serverUrl)
+                    }
+                }
             )
         }
     }
@@ -1012,6 +1033,9 @@ private fun String.withCacheBustIfNeeded(token: String): String {
 private const val HELPER_REQUEST_TIMEOUT_MILLIS = 5_000L
 private const val HELPER_SOCKET_TIMEOUT_MILLIS = 10_000L
 private const val HELPER_CONNECT_TIMEOUT_MILLIS = 5_000L
+
+// MOB-1811: matches CONFIG_REQUEST_TIMEOUT_MS's 15s convention for a bounded network attempt.
+private const val VOTE_SERVER_FAILOVER_TIMEOUT_MILLIS = 15_000L
 private const val TAG = "VotingApiProvider"
 private const val ACTIVE_ROUNDS_PATH = "/shielded-vote/v1/rounds/active"
 private const val ROUNDS_PATH = "/shielded-vote/v1/rounds"
--- a/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/VotingApiProvider.kt
+++ b/feature-voting/src/main/java/co/electriccoin/zcash/ui/common/provider/VotingApiProvider.kt
@@ -242,6 +242,16 @@ class KtorVotingApiProvider(
         }
     }
 
+    // MOB-1811: these two writes go through the same withVoteServerFailover bound as every read,
+    // so a Tor-slow-but-alive server now gets its payload re-POSTed to the next server after
+    // voteServerFailoverTimeoutMillis instead of hanging indefinitely. Accepted trade-off, not a
+    // new hazard: the body is re-posted byte-identical, so at most one of the two attempts can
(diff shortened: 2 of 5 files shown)


nesence-m (commented):
Reviewed at head 49f9060. The core fix is correct and well-tested — nothing blocking — but a few items should be handled before/alongside merge, so holding the approval until the CHANGELOG lands and CI finishes.

**Verified sound:**
- The cancellation semantics are the load-bearing subtlety here, and they're right: `TimeoutCancellationException` is caught before plain `CancellationException` and the latter is rethrown (`VotingServerFailoverException.kt:36-52`), so external cancellation is never eaten and the walk aborts properly. The new catch correctly mirrors the generic branch (records `lastError`, honors `shouldTryNext`), so the endorsed-rounds custom classifier still records statuses.
- Tests genuinely pin the fix: the TCE-falls-through test fails against the pre-fix code (which rethrew TCE and aborted the walk), the genuine-CE regression guard pins the rethrow, and the hanging-server test (`KtorVotingApiProviderTest.kt:449-509`) hangs forever without the fix.
- Worst-case total is ~10 × 15s = 150s before `VotingServerFailoverException` vs. the old infinite hang — acceptable.

**Please acknowledge explicitly — write-path re-submission trade-off.** `submitDelegation`/`submitVoteCommitment` route through the same failover walk (`VotingApiProvider.kt:245-262`), so a Tor-slow-but-alive server now gets its vote re-POSTed to the next server after 15s. I traced the semantics: the payload is built once and re-posted byte-identical, two servers may wrap it into two different chain txs, and the second is a spent-nullifier rejection on-chain — it cannot double-count, and the reconcile/VAN-probe machinery from #2481 exists for exactly this ambiguity, which pre-existed via 5xx/IOException failover (and clearnet's client-level retry). What the timeout changes is probability, not possibility: the degraded-UX window (server N's tx lands after the app accepted N+1's hash → N+1 rejected → user-visible "transaction failed" until the recovery re-run) becomes more likely under Tor. A sentence in the PR accepting this trade-off — and a follow-up idea of probing `fetchTxConfirmation`/VAN before re-posting after a *timeout* specifically — would close this.

**Coordinate merge order with #2483.** "Independent" is overstated: #2483 renames `withConfigRequestTimeoutFallback` → `withTorRequestTimeoutFallback` and edits the same constants/constructor-docs regions of `VotingApiProvider.kt`; this PR calls the helper by its old name (`VotingApiProvider.kt:685`). Whichever merges second either conflicts textually or fails compilation on the rename — loud and CI-caught, so no silent double-wrapping (confirmed #2483 doesn't touch `withVoteServerFailover`), but the second PR needs a rebase pass.

**Missed sibling with the same bug class:** `fetchTxConfirmation`'s hand-rolled server walk (`VotingApiProvider.kt:400-431`) iterates all vote servers manually — not via `executeWithVoteServerFailover` — and got no per-attempt bound. Under Tor a dropping server hangs it indefinitely, and it sits inside the submission critical path (`awaitTxConfirmation` polls it 45×; the #2481 recovery probes use it). The "applied uniformly across all executeWithVoteServerFailover callers" claim is literally true, but this walk was missed. Follow-up ticket is fine; adding it here is also fine.

**Smaller items:**
- **CHANGELOG entry missing** for a user-visible fix on the maint line (sibling #2483 updates it) — please add before merge.
- One shared 15s constant for reads and writes means Tor (the slowest network) gets a tighter bound than clearnet's effective 120s client default — and note the doc inside `withConfigRequestTimeoutFallback` claiming Ktor enforces the same bound per-request is true for the config legs but not for these callers. #2483's live measurements (healthy Tor share POSTs at 0.3–2s) give 15s empirical headroom; consider a more generous write-path bound or a comment accepting 15s deliberately.
- Nit: the TRAP doc's outer-cancellation guarantee (`VotingApiProvider.kt:952-956`) overstates — an outer `withTimeout` expiring mid-walk would surface as a per-attempt TCE and spin the remaining servers before dying. No current caller does this (verified); worth one doc sentence.
- Nit: branch `bugfix/MOB-1811` lacks the kebab description suffix per the team pattern.

**CI:** ktlint/detekt green; build, `test_kotlin_modules`, emulator, and lint still running at review time; `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` pending review + checks.

Will approve once the CHANGELOG entry and the trade-off acknowledgment land and CI completes green — the rest can be follow-ups.

nesence-m (dismissed):
Round 2 verified at head ec70e65 — both gating items closed, and the fix commit is strictly additive docs/comments (CHANGELOG +4, VotingApiProvider.kt +10 comment lines; zero behavior change).

**CHANGELOG — done.** New `### Fixed:` entry under `[Unreleased]`, user-facing wording, accurate to the fix.
**Trade-off acknowledgment — done.** The comment block above `submitDelegation`/`submitVoteCommitment` (`VotingApiProvider.kt:245-254`) states the accepted trade-off without overclaiming safety — pre-existing double-post ambiguity made more likely, bounded by byte-identical re-POST + on-chain spent-nullifier dedup + the #2481 recovery machinery, residual UX window named. The PR reply carries the explicit acceptance and the VAN-probe-before-repost follow-up idea.
**Declined rebase — verified substantive.** The branch genuinely predates #2483's rename (zero occurrences of `withTorRequestTimeoutFallback` at this head) and #2483 is still open, so rebasing now would depend on an unmerged sibling. The coordination point stands, now stronger: whichever of #2483/#2484 merges second will conflict on **both** the helper rename and CHANGELOG (both PRs insert at the same `[Unreleased]` anchor). Loud, not silent — just coordinate and rebase the survivor.

Two asks alongside approval, not gating it: the fetchTxConfirmation unbounded-walk sibling bug and the VAN-probe-before-repost idea currently live only in a code comment and a PR reply — please file MOB follow-up tickets so they don't evaporate.

**CI**: static analysis green; build/emulator/lint jobs were still in progress at review time on a comments-only diff — branch protection holds the merge until they complete, and nothing in this commit can affect them beyond flake.

[#2484 MOB-1811: Bound vote-server failover attempts under Tor](https://github.com/zodl-inc/zodl-android/pull/2484)

nesence-m (approved):
Re-approving after verifying the conflict-resolution merge commit 032baf2 — the previous approval went stale when it landed, so this is a fresh check, not a rubber stamp.

**Structure**: true merge commit (parents ec70e65 — the previously approved head — and 2e0881a, which is both the current maint/v3.10.x tip and #2483's merge commit). No force-push; approved history intact.

**Resolution correctness**: the effective diff vs base is exactly the MOB-1811 change re-expressed on top of #2483 — five files, all MOB-1811 material. The failover call site uses `withTorRequestTimeoutFallback`, with zero remaining references to the old name anywhere. All three constructor seams carry their correct defaults (15s / 15s / 600s) and Koin still passes none of them. CHANGELOG has both PRs' entries exactly once. Spot-checked at head that #2483's content survives identically (VM rethrow guard, TTL cache logic, postShare catch ordering, all its tests) and that #2484's own approved content is complete (failover wrapper, exception catch ordering, trade-off comment, new tests).

**Resolution-specific hazards**: no double-wrapped timeouts — the three `withTorRequestTimeoutFallback` uses sit on disjoint request paths (config walk, postShare, failover per-attempt), and the failover blocks contain no inner wrapper. Test suites coexist without collisions; 262 + 3 = the claimed 265.

**CI**: build, ktlint, detekt, lint, `test_kotlin_modules`, emulator (libs) all green. `test_android_modules_emulator (app)` failed on the chronic restore-wallet screenshot flake again (emulator died mid-run, `device 'emulator-5554' not found`, "Expected 12 tests, received 4") — same signature as every sibling PR today, in a module this merge doesn't touch, and green at the previous head. Re-run it, and let e2e/Keystone finish, before merging.

[#2484 MOB-1811: Bound vote-server failover attempts under Tor](https://github.com/zodl-inc/zodl-android/pull/2484)

PR #2489, opened 2026-08-29
[MOB-1723] Preserve spendable balance during automatic server switching

zcash-harry:
## Issue

MOB-1723

## Summary

- Run automatic server benchmarking during startup.
- Apply its winner only after the SDK publishes the first local balance snapshot.
- Keep the last UI synchronizer, account, wallet snapshot, and balance visible during replacement.
- Replace retained UI state when the new synchronizer is available.
- Clear retained state when the wallet is removed.
- Keep manual server changes and non-UI synchronizer consumers unchanged.

## Root cause

Applying an automatic server winner replaces the synchronizer. The raw synchronizer flow emits `null` during this short rebuild. Send and balance consumers treated that transient value as missing wallet state and showed loading or empty state.

## Result

On cold start, local hydration and server benchmarking run together. If benchmarking wins first, it waits for a non-null local balance map. An empty map is a completed read. The result does not depend on a selected account or a nonzero balance.

The automatic winner can apply while the app is visible. UI consumers keep the prior concrete state during the rebuild. Transaction, worker, lifecycle, and recovery paths still use the raw synchronizer.

Repeated emissions of the same server winner are ignored.

## Testing

- `./gradlew :ui-lib:testZcashmainnetInternalDebugUnitTest --tests co.electriccoin.zcash.ui.common.repository.AutomaticServerRepositoryTest --tests co.electriccoin.zcash.ui.common.provider.SynchronizerProviderTest`
- `./gradlew :feature-voting:compileZcashmainnetInternalDebugUnitTestKotlin`
- `./gradlew ktlintFormat detektAll`
- `git diff --check`

Instrumentation test compilation is currently blocked by existing API mismatches in `MockSynchronizer` and `FakeWalletRepository` against the locally linked SDK. These errors are unrelated to this PR.

## Manual QA

1. Enable automatic server selection and terminate the app.
2. Cold-start and immediately open Send.
3. Confirm the stored spendable balance appears before network synchronization completes.
4. Let the automatic winner apply while the app stays visible.
5. Confirm Send does not show a loading or zero-balance state during replacement.
6. Confirm address validation and sending use the replacement synchronizer.
7. Confirm deleting the wallet clears retained UI state.

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
@@ -3,15 +3,13 @@ package co.electriccoin.zcash.ui.common.datasource
 import android.content.Context
 import cash.z.ecc.android.sdk.Synchronizer
 import cash.z.ecc.android.sdk.model.Account
+import cash.z.ecc.android.sdk.model.AccountBalance
 import cash.z.ecc.android.sdk.model.AccountImportSetup
 import cash.z.ecc.android.sdk.model.AccountPurpose
 import cash.z.ecc.android.sdk.model.AccountUuid
 import cash.z.ecc.android.sdk.model.BlockHeight
 import cash.z.ecc.android.sdk.model.UnifiedAddressRequest
 import cash.z.ecc.android.sdk.model.UnifiedFullViewingKey
-import cash.z.ecc.android.sdk.model.WalletAddress
-import cash.z.ecc.android.sdk.model.WalletBalance
-import cash.z.ecc.android.sdk.model.Zatoshi
 import cash.z.ecc.android.sdk.model.Zip32AccountIndex
 import co.electriccoin.zcash.ui.R
 import co.electriccoin.zcash.ui.common.model.KeystoneAccount
@@ -77,7 +75,7 @@ interface AccountDataSource {
         birthday: BlockHeight? = null
     ): Account
 
-    suspend fun requestNextShieldedAddress(): WalletAddress.Unified
+    suspend fun requestNextShieldedAddress(): String
 
     suspend fun deleteAccount(account: WalletAccount)
 }
@@ -111,71 +109,41 @@ class AccountDataSourceImpl(
                     ?.flatMapLatest { allSdkAccounts ->
                         allSdkAccounts
                             .map { sdkAccount ->
-                                val unifiedAndTransparent =
-                                    combine(
-                                        observeUnifiedAddress(synchronizer, sdkAccount),
-                                        observeUnifiedBalance(synchronizer, sdkAccount),
-                                        observeOrchardBalance(synchronizer, sdkAccount),
-                                        observeTransparentAddress(synchronizer, sdkAccount),
-                                        observeTransparentBalance(synchronizer, sdkAccount),
-                                    ) {
-                                        unifiedAddress,
-                                        unifiedBalance,
-                                        orchardBalance,
-                                        transparentAddress,
-                                        transparentBalance,
-                                        ->
-                                        UnifiedAndTransparentSnapshot(
+                                combine(
+                                    observeAccountBalance(synchronizer, sdkAccount),
+                                    observeUnifiedAddress(synchronizer, sdkAccount),
+                                    observeTransparentAddress(synchronizer, sdkAccount),
+                                    observeSaplingAddress(synchronizer, sdkAccount),
+                                    observeIsSelected(sdkAccount, allSdkAccounts),
+                                ) {
+                                    balance,
+                                    unifiedAddress,
+                                    transparentAddress,
+                                    saplingAddress,
+                                    isSelected,
+                                    ->
+                                    if (isKeystoneAccount(sdkAccount)) {
+                                        KeystoneAccount(
+                                            sdkAccount = sdkAccount,
                                             unifiedAddress = unifiedAddress,
-                                            unifiedBalance = unifiedBalance,
-                                            orchardBalance = orchardBalance,
                                             transparentAddress = transparentAddress,
-                                            transparentBalance = transparentBalance,
-                                        )
-                                    }
-                                val saplingAndSelection =
-                                    combine(
-                                        observeSaplingAddress(synchronizer, sdkAccount),
-                                        observeSaplingBalance(synchronizer, sdkAccount),
-                                        observeIronwoodBalance(synchronizer, sdkAccount),
-                                        observeIsSelected(sdkAccount, allSdkAccounts),
-                                    ) { saplingAddress, saplingBalance, ironwoodBalance, isSelected ->
-                                        SaplingAndSelectionSnapshot(
-                                            saplingAddress = saplingAddress,
-                                            saplingBalance = saplingBalance,
-                                            ironwoodBalance = ironwoodBalance,
+                                            orchardBalance = balance?.orchard,
+                                            ironwoodBalance = balance?.ironwood,
+                                            transparentBalance = balance?.unshielded,
                                             isSelected = isSelected,
                                         )
-                                    }
-                                combine(unifiedAndTransparent, saplingAndSelection) { primary, secondary ->
-                                    when (sdkAccount.keySource?.lowercase()) {
-                                        KEYSTONE_KEYSOURCE -> {
-                                            KeystoneAccount(
-                                                sdkAccount = sdkAccount,
-                                                unifiedAddress = primary.unifiedAddress,
-                                                unifiedBalance = primary.unifiedBalance,
-                                                orchardBalance = primary.orchardBalance,
-                                                ironwoodBalance = secondary.ironwoodBalance,
-                                                transparentAddress = primary.transparentAddress,
-                                                transparentBalance = primary.transparentBalance,
-                                                isSelected = secondary.isSelected,
-                                            )
-                                        }
-
-                                        else -> {
-                                            ZashiAccount(
-                                                sdkAccount = sdkAccount,
-                                                unifiedAddress = primary.unifiedAddress,
-                                                unifiedBalance = primary.unifiedBalance,
-                                                orchardBalance = primary.orchardBalance,
-                                                saplingAddress = secondary.saplingAddress!!,
-                                                saplingBalance = secondary.saplingBalance,
-                                                ironwoodBalance = secondary.ironwoodBalance,
-                                                transparentAddress = primary.transparentAddress,
-                                                transparentBalance = primary.transparentBalance,
-                                                isSelected = secondary.isSelected,
-                                            )
-                                        }
+                                    } else {
+                                        ZashiAccount(
+                                            sdkAccount = sdkAccount,
+                                            unifiedAddress = unifiedAddress,
+                                            transparentAddress = transparentAddress,
+                                            saplingAddress = saplingAddress!!,
+                                            orchardBalance = balance?.orchard,
+                                            saplingBalance = balance?.sapling,
+                                            ironwoodBalance = balance?.ironwood,
+                                            transparentBalance = balance?.unshielded,
+                                            isSelected = isSelected,
+                                        )
                                     }
                                 }
                             }.combineToFlow()
@@ -240,20 +208,20 @@ class AccountDataSourceImpl(
         }
 
     @Suppress("TooGenericExceptionCaught")
-    override suspend fun requestNextShieldedAddress(): WalletAddress.Unified {
-        var result: WalletAddress.Unified? = null
+    override suspend fun requestNextShieldedAddress(): String {
+        var result: String? = null
         scope
             .launch {
                 val accountUuid = getSelectedAccount().sdkAccount.accountUuid
                 log("requestNextShieldedAddress for $accountUuid")
-                val responseChannel = Channel<WalletAddress.Unified>(1)
+                val responseChannel = Channel<String>(1)
                 requestNextShieldedAddressChannel.emit(AddressRequest(accountUuid, responseChannel))
                 try {
                     val address = withTimeoutOrNull(ADDRESS_REQUEST_TIMEOUT) { responseChannel.receive() }
                     if (address == null) {
                         log("timed out waiting for address for $accountUuid")
                     } else {
-                        log("received address ${address.address} for $accountUuid")
+                        log("received address $address for $accountUuid")
                     }
                     result = address
                 } catch (e: Exception) {
@@ -261,7 +229,7 @@ class AccountDataSourceImpl(
                 }
             }.join()
         return (result ?: getSelectedAccount().unifiedAddress).also {
-            log("returning address ${it.address}")
+            log("returning address $it")
         }
     }
 
@@ -284,34 +252,34 @@ class AccountDataSourceImpl(
             }
         }
 
+    private fun isKeystoneAccount(sdkAccount: Account) = sdkAccount.keySource?.lowercase() == KEYSTONE_KEYSOURCE
+
     private fun observeIsSelected(sdkAccount: Account, allAccounts: List<Account>) =
         selectedAccountUUIDProvider
             .uuid
             .map { uuid ->
-                when (sdkAccount.keySource?.lowercase()) {
-                    KEYSTONE_KEYSOURCE -> sdkAccount.accountUuid == uuid || allAccounts.size == 1
-                    else -> uuid == null || sdkAccount.accountUuid == uuid || allAccounts.size == 1
+                if (isKeystoneAccount(sdkAccount)) {
+                    sdkAccount.accountUuid == uuid || allAccounts.size == 1
+                } else {
+                    uuid == null || sdkAccount.accountUuid == uuid || allAccounts.size == 1
                 }
             }
 
     @Suppress("TooGenericExceptionCaught")
-    private fun observeUnifiedAddress(synchronizer: Synchronizer, sdkAccount: Account): Flow<WalletAddress.Unified> {
-        suspend fun rotateAddress(): WalletAddress.Unified {
+    private fun observeUnifiedAddress(synchronizer: Synchronizer, sdkAccount: Account): Flow<String> {
+        suspend fun rotateAddress(): String {
             log("deriving unified address for ${sdkAccount.accountUuid}")
 
             val addressRequest =
-                if (sdkAccount.keySource?.lowercase() == KEYSTONE_KEYSOURCE) {
+                if (isKeystoneAccount(sdkAccount)) {
                     UnifiedAddressRequest.Orchard
                 } else {
                     UnifiedAddressRequest.shielded
                 }
 
-            val address =
-                WalletAddress
-                    .Unified
-                    .new(synchronizer.getCustomUnifiedAddress(sdkAccount, addressRequest))
+            val address = synchronizer.getCustomUnifiedAddress(sdkAccount, addressRequest)
 
-            log("derived address ${address.address} for ${sdkAccount.accountUuid}")
+            log("derived address $address for ${sdkAccount.accountUuid}")
 
             return address
         }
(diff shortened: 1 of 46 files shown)


ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/send/AndroidSend.kt
```diff
@@ -311,7 +336,7 @@ internal fun WrapSend(
                             address = it,
                             // TODO [#342]: Verify Addresses without Synchronizer
                             // TODO [#342]: https://github.com/zcash/zcash-android-wallet-sdk/issues/342
-                            type = synchronizer.validateAddress(it)
+                            type = synchronizer?.validateAddress(it)
```
noop-sk: **Blocking.** With `synchronizer` now nullable here, `type = synchronizer?.validateAddress(it)` can leave `recipientAddressState.type == null`. `sendButtonEnabled` only checks `type !is AddressType.Invalid`, which is true for `null`, so Send becomes tappable on a never-validated address. The `null ->` branch in `SendView.kt`'s destination-`when` (~L351-375) then unconditionally does `WalletAddress.Unified.new(address)` regardless of real format — previously impossible since `validateAddress` was non-nullable. Reachable exactly in this PR's target scenario (typing/pasting an address while an automatic switch tears down the Synchronizer).

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/send/AndroidSend.kt
```diff
@@ -77,7 +79,17 @@ internal fun WrapSend(args: Send) {
 
     val synchronizer = walletViewModel.synchronizer.collectAsStateWithLifecycle().value
 
-    val selectedAccount = accountDataSource.selectedAccount.collectAsStateWithLifecycle(null).value
+    // Keep the database-backed account snapshot while an automatic endpoint change replaces the
+    // Synchronizer. The account is replaced as soon as the new instance publishes its local state.
+    val selectedAccountFlow =
+        remember(accountDataSource) {
+            accountDataSource.selectedAccount.filterNotNull()
```
noop-sk: **Blocking.** `filterNotNull()` here swallows *all* `selectedAccount` nulls, not just the transient rebuild gap this PR targets — `AccountDataSource` also emits null on `resetSynchronizer()` (resync confirm, sync-error recovery, Keystone account creation), a legitimate non-transient event. Previously the `synchronizer == null || selectedAccount == null` guard at L312 caught that and fell back to the loading spinner; now it never re-fires, so Send can keep showing a stale/deleted account indefinitely.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
```diff
@@ -109,13 +115,36 @@ class SynchronizerProviderImpl(
...
+    override val isLocalBalanceSnapshotLoaded: StateFlow<Boolean> = mutableIsLocalBalanceSnapshotLoaded
+
     override val walletBalances: Flow<Map<AccountUuid, AccountBalance>?> =
         synchronizer
             .flatMapLatest { it?.walletBalances ?: flowOf(null) }
-            .onEach { if (it != null) lastKnownWalletBalances.value = it }
             .map { it ?: lastKnownWalletBalances.value }
 
     init {
+        scope.launch {
```
noop-sk: This reset (from `persistableWallet`) and the balance collector below (L137-145, from `synchronizer.walletBalances`) are two independently-scheduled coroutines over two different upstream sources, racing to write the same mutable state. On logout/wallet-switch, a trailing non-null emission from the closing synchronizer's `walletBalances` can land after this reset and flip `mutableIsLocalBalanceSnapshotLoaded` back to `true` with the old wallet's stale balances — `AutomaticServerRepository`, which now gates on this flag, could act on that stale signal for the wrong wallet.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/send/AndroidSend.kt
```diff
@@ -136,6 +148,19 @@ internal fun WrapSend(
 
     val recipientAddressState by viewModel.recipientAddressState.collectAsStateWithLifecycle()
 
+    // Address validation is the only form behavior that needs the live Synchronizer. Revalidate
+    // after replacement; the local account snapshot can remain visible during the short gap.
+    LaunchedEffect(synchronizer) {
```
noop-sk: This captures `recipientAddressState.address` and awaits the suspend `validateAddress` call. If the user types further characters before it resolves, the live `onRecipientAddressChange` path (L339) already pushed the newer address/type to the ViewModel; this effect's completion then overwrites it with the stale pre-edit pair — the address field can visibly revert right after a server switch replaces the Synchronizer.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/repository/AutomaticServerRepository.kt
```diff
@@ -85,13 +89,18 @@ class AutomaticServerRepositoryImpl(
         isServerAutomatic
             .flatMapLatest { isAutomatic ->
                 if (isAutomatic) {
-                    walletRepository.fastestEndpoints
-                        .filter { !it.isLoading }
-                        .mapNotNull { it.servers?.firstOrNull() }
+                    combine(
```
noop-sk, line 110: No `distinctUntilChanged()` on this combine's output. If `isLocalBalanceSnapshotLoaded` flips again after a wallet switch (see the SynchronizerProvider race noted separately) while `fastestEndpoints` is unchanged, this re-emits the same candidate and re-triggers `updateWalletEndpoint()` for an endpoint that hasn't actually changed — a redundant rebuild, working against the goal of this PR.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
```diff
@@ -109,13 +115,36 @@ class SynchronizerProviderImpl(
...
+            persistableWalletProvider.persistableWallet.collect { wallet ->
+                if (wallet == null) {
+                    lastKnownWalletBalances.value = null
+                    mutableIsLocalBalanceSnapshotLoaded.value = false
+                }
+            }
+        }
+
+        scope.launch {
```
noop-sk: This duplicates the `flatMapLatest { it?.walletBalances }` subscription already set up for the public `walletBalances` property (L122-125) instead of sharing it — every synchronizer rebuild now sets up the SDK's balance flow twice concurrently for no behavioral benefit.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
```diff
@@ -48,6 +47,13 @@ interface SynchronizerProvider {
      */
     val walletBalances: Flow<Map<AccountUuid, AccountBalance>?>
 
+    /**
+     * True after the SDK has completed at least one database-backed balance read for the current
+     * wallet. An empty map is still a completed read (for example, for a newly created wallet).
+     */
+    val isLocalBalanceSnapshotLoaded: Flow<Boolean>
```
noop-sk: Minor: this default (`flowOf(false)`) is a cold, single-emission, *completing* flow, not a held-open `false`. Any future `SynchronizerProvider` implementation/test fake that doesn't override it and gets consumed inside a `combine(...)` (as `AutomaticServerRepositoryImpl.init()` does) would see the whole combined flow complete after one emission instead of staying at `false`. Latent today since the only prod implementation overrides it, but worth a doc note or `MutableStateFlow`-backed default.

noop-sk (dismissed):
Two correctness issues block merge (send-button/destination-resolution gap, and account-loss guard silently disabled); five more are lower-severity races/inefficiencies. Details inline.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
zcash-harry: Valid. I removed the separate loaded flag and both competing state writers. The automatic gate now reads the existing walletBalances flow directly, so there is no independent reset collector racing a balance collector.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/send/AndroidSend.kt
zcash-harry: Valid. I removed the screen-level filterNotNull account workaround. Account state now follows a provider-owned retained synchronizer only during rebuilds, and retained state clears when the wallet is removed. A new synchronizer replaces it immediately.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
zcash-harry: Valid. The isLocalBalanceSnapshotLoaded interface property and its completing default were removed. The repository now gates directly on the existing walletBalances contract.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/repository/AutomaticServerRepository.kt
zcash-harry: Valid. The non-null candidate stream now uses distinctUntilChanged, so balance updates or rebuild-state changes cannot persist the same endpoint repeatedly.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/provider/SynchronizerProvider.kt
zcash-harry: Valid. The extra provider balance collector was removed. This version does not create a second internal subscription to maintain a separate loaded flag.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/send/AndroidSend.kt
zcash-harry: Valid. I removed the nullable Send validation path. Send now receives the retained UI synchronizer during replacement, so the existing non-null address validation and disabled-until-valid behavior remain intact.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/send/AndroidSend.kt
zcash-harry: Valid. I removed the revalidation LaunchedEffect. Address validation remains on the normal input path and uses the retained UI synchronizer until its replacement is available, so no stale captured address can overwrite newer input.

nesence-m (dismissed):
utACK

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/repository/AutomaticServerRepository.kt
nesence-m: Heads-up — the current head reverses this again: distinctUntilChanged was removed in 1f0cd834d. Deduping after filterNotNull permanently suppressed a winning candidate that had been skipped while a transaction was in flight (nulls never reset the dedupe), so the switch was never retried. The redundant-rebuild concern is covered one layer down instead: updateWalletEndpoint already no-ops when the endpoint is unchanged. Rationale documented in the KDoc on init().

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetBalancePoolsUseCase.kt
```diff
@@ -40,35 +41,34 @@ import kotlinx.coroutines.flow.map
...
-            accountDataSource.selectedAccount
-        ) { synchronizer, account ->
-            synchronizer to account
-        }.flatMapLatest { (synchronizer, account) ->
-            if (synchronizer == null || account == null) {
-                flowOf(null)
+            accountDataSource.selectedAccount,
+            synchronizerProvider.synchronizer.flatMapLatest { it?.walletBalances ?: flowOf(null) },
+        ) { account, balances ->
+            if (account == null || balances == null) {
```
nesence-m: lets make this condition simpler - if balances[account.sdkAccount.accountUuid] is null then we fallback to null too.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/WalletAccount.kt
```diff
@@ -16,12 +16,15 @@ sealed interface WalletAccount : Comparable<WalletAccount> {
     val unified: UnifiedInfo
```
nesence-m: lets make WalletAccount more tidy than before - lets have properties saplingBalance, orchardBalance, ironwoodBalance and their  unifiedBalance, unifiedAddr. Then lets have also transparentBalance and transparentAddr.

nesence-m (changes requested):

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/WalletAccount.kt
nesence-m: Fixed in cc3396195 — UnifiedInfo/TransparentInfo/SaplingInfo are gone; WalletAccount now has flat unifiedBalance (still the Orchard+Ironwood fold), orchardBalance (new, Orchard-only), saplingBalance, ironwoodBalance, transparentBalance, plus unifiedAddress/transparentAddress/saplingAddress. Went with the full-word *Address names over *Addr — say the word if you want the short form.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetBalancePoolsUseCase.kt
nesence-m: Fixed in cc3396195 — a missing account entry now emits null (loading) instead of a zero-filled BalancePools. Verified every AccountBalance field is non-null (SDK source + resolved jar), so once the entry is found nothing needs defaulting; all the ?: Zatoshi.ZERO noise is gone.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
```diff
@@ -106,32 +111,69 @@ class AccountDataSourceImpl(
...
-                                    observeUnified(synchronizer, sdkAccount),
-                                    observeTransparent(synchronizer, sdkAccount),
-                                    observeSapling(synchronizer, sdkAccount),
-                                    observeIronwoodBalance(sdkAccount),
-                                    observeIsSelected(sdkAccount, allSdkAccounts),
-                                ) { unified, transparent, sapling, ironwoodBalance, isSelected ->
+                                val unifiedAndTransparent =
+                                    combine(
+                                        observeUnifiedAddress(synchronizer, sdkAccount),
+                                        observeUnifiedBalance(synchronizer, sdkAccount),
```
nesence-m: unified balance should be calculated inside WalletAccount itself. No need for this function

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
```diff
@@ -106,32 +111,69 @@ class AccountDataSourceImpl(
...
+                                        ->
+                                        UnifiedAndTransparentSnapshot(
+                                            unifiedAddress = unifiedAddress,
+                                            unifiedBalance = unifiedBalance,
+                                            orchardBalance = orchardBalance,
+                                            transparentAddress = transparentAddress,
+                                            transparentBalance = transparentBalance,
+                                        )
+                                    }
+                                val saplingAndSelection =
```
nesence-m: unifiedAndTransparent and saplingAndSelection should not be 2 flows - have all this logic inside 1 combine or similar function

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
```diff
@@ -377,13 +433,29 @@ private data class AddressRequest(
     val responseChannel: Channel<WalletAddress.Unified>
 )
 
+private data class UnifiedAndTransparentSnapshot(
```
nesence-m: having 2 classes UnifiedAndTransparentSnapshot and SaplingAndSelectionSnapshot is messy - we either dont need that or we need only 1 union of these or smth

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/WalletAccount.kt
```diff
@@ -13,16 +13,35 @@ import co.electriccoin.zcash.ui.design.util.stringRes
...
-    val sapling: SaplingInfo?
-
-    // Ironwood shares the same unified address as Orchard (no address of its own), so this is a
-    // plain balance, not an address-carrying *Info like sapling/transparent/unified. Note that
-    // `unified`'s *balance* folds Orchard + Ironwood together (see observeUnified in
-    // AccountDataSource) — callers that need the Orchard-only balance should use
-    // GetOrchardBalanceUseCase, not this field or `unified`.
-    val ironwoodBalance: WalletBalance
-    val transparent: TransparentInfo
+    val unifiedAddress: WalletAddress.Unified
```
nesence-m: these addresses can be of type String. No need for these fancy WalletAddress.whatever types

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/WalletAccount.kt
```diff
@@ -76,15 +104,24 @@ sealed interface WalletAccount : Comparable<WalletAccount> {
...
 }
 
 data class ZashiAccount(
     override val sdkAccount: Account,
-    override val unified: UnifiedInfo, // TODO [#26]: technical debt, aggregates ORCHARD + IRONWOOD, sync with iOS
-    override val sapling: SaplingInfo,
-    override val ironwoodBalance: WalletBalance,
-    override val transparent: TransparentInfo,
+    override val unifiedAddress: WalletAddress.Unified,
+    override val unifiedBalance: WalletBalance?,
```
nesence-m: unified balance should be calculated inside. Not a constructor param

nesence-m (changes requested):

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
nesence-m: Fixed in 430b62b81 — observeUnifiedBalance is deleted; unifiedBalance is now a derived property on WalletAccount computed from orchardBalance + ironwoodBalance (the relocated plus operator also folds locked now, which the old fold silently dropped).

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/WalletAccount.kt
nesence-m: Fixed in 430b62b81 — unifiedBalance left both constructors; it is a get() on the interface.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
nesence-m: Fixed in 430b62b81 — the account is built from one 5-arg combine: the account's AccountBalance entry, the two addresses, the sapling address, and isSelected. All per-pool balances read off the single entry inside the transform.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/datasource/AccountDataSource.kt
nesence-m: Fixed in 430b62b81 — both snapshot classes are gone; nothing replaced them (the single combine made them unnecessary).

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/model/WalletAccount.kt
nesence-m: Fixed in 430b62b81 — addresses are plain Strings end to end (the WalletAddress wrappers carry no validation — new() is a no-op stub pending SDK #342). The one remaining typed consumer, fromReceiveAddressType for the QR/request flows, constructs the typed value locally. requestNextShieldedAddress also returns String now.

nesence-m (approved):
Reviewed through the full rework: data-level retention (account-keyed), null-while-loading end to end, flattened WalletAccount, single-combine account pipeline. Deep review findings all addressed except the two explicitly deferred to MOB-1832 (balance-gate failover for never-synced wallets, invoke() timeout hardening).

PR #2490, opened 2026-08-31
MOB-1744: report pre-submission failures accurately instead of a fake gRPC status

noop-sk:
## Summary

- Support/error reports for `SubmitResult.Error` (built when a transaction submission fails **before** any gRPC call is made -- e.g. proposal/PCZT missing, or a proving/signing failure such as a Sapling parameter download timeout) previously hardcoded `gRPC: false, code: -1` in the report, using the same template as a genuine gRPC/mempool status. This made a non-gRPC failure look like a real protocol error to whoever reads the support report.
- `SendEmailUseCase.invoke(SubmitResult.Error)` now reports these as a distinct pre-submission-failure category, carrying the real exception type and detail (via a new `submitErrorPreSubmissionDetail()` helper and a new `send_confirmation_report_status_presubmission_failure` string resource), instead of a fake gRPC status.
- Traced how `SubmitResult.Error` is constructed (`ZashiProposalRepository.submit()`, `KeystoneProposalRepository.submit()`, `ProposalDataSource.submitTransactionInternal()`): the real exception (`cause: Exception`) is already preserved end to end and never generically wrapped away, so no upstream propagation fix was needed -- only the reporting/formatting step in `SendEmailUseCase` was wrong.
- `SubmitResult.Failure`'s existing report (which is a genuine, non-gRPC mempool rejection with a real `code`/`description` from the server) is unchanged, extracted into a small `submitFailureReportStatus()` helper for testability, with a comment explaining why its `index=0, grpcError=false` values are real and not placeholders -- to avoid this being mistaken for the same bug in the future.
- Added a CHANGELOG entry.

Companion SDK-side PR (separate repo, not part of this PR): adds retry/backoff to the actual Sapling-params download that causes this bug's user-visible symptom (worktree `z/wt/mob1744-sdk-retry`). This PR is purely about making the app's error reporting accurate, independent of whether that fix has landed.

## Test plan

- [x] `./gradlew :ui-lib:testZcashmainnetInternalDebugUnitTest --tests "co.electriccoin.zcash.ui.common.usecase.SendEmailUseCaseTest"` -- new tests cover a genuine gRPC/mempool failure still reporting its real code, and a pre-submission failure (e.g. an `IOException` from a param download) reporting its real exception type/detail distinctly from a fake gRPC status.
- [x] `./gradlew :ui-lib:testZcashmainnetInternalDebugUnitTest` (full module) -- all pass.
- [x] `./gradlew ktlintFormat` -- no changes needed.
- [x] `./gradlew detektAll` -- clean.

Linear: https://linear.app/zodl/issue/MOB-1744/timeouts-while-on-android-v393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 ### Fixed:
 
 - Coinholder Polling's poll list and proposal loading no longer hang indefinitely when a vote server is unreachable over Tor; a slow or dropped connection now fails over to the next server within a bounded time.
+- Support/error reports for a failed transaction no longer claim a fake `gRPC: false, code: -1` status when the failure actually happened before submission (e.g. a Sapling parameter download failure); such reports now correctly identify the real exception instead.
 
 ## [3.10.1 (2512)] - 2026-08-25
 
--- a/ui-design-lib/src/main/res/ui/common/values-es/strings.xml
+++ b/ui-design-lib/src/main/res/ui/common/values-es/strings.xml
@@ -698,6 +698,11 @@
         código: <xliff:g example="123" id="code">%3$d</xliff:g>,
         descripción: <xliff:g example="Network unreachable" id="desc">%4$s</xliff:g>
     </string>
+    <string name="send_confirmation_report_status_presubmission_failure" formatted="true">Fallo antes del envío
+        de la transacción (no se realizó ninguna llamada gRPC):
+        <xliff:g example="IOException" id="type">%1$s</xliff:g>:
+        <xliff:g example="Failed to download Sapling parameters" id="detail">%2$s</xliff:g>
+    </string>
     <string name="send_confirmation_multiple_trx_failure_report_unable_open_email">No se pudo abrir la app de correo.</string>
     <string name="send_confirmation_multiple_trx_failure_copy_tag">IDs de Transacción</string>
     <string name="swapAndPay_successSwapInfo">Iniciaste un intercambio exitosamente.\nSigue su estatus en la pantalla de transacciones.</string>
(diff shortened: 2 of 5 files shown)


nesence-m (approved):
Traced the classification end to end at head: `SubmitResult.Error` is only constructed from the repositories' catches around `proposalDataSource.submitTransaction(...)`, and the broadcast phase can't leak an exception there — `MultiEndpointTransactionSubmitter.submitToEndpoint` converts every broadcast exception into `TransactionSubmitResult.Failure(grpcError=true)`, so a gRPC-time IOException/timeout correctly lands in the `GrpcFailure`/`Failure` categories, not in the new pre-submission one. Localization checked: both existing locales (`values`, `values-es`) got the new key with matching placeholders. No in-repo consumer parses the old `gRPC: false, code: -1` format (only the strings themselves and the MOB-1430 audit CSV reference it). CHANGELOG and base branch (`maint/v3.10.x`) are right.

Two non-blocking notes:

1. `submitTransactionInternal` calls `refreshTransactions()`/`refreshAllBalances()` *after* the broadcast; if either throws, the repositories' catch-alls turn it into `SubmitResult.Error`, and the report now affirmatively says "no gRPC call was made" for a transaction that may already have been broadcast. Pre-existing mislabeling (the old fake-gRPC report was equally wrong there), but since the new wording makes a claim support may act on, consider wrapping those two best-effort refresh calls in `runCatching` here or in a follow-up — that would make the KDoc exactly true.
2. Nit: `exceptionType` duplicates the first token of the stack-trace detail ("IOException: java.io.IOException: …"). Fine as a grep-friendly triage field; drop it if you prefer a cleaner report.

CI: the `test_android_modules_emulator (app)` and `Keystone` failures are emulator-infra flakes (emulator console never started / adb daemon unreachable, run truncated mid-suite on surfaces this diff doesn't touch) — please re-run them to green before merging.

[#2490 MOB-1744: report pre-submission failures accurately instead of a fake gRPC status](https://github.com/zodl-inc/zodl-android/pull/2490)

PR #2491, opened 2026-08-31
MOB-1787: make backup prompt outrank migration/error home widgets

noop-sk:
## Summary

Ticket: https://linear.app/zodl/issue/MOB-1787/seed-backup-prompt-should-take-priority-over-migration-and-error-state

Three users lost funds because they never backed up their recovery phrase, and the backup
prompt currently sits below the migration widget and several error states on the home screen —
it could take days before a user even sees it. This PR makes the backup prompt outrank every
other home-screen widget (migration banner, sync/connectivity errors, etc.) whenever the wallet
has received any nonzero balance and the seed has not been backed up (no threshold).

**Root cause:** `HomeMessageData.RuntimeMessage` (migration, disconnected, sync error, syncing,
etc.) always carries `priority = Int.MAX_VALUE` and was checked *before* `Backup` in both
`GetHomeMessageUseCase.createMessage()`'s ordering and `prioritizeMessage()`'s hysteresis gate.
Backup's finite priority (5) meant it could never win against a RuntimeMessage no matter how
urgent the underlying condition.

**Fix:**
- `createMessage`/`createHomeMessage` (extracted to a top-level `internal` function, mirroring
  the existing `syncingMessageFor` pattern, for direct unit testing): check
  `backup is WalletBackupData.Available` before `runtimeMessage != null` instead of after. Every
  occurrence of `HomeMessageData.Backup` is already gated by
  `WalletBackupMessageUseCaseImpl.observe()` on "balance received + not backed up" (MOB-909's
  fix, untouched), so no separate "urgent" flag is needed.
- `prioritizeMessage`/`prioritizeHomeMessage` (same extraction): let an urgent `Backup` message
  bypass the same-priority/hysteresis suppression that `RuntimeMessage` already bypasses —
  otherwise a `RuntimeMessage` cached as `lastShownMessage` (priority `Int.MAX_VALUE`) would make
  Backup's finite priority look "lower" and get silently filtered out.
- Normal relative ordering of `RuntimeMessage` internals (disconnected → error → syncing →
  migration → shield funds) and the rest of the `Prioritized` family (CoinholderPolling, Tor,
  currency conversion, crash report) is unchanged when the urgent-backup condition is false.

## Scope

- **Logic/ordering only — no visual or copy changes.** The Figma link originally attached to the
  ticket resolves to an unrelated "Swap details didn't match" iOS mockup in the shared file, not
  this feature. The existing `WalletBackupMessage.kt` widget visuals are reused as-is. New
  visual/copy work should follow once a corrected Figma link/node is confirmed.
- **Android only** — iOS is tracked separately per the ticket.

## Open product question — needs a decision

`WalletBackupMessageUseCaseImpl` layers a remind-me snooze/lockout (2 days → 2 weeks → 1 month)
on top of the balance/backup condition. This PR does **not** change that mechanism — a user who
tapped "remind me later" can still have the urgent prompt suppressed during the snooze window
even after a balance arrives, since `WalletBackupData.Available` (the trigger this PR keys off)
is only emitted once the snooze window has elapsed. The ticket's "no threshold" wording doesn't
address the snooze mechanism directly, so this is left as-is pending an explicit product
decision on whether the urgent case should bypass snooze. Flagging here rather than guessing.

## Testing

- New `GetHomeMessageUseCaseBackupPriorityTest` (`ui-lib/src/test/.../usecase/`): backup wins
  over an active migration message, backup wins over an active sync error message, balance=0
  regression guard (MOB-909), already-backed-up regression guard, normal `Prioritized` ordering
  preserved when no runtime message/urgent backup, plus two `prioritizeHomeMessage` cases
  confirming the hysteresis bypass.
- `./gradlew :ui-lib:testZcashmainnetInternalDebugUnitTest :feature-migration:testZcashmainnetInternalDebugUnitTest` — all green, no regressions.
- `./gradlew ktlint` and `./gradlew detektAll` — clean, zero findings.
- Built `assembleZcashmainnetInternalDebug`, installed and launched on a local emulator — app
  boots to onboarding without crashing. Full live reproduction of the compound scenario (a
  synced wallet with a real received balance, unbacked seed, and a simultaneously active
  migration/error banner) needs a pre-existing funded test wallet that wasn't available in this
  session, so it was not exercised end-to-end live; the exact decision logic is covered directly
  by the unit tests above instead.

## Test plan

- [ ] On a synced wallet with a received balance and an unbacked seed, trigger a migration or
      sync-error banner and confirm the backup prompt now wins
- [ ] Confirm normal migration/error banners still show as before once the seed is backed up (or
      balance is zero)
- [ ] Product decision recorded on the snooze-bypass question above

https://claude.ai/code/session_01KX1k7Wox8bhSDz47hDrVqd

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+### Changed:
+
+- The seed backup prompt now always takes priority over the migration banner and any sync/connectivity error banner once your wallet has received a balance and you haven't backed up your recovery phrase yet, so it can no longer be hidden behind other home screen messages for days.
+
 ## [3.11.0 (2653)] - 2026-09-08
 
 ### Added:
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetHomeMessageUseCase.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetHomeMessageUseCase.kt
@@ -68,7 +68,7 @@ class GetHomeMessageUseCase(
                         TrailingPrioritizedInputs(isCoinholderPollingVisible, isCrashReportingVisible)
                     },
                 ) { runtimeMessage, backup, isTorAvailable, isCCAvailable, trailing ->
-                    createMessage(
+                    createHomeMessage(
                         runtimeMessage = runtimeMessage,
                         backup = backup,
                         isTorVisible = isTorAvailable,
@@ -201,57 +201,13 @@ class GetHomeMessageUseCase(
     private fun observeIsTorMessageVisible() =
         isTorEnabledStorageProvider.observe().map { it == null }.distinctUntilChanged()
 
-    private fun createMessage(
-        runtimeMessage: RuntimeMessage?,
-        backup: WalletBackupData,
-        isTorVisible: Boolean,
-        isCurrencyConversionEnabled: Boolean,
-        isCoinholderPollingVisible: Boolean,
-        isCrashReportingVisible: Boolean,
-    ) = when {
-        runtimeMessage != null -> runtimeMessage
-        backup is WalletBackupData.Available -> HomeMessageData.Backup
-        isCoinholderPollingVisible -> HomeMessageData.CoinholderPolling
-        isTorVisible -> HomeMessageData.EnableTor
-        isCurrencyConversionEnabled -> HomeMessageData.EnableCurrencyConversion
-        isCrashReportingVisible -> HomeMessageData.CrashReport
-        else -> null
-    }
-
     private fun prioritizeMessage(message: HomeMessageData?): HomeMessageData? {
-        val isSameMessageUpdate =
-            message?.priority == cache.lastMessage?.priority // same but updated
-        val someMessageBeenShown =
-            cache.lastShownMessage != null // has any message been shown while app in fg
-        val hasNoMessageBeenShownLately = cache.lastMessage == null // has no message been shown
-        val isHigherPriorityMessage =
-            (message?.priority ?: 0) > (cache.lastShownMessage?.priority ?: 0)
         val result =
-            when {
-                message == null -> {
-                    null
-                }
-
-                message is RuntimeMessage -> {
-                    message
-                }
-
-                isSameMessageUpdate -> {
-                    message
-                }
-
-                isHigherPriorityMessage -> {
-                    if (hasNoMessageBeenShownLately) {
-                        if (someMessageBeenShown) null else message
-                    } else {
-                        message
-                    }
-                }
-
-                else -> {
-                    null
-                }
-            }
+            prioritizeHomeMessage(
+                message = message,
+                lastMessage = cache.lastMessage,
+                lastShownMessage = cache.lastShownMessage,
+            )
 
         if (result != null) {
             messageAvailabilityDataSource.onMessageShown()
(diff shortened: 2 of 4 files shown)


nesence-m (commented):
Verified the fix against the head revision — the logic is correct and the extraction is faithful:

- The urgent condition (`WalletBackupData.Available`) is gated in `WalletBackupMessageUseCaseImpl` purely on "any `ReceiveTransaction` exists" (Success/Pending/Failed) for the selected Zashi account — no amount threshold, no spendable/confirmed filtering, no Zatoshi math, matching the ticket's "any nonzero received balance, no threshold" (and erring safe beyond it).
- Nothing is permanently hidden by the reordering: the backup widget has no dismiss affordance, and both of its exits (backup completed, remind-me-later) immediately return the migration/error banner. Restored wallets set the backup flag at restore time, so the Restoring banner is never masked for imports.
- The `prioritizeHomeMessage` Backup bypass is necessary, not incidental — without it a cached `RuntimeMessage` (`Int.MAX_VALUE`) as `lastShownMessage` would filter Backup back out, and the new tests cover exactly that case.
- Snooze behavior confirmed as described: `Available` only emits after the lockout elapses, so a snoozed user with newly arrived funds keeps the old ordering until the window expires. Agreed this is a product call, not a code question — would be good to get it decided before release, since it partially reopens the MOB-1787 window (up to 30 days for a thrice-snoozed user).

Remaining items:

1. **CI**: `test_android_modules_emulator (app)` is red on the current head (ScreenshotTest new-wallet flow, run died mid-suite with emulator ColorBuffer/console errors — https://github.com/zodl-inc/zodl-android/actions/runs/33390662160/job/99812343975). The failing flow can't reach the changed branch (fresh wallet has no receives), so this looks like flake, but it needs a green re-run before merge.
2. **Non-blocking**: the "zero balance regression guard" and "already backed up seed" tests feed identical inputs (`Unavailable`) to `createHomeMessage` — the distinction they name lives in `WalletBackupMessageUseCaseImpl`, which still has no direct unit coverage. Since the whole fix keys off that gate, a small test over `observe()` (receive-tx gate + snooze chain) would be a worthwhile follow-up; at minimum consider renaming the two tests.
3. **Nit**: the CHANGELOG entry is under "Changed" — given this is the MOB-1787 fund-loss fix, "Fixed" might be the more discoverable section.

Nothing blocking in the code itself — once CI is green (and assuming no product objection on the migration/shield-funds masking noted above), this is good to go.

noop-sk (commented):
Note: this PR's base commit is authored by my own account, so I can't formally APPROVE or REQUEST_CHANGES here (self-review) — posting findings as a comment for a human reviewer to act on.

Follow-up commit (`a0df40b6`, line-comments → KDoc) confirmed comment-only as claimed — two local-variable renames for clarity plus doc conversion, zero logic/behavior change.

Priority-ordering logic itself checks out: making the backup prompt outrank `RuntimeMessage`s is narrowly scoped to the specific already-conservative trigger condition (balance received + seed unbacked + snooze elapsed), it isn't permanent (the underlying widget reappears once backup completes/is re-snoozed), and critical sync errors have a separate dedicated surfacing path independent of this home-banner system. Reasonable, matches the PR's stated fund-loss-prevention intent.

**Blocking — still open:** the product question from the earlier review round is unanswered. `WalletBackupData.Available` (the trigger this fix keys off) only fires after the remind-me-later snooze window elapses, so a thrice-snoozed user can still have the backup prompt suppressed for up to 30 days even after a balance arrives — partially reopening the exact fund-loss window this PR is meant to close. Checked both the inline PR comments and the review thread — no answer appears anywhere. Needs an explicit product sign-off before merge, not a silent pass.

Non-blocking:
1. The specific CI ask ("green re-run of `test_android_modules_emulator`") hasn't literally happened — it's still FAILURE on this head, but the failure (Restore Wallet screenshot timeout) is unrelated to this diff's home-message code, consistent with the other two red jobs (Keystone, e2e phase 2) also reading as infra flake.
2. New KDoc says "Backup's finite priority (5)" in two places (`GetHomeMessageUseCase.kt:292`, test file `:106`) but the actual value in `HomeMessageCacheRepository.kt:87` is `priority = 4`. Doc-only, no functional impact, but worth fixing since the follow-up commit specifically exists to make this doc accurate.
3 & 4. Carried over from the earlier review round, still unaddressed: the two "priority" tests both feed identical `WalletBackupData.Unavailable` input rather than distinguishing the case they're named for; and the CHANGELOG entry sits under "Changed" rather than "Fixed" for a fund-loss fix.

Given the open product question, not adding a ready-to-approve line here.

noop-sk (commented):
Re-verified against the current head, independently (not taking the claim at face value).

Item 2 (KDoc priority mismatch) — my original finding was wrong, not yours. Re-checked the live file: `HomeMessageCacheRepository.kt:87` is `Backup` at priority 5, `:108` is `CoinholderPolling` at priority 4 — exactly as you said. The KDoc's "Backup's finite priority (5)" was correct from the very first commit; I mis-attributed the numbers to the wrong entity in the earlier pass. Thanks for catching it — correctly left alone.

Item 3 — FIXED. New `WalletBackupMessageUseCaseImplTest.kt` (5 tests) now directly exercises `observe()` with genuinely distinct mock setups per case, and the two carried-over tests in `GetHomeMessageUseCaseBackupPriorityTest.kt` were renamed to describe what they actually assert.

Item 4 — FIXED. CHANGELOG entry now sits under `### Fixed:`, no stray duplicate.

The `maint/v3.11.x` merge (CHANGELOG conflict from #2509) resolved cleanly, nothing new introduced.

Still can't formally approve — self-authored, same as last round.

Item 1 (snooze-bypass product question) is still open, confirmed via your own PR comment explicitly deferring it as a product decision. That comment also adds a regression-pinning test so a future decision has somewhere to land, which is a reasonable way to leave it — but the actual call (and the merge) still needs a human.

nesence-m (approved):
utACK

noop-sk (commented):
Noting for the record, not blocking further action on my end (still can't formally approve — self-authored).

Item 2 re-confirmed unchanged at the current head — `HomeMessageCacheRepository.kt` still has Backup=priority 5, CoinholderPolling=priority 4, file isn't even touched by the latest commits. Solid.

On the product-decision comment: I want to flag, transparently, that both the "deliberately untouched, that's a product decision not a code fix" comment and the follow-up "decided, keep the behaviour as is" comment came from the same account, which then also submitted the formal APPROVE review. So procedurally this PR now has an approval from someone other than the author, which is what was needed — but the product call itself doesn't show any input from an independent product owner/PM, just the same reviewer resolving the question he raised. Not asserting that's wrong for this org's process, just noting it's self-referential rather than independently sourced, in case that distinction matters here.

Also FYI, unrelated to the above: `Keystone` and `test_android_modules_emulator (app)` are currently red on this head — consistent with the same emulator/adb infra flake pattern seen on every other PR in this review round, not something I'd block on, but flagging in case it's worth a rerun before merge.

PR #2494, opened 2026-09-03
MOB-1340: Show "Swap details didn't match" sheet with Report email on quote validation failures

nesence-m:
## Summary

Android user-facing surface for the NEAR swap-quote validation hardening from #2299 ([MOB-1340](https://linear.app/zodl/issue/MOB-1340/security-validate-every-field-of-the-near-intents-swap-quote-trust-the)).

Until now every request-vs-response validator threw a bare `IllegalArgumentException`, so a rejected quote fell into the generic "Quote Unavailable" sheet with no way to report it. Per the product decision on PRO-343 (generic sheet, no technical detail, Report button that bundles the data into an email):

- Every quote-vs-request check (recipient/refund address, origin/destination asset, swap type, slippage-tolerance echo, amountIn/amountOut raw-vs-formatted, requested amount, server min-amount outside slippage, non-positive amount) now throws a typed `SwapQuoteMismatchException` carrying which check failed and the quote's deposit address. It still extends `IllegalArgumentException`, so the fail-closed behaviour is unchanged.
- `RequestSwapQuoteUseCase` routes such a rejection to a new **"Swap details didn't match"** sheet (`screen/swap/mismatch/`, `dialogComposable`) instead of `SwapQuoteArgs`, mirroring the `InsufficientFundsArgs` routing.
- **Go back** dismisses; **Report** composes a support email "Zodl Swap: API mismatch" listing swap provider, swap type (e.g. `CrossPay - ZEC > USDC (Arbitrum)`), mismatch type and quote id (= deposit address), plus the usual device block, then dismisses.
- The sanitized crash-monitoring signal introduced for MOB-1371 now fires once from the repository for every mismatch type (type only, never amounts).
- A mismatch that lands while the cancel sheet is open stays in the repository's quote state and is routed to the mismatch sheet on dismissal, like every other quote error.
- Follow-up commits: report labels moved to string resources (review comment), and a deep-review pass (context carried on the exception in place, explicit mismatch types at every call site, stale local asset selection kept on the generic path, `StringResource`-composed email labels, no `requestId`).
- EN + ES strings, changelog entry.

Design: Figma [Dev Ready Q3 2026 → Swaps: Validation Error Handling](https://www.figma.com/design/Y6OIQ6mHVJ5e69cHoVkfCF/Dev-Ready-Q3-2026?node-id=742-2796) (light + dark). Note: iOS has not shipped this yet (MOB-1405 is still open), so this is built from the spec, not ported.

## Test plan

- [x] `ui-lib` unit suite (474 tests) green, built against the pinned SDK source (`SDK_COMMIT_PIN`): new `SwapQuoteMismatchVMTest`; extended `RequestSwapQuoteUseCaseTest` (mismatch → `SwapQuoteMismatchArgs`, quote cleared, `SwapQuoteArgs` not forwarded), `NearSwapQuoteValidationTest`, `SwapRepositoryImplTest`, `NearSwapDataSourceTest`, `SendEmailUseCaseTest`.
- [x] `ktlint`, `detektAll`, `:ui-design-lib:lintDebug` (MissingTranslation) clean.
- [x] `assembleZcashmainnetStoreDebug` + APK native-lib check.
- [x] Emulator (Pixel 9 Pro XL), with a temporary local hack forcing the swap-type check to fail on a Swap-into-ZEC quote: sheet renders per design in dark and light; Go back returns to the form with the quote cleared; Report opens the mail intent with the expected subject/body (verified via the decoded `mailto:` payload, quote id = real NEAR deposit address). Hack reverted before commit. Re-run after the review fixes with the same result (mailto payload identical, crash signal logged from the repository).
- [ ] Spanish copy is a fresh translation, not yet reviewed by a translator.

https://claude.ai/code/session_01JqbL85ACny1iQxbUW87Mwv


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/SendEmailUseCase.kt
```diff
@@ -302,3 +336,45 @@ internal fun submitErrorPreSubmissionDetail(cause: Exception): SubmitErrorPreSub
...
+ * [chainName] resolves a chain ticker to its display name; the chain is omitted for ZEC, which has
+ * only one. English only: it is never shown in the UI.
+ */
+internal fun swapQuoteMismatchSwapTypeLabel(
+    report: SwapQuoteMismatchReport,
+    chainName: (String) -> String
+): String {
+    val mode =
+        when (report.mode) {
+            SwapMode.EXACT_INPUT -> "Swap"
```
nesence-m: use strings

nesence-m: Done in 12a1f2729: the mode names, all eleven mismatch-type labels, the provider name, the unknown-quote-id fallback and the two composed shapes are now string resources (EN + ES). `SwapQuoteMismatchType` carries `@StringRes reportLabelRes` (same idiom as `WalletBackupLockoutDuration`), and the helpers take a string resolver so the unit tests keep asserting the exact final strings.

LukasKorba (commented):
Second pass — both blocking items are genuinely fixed in code, not just in the reply. Nothing I found is a correctness regression a user can hit on the current UI. **LGTM with nits**, none of which need to block.

**Verified addressed**

- **Mutable exception fields** — `SwapQuoteMismatch.kt:44-49`, all four `var`s gone, `sealed class` with `val` constructor params (final fields ⇒ JMM-safe publication). `NearSwapDataSource.kt:179-184` rethrows a fresh instance with `cause = this` instead of writing to a thrown object. Report context is the immutable `SwapQuoteMismatch` on `SwapQuoteData.Error.mismatch`, assembled once in `SwapRepository.toMismatch:519-529`. Good shape.
- **Sealed hierarchy** — `SwapAmountInconsistencyException` deleted from `NearSwapQuote.kt`; I grepped head for stragglers (data-source import, catch site, tests), none remain. `SwapQuoteMismatchException` still extends `IllegalArgumentException`, and `checkSwapStatus` still declares `@Throws(..., IllegalArgumentException::class)`, so no catch site breaks.
- **ES string** — I checked the whole set rather than the one key: `values` and `values-es` each have exactly 25 `swap_mismatch_*` keys, names diff-clean, and the placeholders match (`%1$s..%4$s` in the email body, `%1$s - %2$s > %3$s`, `%1$s (%2$s)`).
- **`NEAR_SWAP_PROVIDER` fallback** — you kept it and said so, fine. Noting it's currently dead: every repository throw site runs after `receivedQuote = result` (`:248`, `:322`) and every data-source rejection already carries `provider = NEAR_SWAP_PROVIDER`, so it can only fire in the second-provider future the bot was imagining.

**Nits**

1. `SwapRepository.kt:230,:304` — the added `cancel()` isn't paired with a `CancellationException` rethrow. The job body's terminal `catch (e: Exception)` (`:289-291`, `:376-378`) swallows it and writes `SwapQuoteData.Error(mode, e)` into the shared flow; `cancel()` isn't `join()`, so the dying coroutine resumes after the new job has set `Loading`, and `RequestSwapQuoteUseCase.kt:122` — suspended in `quote.first { it !in listOf(null, Loading) }` — takes it as a real failure and routes to the generic error screen. Latent only, because `SwapVMMapper.kt:498` disables the button while `isRequestingQuote`, so there's never a live job to cancel. One line (`catch (e: CancellationException) { throw e }` above the generic catch), same for the pre-existing cancel in `clearQuote():442`.
2. `SendEmailUseCaseTest.kt` — the KDoc says the assertions are structural so "a copy edit in strings.xml cannot leave this green while changing the email", but `StringResource.ByResource` is `resource: Int` + `args`, and both sides use the same `R.string.*` constants, so a copy edit is exactly what these can't catch. 12a1f27 asserted rendered strings through a resolver map and the deep-review round traded that away while keeping the comment. Also `invoke(args: SwapQuoteMismatchArgs)` — the overload that builds the email — has no test at all, so the 4-placeholder contract of `swap_mismatch_support_email_body` is guarded only by lint.
3. `NearSwapQuoteValidationTest.requireConsistent_noOpWhenEitherValueNull` was deleted with no replacement; it covered the `if (raw == null || formatted == null) return` early-out. The same hunk dropped the `// endregion` + `// region requireWithinSlippage` header, so the ten slippage tests now sit under the `requireConsistent` banner.
4. `strings.xml:972` — "allows us raise it with them" (missing *to*) and "won't relay your personal details **with** the swap provider" (should be *to*). If that's verbatim from Figma, kick it back rather than patching here.
5. The whole report body is localized, mismatch type and mode labels included ("Tipo de discrepancia: Monto no positivo"). Support greps these; a stable identifier would triage better. Not urgent — the crash signal keeps `type=NON_POSITIVE_AMOUNT`.

**Checked and clear:** no NPE in the data-source catch (`QuoteDetails.depositAddress` is non-null, serialization fails first); no new PII class in the Report email — provider, mode, tickers, mismatch type, deposit address, device info, which is the same set as the existing `invoke(swapData: SwapData)` support email, and a mismatch fires before any deposit so the address is an unused provider-generated one. No seed, key, user address or amount anywhere.

One thing to be aware of rather than fix: the crash signal used to be AmountInconsistency-only from the data source and is now every mismatch type from both repository quote paths. Payload is sanitized (`type=`, `field=`, `decimals=`, never amounts), so it's intended — just a monitoring-volume change landing on a maint branch.

ui-lib/src/main/java/co/electriccoin/zcash/ui/common/repository/SwapRepository.kt
```diff
@@ -86,7 +92,13 @@ sealed interface SwapQuoteData {
...
     data class Error(
         val mode: SwapMode,
-        val exception: Exception
+        val exception: Exception,
+        /**
+         * The report context of a rejected quote, set when [exception] is a
+         * [SwapQuoteMismatchException]; null for every other quote failure, which stays on the
+         * generic quote-error path.
+         */
+        val mismatch: SwapQuoteMismatch? = null
```
nesence-m: its better to use mismatch exception and remove SwapQuoteMismatch class entirely. All the properties needed should be within that exception

nesence-m: Done in 3e072bfde: `SwapQuoteMismatch` is gone and the whole report context now lives on `SwapQuoteMismatchException` itself — `type`, `message`, `provider`, `depositAddress`, `originAsset` and `destinationAsset` are all constructor `val`s (nullable only where a throw site genuinely cannot know them), so nothing became mutable again: context is attached by constructing a new rejection chained to the original (`withQuoteContext` for what the provider's response says, `withReportContext` for the full context), `SwapQuoteData.Error` carries only the exception, the repository still owns the `NEAR_SWAP_PROVIDER` fallback, and `toMismatchArgs()` reads the sheet's arguments straight off the exception.

LukasKorba (commented):
The nits are done — but I'm not approving at head, because head isn't the diff I LGTM'd.

## `3e072bf`

My review said of the value model: *"Report context is the immutable `SwapQuoteMismatch` on `SwapQuoteData.Error.mismatch`, assembled once in `SwapRepository.toMismatch`. Good shape."* Forty-four minutes later `3e072bf` deleted that class and moved everything onto the exception, driven by a comment you left on your own PR at 14:24 rather than anything I asked for.

I'd rather you didn't restructure a shape a reviewer just signed off without flagging it — not because the new shape is wrong (it isn't; I read it), but because an approval on `af5e349` doesn't transfer to `3e072bf`, and if I'd merged on the strength of my earlier comment I'd have merged something I hadn't reviewed. If you want the change, say so and I'll look — it's a 30-second ask.

On the merits I'm fine with exception-carries-everything. The property that actually mattered in the original blocking finding — immutable, no `var`, nothing written to a thrown object, context attached by constructing a chained rejection — is fully preserved, and I found no stragglers: both consumers are consistent and no unused imports were left behind.

**One thing to fix before I approve.** `SwapQuoteMismatchException.kt:51-52` makes `originAsset`/`destinationAsset` nullable, where `SwapQuoteMismatch` had them non-null. So `SwapQuoteMismatchScreen.toMismatchArgs():38-56` now does a runtime null-check on `provider`/`originAsset`/`destinationAsset` and **silently returns null → generic error screen** when it fails. It's unreachable today, since `withQuoteReportContext:520-529` always routes through `withReportContext(provider: String, originAsset: SwapAsset, destinationAsset: SwapAsset)` with all three non-null — but that's a compile-time guarantee traded for a silent fail-open runtime branch, which is the same latency class as the cancel bug in nit 1: fine until it isn't, and invisible when it goes wrong. Either keep the reachable path non-null (a subclass that requires the report context would do it), or at minimum report it instead of silently falling through.

## The nits

**1 — good.** `SwapRepository.kt:287-288` and `:375-376`, both rethrows sit above the generic `catch (e: Exception)`, and the preceding `catch (e: SwapQuoteMismatchException)` is disjoint so it can't shadow them. You're right that `clearQuote():442` needed no separate fix — its `cancel()` resolves through those same two bodies. Noting the behavioural shape for the record: a cancelled job now writes nothing, so a caller suspended in `RequestSwapQuoteUseCase.kt:122` gets no terminal state where it previously got a bogus `Error` that at least unblocked it. Both cancel callers navigate away and tear the scope down, so that's correct — just no longer accidentally load-bearing.

**2 — the important half is fixed.** `swapQuoteMismatchReportBody()` extracted at `SendEmailUseCase.kt:353-366`, and `SendEmailUseCaseTest.kt:208-217` pins all four placeholders and their order against a full literal, plus the null-deposit-address `"unknown"` fallback. That's the gap that mattered.

The copy-edit claim is still not true though, and the new KDoc at `:140-142` and `:257-261` states it more strongly than before. `strings` at `:238-255` is a hand-maintained duplicate of strings.xml, not a read of it — edit `swap_mismatch_support_email_body` in the XML and the test renders from the stale map and stays green. I checked all ten entries against `strings.xml:972-990` and they match today, so nothing's broken; 12a1f27 had the same property. Just drop the claim from the KDoc rather than restating it.

**3 — good.** `NearSwapQuoteValidationTest.kt:79-94` restores the test with both null legs, and the region banners at `:96`/`:98` are back.

**4 — good**, and you were right to leave `values-es` alone: "con el proveedor" is correct Spanish. Kick the Figma copy back rather than carrying the deviation.

**5 — agreed, deferred.**

## CI

`e2e-android (phase 2)` is red only at head — cancelled on `6d6fb1f`, green on #2491. Annotations are "Post-swap-Confirm Sent! screen did not render in time (deposit tx still broadcast)" and a missing reverse-swap QR screenshot. Reads like a render-timeout flake and this PR only touches the validation-failure path, but it's in the swap flow with no green baseline on the parent, so please re-run it. `test_android_modules_emulator (app)` and `Keystone` are red on the commit I already LGTM'd and on #2497/#2491 too — pre-existing, ignoring those.

LukasKorba (commented):
Both items done, and the non-null fix is the right one — the distinction I was worried about is exactly the one you got right.

`SwapQuoteMismatchException.kt:58,61,64` drops the three fields to `open val … get() = null` on the base, and `Reported:112-118` **covariantly narrows** them to `String` / `SwapAsset` / `SwapAsset`. So `SwapQuoteMismatchScreen.kt:47`'s `exception as? SwapQuoteMismatchException.Reported` gives the compiler three non-nulls via smart-cast, and `:55-60` reads them with no null branch and no `!!`. That's a compile-time guarantee restored, not a subclass quietly passing non-null values into nullable base fields — the latter would have left the consumer staring at nullables and changed nothing. Making `withReportContext` return `Reported` and keeping `withQuoteReportContext` private as its sole caller closes the loop, and dropping the assets from `Rejected`/`AmountInconsistency` entirely means a throw site can't half-populate the context.

The remaining `return null` at `:46-53` is correct — `SwapQuoteData.Error` carries ordinary quote failures too and both callers use null to mean "generic path" — and it now reports through `swapQuoteMismatchSignal` before falling through, which is what I asked for. Unreachable in practice: both repository catch blocks convert to `Reported`, and the typed catch precedes the generic one.

KDoc overclaim gone at both sites, and you went further than asked — `:233-238` now states outright that the `strings` map is a hand-maintained duplicate and that a copy edit in the XML doesn't reach the suite. That's the honest version. `swapMismatchReportBodyPinsAllFourPlaceholders` still pins all four in order.

Thanks for the note on the restructure — that's all I wanted.

**On CI: don't chase `e2e-android (phase 2)`.** I had asked for the re-run; it ran and is still red, but the two annotations ("Post-swap-Confirm Sent! screen did not render in time", the missing reverse-swap QR) appear **verbatim on six unrelated branches** between 09-02 and 09-06, including ones that touch no swap code — mob-1832, MOB-1836, mob-1656, and three others. It's intermittent rather than permanently broken (phase 2 was green on `fix/automatic-server-send-balance` and `chore/near-e2e-balanced-amounts`), but the exact-signature match across unrelated branches settles it. Not yours. Same for `Keystone` and the emulator screenshot job, whose only failure is a restore-wallet screenshot on a device that went offline.

Everything that actually gates this is green.

LukasKorba (approved):
Approving.

The exception hierarchy ended up in better shape than the value model I originally signed off — `Reported` narrowing the three fields is a stronger guarantee than a separate value class carried alongside, because the context can't be absent on a type that requires it. Fail-closed behaviour preserved throughout, no PII beyond the existing support-email set, and the ES strings check out.

Not blocking, for whenever you touch it: the Spanish copy still hasn't been through a translator, and the crash-signal volume is now every mismatch type from both quote paths rather than AmountInconsistency-only — intended, just worth watching on a maint branch.

Issue #14, opened 2021-10-05
Add Splash Screen


ccjernigan:
Using the new Android X [Splash Screen Compat library](https://developer.android.com/reference/kotlin/androidx/core/splashscreen/SplashScreen), configure a splash screen for the app.

An open question for this implementation is to decide how to handle graphical assets that might need to be considered for whitelabeling.  While various assets in the app are generic, something like the splash screen might need to be implemented in a way as to be more configurable for different use cases.

PR #2495, opened 2026-09-03
MOB-1832: Apply server switch hysteresis to automatic server selection

nesence-m:
## Summary

App half of the server-switch hysteresis for automatic server selection, tracked in [MOB-1832](https://linear.app/zodl/issue/MOB-1832/add-hysteresis-to-automatic-server-selection-to-prevent-flip-flopping) and mirroring iOS (zodl-ios #2049). Companion SDK PR: https://github.com/zodl-inc/zodl-android-wallet-sdk/pull/14 (this branch pins it via `SDK_COMMIT_PIN`).

- `AutomaticServerRepositoryImpl.init()` is now a single lane: every foreground edge (including the first real foreground on launch) calls the SDK's `evaluateServerSwitch(current, allBundledEndpoints, 5 s, 1 block)`; a new edge cancels an in-flight benchmark (`mapLatest`). A non-null answer is re-validated (automatic still on, no transaction in flight, candidate still bundled) and applied through the existing `updateWalletEndpoint` pipeline. A `null` answer changes nothing.
- Previously any benchmark winner that differed from the persisted endpoint triggered a Synchronizer rebuild, even one measured 5 ms faster. Now a switch needs ≥ 200 ms and ≥ 25 % improvement, or an unhealthy current server (policy and thresholds live in the SDK).
- The settings-screen benchmark and the manual "Automatic" save path (plain winner, no hysteresis) are unchanged.

Merged with `maint/v3.10.x` after #2489 (MOB-1723) landed: its balance-snapshot gate now lives inside the hysteresis pipeline, so a switch candidate waits for the first local balance snapshot (inside the cancellable `mapLatest` block) before it is applied; `resolveAutomaticServerCandidate` went away with the `fastestEndpoints` lane it gated. Shipping on 3.10.x requires an SDK release from `maint/v3.1.x` first, then clearing `SDK_COMMIT_PIN` and bumping `ZCASH_SDK_VERSION`.

## Test plan

- [x] `AutomaticServerRepositoryTest` (12 tests): automatic off → no SDK call; in-transaction state → no SDK call; SDK null/endpoint passed through; SDK called with current endpoint, every bundled candidate, 5 s and 1 block; candidate waits for / is released by the local balance snapshot
- [x] `./gradlew :zashi-android:ktlint :zashi-android:detektAll` clean (from the task-workspace root)
- [x] `./gradlew :zashi-android:app:assembleZcashmainnetStoreDebug` against the pinned SDK branch
- [ ] Device check: background/foreground the app, confirm one `Server Switch:` log line per foreground and no wallet reconnect when the decision is "stay"

> **Note**

# Author

> **Note**

# Reviewer

https://claude.ai/code/session_01PYn1vxkynZJHRJQWVSkvtE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ### Changed:
 
-- Automatic server selection now switches only to a server that is meaningfully faster than the one you're on (at least 200 ms and 25 % faster), or when the current server fails its health check twice in a row, so the wallet no longer flips between near-equal or briefly slow servers. A switch is never followed by another within ten minutes, and the servers are re-checked at most once every ten minutes instead of on every return to the app.
+- Automatic server selection now switches only to a server that is meaningfully faster than the one you're on (at least 200 ms and 25 % faster), or when the current server fails its health check twice in a row, so the wallet no longer flips between near-equal or briefly slow servers. A switch is never followed by another within thirty minutes, and the servers are re-checked at most once every ten minutes instead of on every return to the app.
 - Coinholder Polling's poll list no longer refreshes itself in the background while you're just browsing it; it now updates when you open the screen, return to it, or pull to refresh.
 - Vote submission is faster and more reliable over Tor: a slow or unresponsive vote-helper server no longer stalls the whole submission.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ### Changed:
 
+- Automatic server selection now switches only to a server that is meaningfully faster than the one you're on (at least 200 ms and 25 % faster), or when the current server fails its health check, so the wallet no longer flips between near-equal servers on every check.
 - Coinholder Polling's poll list no longer refreshes itself in the background while you're just browsing it; it now updates when you open the screen, return to it, or pull to refresh.
 - Vote submission is faster and more reliable over Tor: a slow or unresponsive vote-helper server no longer stalls the whole submission.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ### Changed:
 
-- Automatic server selection now switches only to a server that is meaningfully faster than the one you're on (at least 200 ms and 25 % faster), or when the current server fails its health check, so the wallet no longer flips between near-equal servers on every check.
+- Automatic server selection now switches only to a server that is meaningfully faster than the one you're on (at least 200 ms and 25 % faster), or when the current server fails its health check twice in a row, so the wallet no longer flips between near-equal or briefly slow servers. A switch is never followed by another within ten minutes, and the servers are re-checked at most once every ten minutes instead of on every return to the app.
 - Coinholder Polling's poll list no longer refreshes itself in the background while you're just browsing it; it now updates when you open the screen, return to it, or pull to refresh.
 - Vote submission is faster and more reliable over Tor: a slow or unresponsive vote-helper server no longer stalls the whole submission.
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -124,7 +124,7 @@ SDK_INCLUDED_BUILD_PATH=
 # from something more durable than the SDK's own PR branch (which stops being fetchable once
 # that branch is deleted): repoint to the merged sha on origin/main (or the relevant
 # origin/maint/vX.Y.x line) once the companion SDK PR lands.
-SDK_COMMIT_PIN=ed11a76d1ff064950689a483f6f16881e8d4c396
+SDK_COMMIT_PIN=37bd2393ab3f206bc91e2b5fda9a9f4c85442afb
 # When blank, it pulls the BIP-39 library from Maven.
 # When set, it uses the path for a Gradle included build.  Path can either be absolute or relative to the root of this app's Gradle project.
 BIP_39_INCLUDED_BUILD_PATH=
--- a/gradle.properties
+++ b/gradle.properties
@@ -124,7 +124,7 @@ SDK_INCLUDED_BUILD_PATH=
 # from something more durable than the SDK's own PR branch (which stops being fetchable once
 # that branch is deleted): repoint to the merged sha on origin/main (or the relevant
 # origin/maint/vX.Y.x line) once the companion SDK PR lands.
-SDK_COMMIT_PIN=
+SDK_COMMIT_PIN=ba6d49acf1cb8a8c2d885fe618d9612799cdc309
 # When blank, it pulls the BIP-39 library from Maven.
 # When set, it uses the path for a Gradle included build.  Path can either be absolute or relative to the root of this app's Gradle project.
 BIP_39_INCLUDED_BUILD_PATH=
--- a/gradle.properties
+++ b/gradle.properties
@@ -124,7 +124,7 @@ SDK_INCLUDED_BUILD_PATH=
 # from something more durable than the SDK's own PR branch (which stops being fetchable once
 # that branch is deleted): repoint to the merged sha on origin/main (or the relevant
 # origin/maint/vX.Y.x line) once the companion SDK PR lands.
-SDK_COMMIT_PIN=6a30e60f3fb39c435893c00ef25e2334d1657110
+SDK_COMMIT_PIN=ed11a76d1ff064950689a483f6f16881e8d4c396
 # When blank, it pulls the BIP-39 library from Maven.
 # When set, it uses the path for a Gradle included build.  Path can either be absolute or relative to the root of this app's Gradle project.
 BIP_39_INCLUDED_BUILD_PATH=
--- a/gradle.properties
+++ b/gradle.properties
@@ -109,7 +109,7 @@ ZCASH_GOOGLE_PLAY_DEPLOY_STATUS=draft
 # When blank, it pulls the SDK from Maven.
 # When set, it uses the path for a Gradle included build.  Path can either be absolute or relative to the root of this app's Gradle project.
 SDK_INCLUDED_BUILD_PATH=
-# Optional commit hash on zcash/zcash-android-wallet-sdk. When set, CI checks out the SDK at
+# Optional commit hash on zodl-inc/zodl-android-wallet-sdk. When set, CI checks out the SDK at
 # this exact commit and builds against it (via SDK_INCLUDED_BUILD_PATH) instead of the
 # published Maven artifact — for co-developing an app change against unreleased SDK work,
 # without waiting on an SDK release. Locally this has no effect: set
@@ -124,7 +124,7 @@ SDK_INCLUDED_BUILD_PATH=
 # from something more durable than the SDK's own PR branch (which stops being fetchable once
 # that branch is deleted): repoint to the merged sha on origin/main (or the relevant
 # origin/maint/vX.Y.x line) once the companion SDK PR lands.
-SDK_COMMIT_PIN=ba6d49acf1cb8a8c2d885fe618d9612799cdc309
+SDK_COMMIT_PIN=6a30e60f3fb39c435893c00ef25e2334d1657110
 # When blank, it pulls the BIP-39 library from Maven.
 # When set, it uses the path for a Gradle included build.  Path can either be absolute or relative to the root of this app's Gradle project.
 BIP_39_INCLUDED_BUILD_PATH=
--- a/gradle.properties
+++ b/gradle.properties
@@ -124,7 +124,7 @@ SDK_INCLUDED_BUILD_PATH=
 # from something more durable than the SDK's own PR branch (which stops being fetchable once
 # that branch is deleted): repoint to the merged sha on origin/main (or the relevant
 # origin/maint/vX.Y.x line) once the companion SDK PR lands.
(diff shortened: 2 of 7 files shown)


LukasKorba (commented):
Reviewed together with zodl-inc/zodl-android-wallet-sdk#14 — most of the substance is over there (the failure path has no hysteresis, so A→B→A flapping is still reachable; plus two gRPC leak paths). This side is smaller, but one blocking item and a bad merge artifact.

## Blocking — the only lane has no error handling

`AutomaticServerRepository.kt:104-111`:

```kotlin
applicationStateProvider.observeOnForeground()
    .mapLatest { resolveSwitchCandidate() }
    .filterNotNull()
    .onEach { applyServerSwitch(it) }
    .launchIn(scope)
```

No `.catch`, no `runCatching`, no `retry`. `resolveSwitchCandidate` (`:117-129`) reaches a DataStore read (`isServerAutomatic()`), an encrypted-prefs read (`getPersistableWallet()`), and a multi-endpoint network benchmark. Any throw kills the flow — a `StatusRuntimeException` escaping the un-`runCatching`'d `getLatestBlockHeight` at `FastestServerFetcher.kt:291`, an `IOException` from prefs, or the `IllegalArgumentException` from the SDK's `require` at `:96`.

`init()` is called once, and `scope` is `CoroutineScope(Dispatchers.IO + SupervisorJob())` (`:48`) with **no `CoroutineExceptionHandler`** — so the failure goes to the global handler and then the thread's uncaught handler. Likely a crash; at best automatic server selection is permanently dead until process restart, silently. Wrap `resolveSwitchCandidate()` in `runCatching` and log.

## Should-fix

**`gradle.properties:112` — the merge reverted this comment to the wrong upstream repo.** Base `maint/v3.10.x` reads `# Optional commit hash on zodl-inc/zodl-android-wallet-sdk.`; this PR changes it to `zcash/zcash-android-wallet-sdk`. Nothing in the description mentions it and neither commit needed it — looks like a bad conflict resolution in merge commit `69f48b5`. The pinned SHA `ba6d49dd…` doesn't exist in `zcash/zcash-android-wallet-sdk`, so the comment now points a reader at a repo where the pin is unresolvable. Revert that line.

**The fake `Synchronizer` can't fail**, so the blocking item above is invisible to the suite — every stub in `AutomaticServerRepositoryTest.kt` is `returns null` or `returns known.last()`, no `throws` case, and nothing tests that a cancelled `mapLatest` leaves nothing applied. The 12 tests do faithfully pin the arguments (`sdkIsCalledWithTheCurrentEndpointAndEveryKnownCandidate` — `current`, all `known`, `5.seconds`, `blocksToFetch = 1`) and the balance-snapshot gate, which is the right thing to pin.

## Merge order — yes, this needs unpinning before it goes anywhere

1. SDK #14 lands on `maint/v3.1.x` first — this can't compile without it, since `Synchronizer` gains an abstract `evaluateServerSwitch` (breaking for every implementer; the CHANGELOG flags it correctly and `SlipstreamSynchronizer.kt:933-944` delegates properly).
2. Cut an SDK release from `maint/v3.1.x`.
3. **Immediately after #14 merges:** repoint `SDK_COMMIT_PIN` — `gradle.properties:127` currently pins `ba6d49dd…`, which is the SDK PR *head*, i.e. a feature-branch commit that stops being fetchable once GitHub deletes the branch. That violates the constraint written a few lines above it in the same file ("must be reachable from something more durable than the SDK's own PR branch"). Same for the base branch's stale `0865690d…` if it's in the same state.
4. **Before this reaches `release/*` or `candidate/*`:** clear `SDK_COMMIT_PIN` entirely and bump `ZCASH_SDK_VERSION` from `3.1.1-SNAPSHOT` to the released version. Per the file's own comment CI hard-fails on a non-blank pin there, so it's a gate, not a nicety.

Base branches pair correctly (SDK `maint/v3.1.x` ↔ app `maint/v3.10.x` at `3.1.1-SNAPSHOT`), and the call site itself is clean — named arguments throughout (`:123-128`), `Duration` makes a ms/s mixup impossible, the explicit `5.seconds`/`1` match the SDK defaults, nullable return handled. The one real integration gap is the unenforced "`candidates` must include `current`" invariant, which I've written up on the SDK PR — worth fixing on whichever side you prefer, but not leaving to a doc comment.

LukasKorba (commented):
Lane protection and the comment revert are both properly fixed. The substantive findings this round are cross-repo, so they're written up on zodl-inc/zodl-android-wallet-sdk#14 — the short version is that **the SDK's 10-minute cooldown can never fire on this app's path**, because `evaluationInterval.markCompleted()` (`:158`) runs after the SDK has already stamped `lastSwitchAt`, and both windows are 10 minutes. K=2 is the only hysteresis actually running. Two more over there: `markCompleted()` only fires on success, so a user toggling foreground faster than an evaluation completes gets an unthrottled full benchmark every edge; and `TimeSource.Monotonic` at `:177` doesn't advance in deep sleep, so an overnight sleep can filter out the morning foreground.

**A1 fixed properly.** `:121-128` — `guarded(...)` wraps *both* `resolveSwitchCandidate()` and `applyServerSwitch(candidate)`, and `guarded` (`:130-139`) rethrows `CancellationException` at `:136` before logging, so `mapLatest` semantics are preserved. Failures return null and get dropped by `filterNotNull()`, lane stays alive for the next edge. The mockk tests with explicit throws (`:207-227`, `:230-244`) are the right shape, and asserting that the *next* edge still applies a switch is the assertion that matters.

Two leftovers: `observeOnForeground()` itself is still uncaught, so a throw from the upstream flow rather than the transform still kills the lane at `:106` (low). And there's still no test that a cancelled `mapLatest` leaves nothing applied — `theLaneSurvivesAFailureToApplyTheSwitch` (`:230`) asserts one call happened, not that the lane survived, because the 10-minute interval makes a second edge unobservable in that test.

**A2 fixed** — `gradle.properties:112` reads `zodl-inc/zodl-android-wallet-sdk` again.

**A4 — not fixed, and it can't be yet.** To be clear, I don't read this as a misrepresentation: your commit message says "repoints SDK_COMMIT_PIN at the companion SDK commit", which is exactly what happened. But `gradle.properties:127` is `6a30e60f…`, and I resolved it against the SDK repo:

- `branches-where-head` → `feature/mob-1832-server-switch-hysteresis`, and nothing else
- `compare maint/v3.1.x...6a30e60f` → `ahead_by: 4, behind_by: 0`, so it is **not** an ancestor of the maint branch

It's still a feature-branch commit that stops being fetchable when the branch is deleted, which is what the constraint at `gradle.properties:123-126` exists to prevent. Unchanged sequence: merge #14 → repoint to the merged SHA on `origin/maint/v3.1.x` → blank the pin and bump `ZCASH_SDK_VERSION` off `3.1.1-SNAPSHOT` (`:206`) before this touches `release/*` or `candidate/*`.

LukasKorba (commented):
All three app findings fixed correctly, and the confirm handshake is called on exactly the right paths. Good from my side — the remaining work is on zodl-inc/zodl-android-wallet-sdk#14 (the SDK cooldown still uses `TimeSource.Monotonic`, which your own KDoc at `:227-230` explains is the wrong clock — so a phone that sleeps overnight can have the app's interval elapse while the SDK's cooldown hasn't, deferring the morning failover off a dead server for hours).

**N2** — `markAttempted()` in the `finally` at `:180-182`, and it's non-suspending so it runs on the cancellation path too. I checked the starvation risk I warned about and it isn't reachable: a user toggling every 30s gets edges dropped for 10 minutes, then the next window opens an attempt again — it retries once per 10 min indefinitely and never latches off. Early returns at `:168-172` sit before the `try` and so don't mark, which is right: no network work was done.

**N3** — `ElapsedRealtimeTimeSource` at `:232-240`, wired as the default at `:215`. The rationale comment is the clearest thing in the diff.

**N7** — bounded at 30s (`:247`), and you made the right call on semantics: **the switch still applies on timeout** and which path was taken is logged (`:152-164`). That's correct for the `CurrentUnhealthy` case, which is exactly why the unbounded wait was the wrong gate. Both ends pinned.

**The handshake.** Taking the synchronizer *before* the write (`:197` vs `:199`) is load-bearing and the comment at `:185-192` gets it right — `getSynchronizerOrNull()` suspends on `filterNotNull().first()`, so calling it after would block until the rebuild finished. I traced the decline, throw and process-death paths and they all preserve the count correctly.

Two lows, neither blocking:

- **The cancellation gap at `:199-200` is closed by accident, not by design.** `markAttempted()` runs in `evaluateServerSwitch`'s finally, *before* the balance wait and before `applyServerSwitch`, so by the time you reach the write `lastAttemptAt` is ~30s old and the filter at `:128` drops every edge for ~9.5 min — meaning nothing can cancel you between the write and the confirm. That holds today, but it depends on `MINIMUM_EVALUATION_INTERVAL` staying long and the filter staying where it is. If a confirm is ever missed after a successful write and the new server also fails to measure, the stale count of 2 increments to 3, already past threshold, with no cooldown — a switch every 10 minutes walking the host list. Cheap insurance: `withContext(NonCancellable) { updateWalletEndpoint(candidate); synchronizer?.confirmServerSwitch(candidate) }`.
- **`confirmServerSwitch` fires even when `updateWalletEndpoint` no-ops.** `WalletRepository.kt:218-223` returns silently on a null wallet or an unchanged endpoint, and `:200` confirms unconditionally — starting a 30-minute cooldown for a switch that never happened, which is the hazard your own SDK doc at `Synchronizer.kt:285-286` warns about. Narrow reachability, but returning a `Boolean` from `updateWalletEndpoint` and gating the confirm on it would be exact.

Nit: `EvaluationInterval.lastAttemptAt` (`:217`) is a non-volatile `var` written from the `mapLatest` child and read from the collector in `filter`, on a shared Koin `single`. Worst case is one extra benchmark and dispatch makes a stale read unlikely, but `@Volatile` is free.

**Pin** — correctly repointed to `ed11a76`, the new SDK head. Still a feature-branch SHA, as expected; same sequence stands: merge #14 → repoint to the merged SHA on `origin/maint/v3.1.x` → blank the pin and bump `ZCASH_SDK_VERSION` before `release/*` or `candidate/*`.

**CI:** `Keystone` is red on every completed run of `e2e-smoke.yml` across six unrelated branches over the last three days — repo-wide, not this PR. `test_android_modules_emulator (app)` shows `AdbCommandRejectedException: device offline` and "Expected 12 tests, received 0", i.e. the emulator died rather than an assertion failing; the prior head passed the same job. Worth one rerun, but the signature is infra.

LukasKorba (commented):
`@Volatile` at `:217` and the pin repoint are in. Your summary said "pin repointed, volatile nit" and that's exactly what landed — no complaint about the description, but flagging that the other two lows from my last round are still open, in case they were meant to be in scope:

- **`:199-200`** — the write→confirm pair is still bare. Today nothing can cancel you in that gap, because `markAttempted()` runs in `evaluateServerSwitch`'s finally before the balance wait, so the filter at `:128` has already closed the window. That's incidental protection, not structural: shorten `MINIMUM_EVALUATION_INTERVAL` or move the filter and it reopens. `withContext(NonCancellable) { … }` around the pair is the cheap insurance.
- **`WalletRepository.kt:218-223`** — still `suspend fun updateWalletEndpoint(endpoint): Unit`, returning silently at `:219` (null wallet) and `:221` (unchanged endpoint), with `:200` confirming unconditionally afterwards. A `Boolean` return gated at the call site would make it exact.

Good news on both: **the endpoint keying you added on the SDK side has already shrunk their blast radius.** A lost confirm no longer poisons the new server's failure count — the next evaluation sees `countedAgainst ≠ current` and resets — so what's left in both cases is a spurious cooldown on *improvement* switches only, not a mis-scoped failure count. That drops them from "worth doing" to "do it if you're touching the file". Your call whether they land here or as a follow-up.

**Pin** is `37bd239`, correctly tracking the new SDK head. Still a feature-branch SHA (`ahead_by: 6` of `maint/v3.1.x`, not an ancestor), which is expected and unresolvable until #14 merges.

**CI:** `test_kotlin_modules` is green, which is the job that would catch the `@Volatile` change. `Keystone` and `test_android_modules_emulator (app)` are red on `bugfix/MOB-1340` and `feature/mob-1740` too, and on your own previous head. This run's emulator booted after `device offline` retries and a failed snapshot load, then all 12 `ScreenshotTest` cases failed identically with `ComposeTimeoutException` after 100s — a whole suite timing out uniformly on a sick emulator, reproducing on unrelated branches. A one-line `@Volatile` plus a pin bump cannot reach `ScreenshotTest`. Not yours.

LukasKorba (approved):
Approving. `SDK_COMMIT_PIN=66701636…` resolves to zodl-inc/zodl-android-wallet-sdk#14's merge commit and is `identical` to `maint/v3.1.x` head, so the durability constraint at `gradle.properties:123-126` is satisfied. That was the last blocker.

Four rounds of findings closed: the lane is guarded on both call sites and survives a failure, the interval is marked in a `finally` on `elapsedRealtime`, the balance wait is bounded at 30s and — correctly — still applies the switch on timeout, and the confirm handshake fires on exactly the paths that should confirm.

The two lows I raised are still open and I'm signing off with them open, deliberately: `withContext(NonCancellable)` around the write→confirm pair at `:199-200`, and gating the confirm on a `Boolean` from `updateWalletEndpoint`. The endpoint keying you added on the SDK side shrank both — a lost confirm no longer poisons the new server's failure count — so what remains is a spurious cooldown on improvement switches only, bounded by the 200ms/25% gates. Worth picking up next time you're in this file, not worth another round now.

CI reds are the repo-wide `e2e-smoke.yml` outage and the sick-emulator screenshot job, neither attributable here.

PR #2496, opened 2026-09-03
fix(detekt): keep isServerCustom under the ReturnCount limit

nesence-m:
`./gradlew detektAll` has been failing on `main` since #2448 merged (2026-08-19):

```
ui-lib/src/main/java/co/electriccoin/zcash/ui/common/repository/AutomaticServerRepository.kt:88:26: Function isServerCustom has 3 return statements which exceeds the limit of 2. [ReturnCount]
```

Because `pull_request` runs build the merge commit, every PR targeting `main` since then shows a red `static_analysis_detekt` check regardless of its own content (#2430, #2455, #2457, #2462, #2488).

This folds the `?: return false` on the endpoint into the final boolean expression. Evaluation order and result are unchanged: automatic mode still short-circuits before the wallet is read, and a missing endpoint still yields `false`.

Verified locally with `./gradlew detektAll` (green).

https://claude.ai/code/session_01ALk8No4AemVzm2PcV5WrWS


Chlup (approved):

LukasKorba (approved):
utACK

PR #2497, opened 2026-09-03
MOB-1836: Return to onboarding after the encrypted store is wiped; retry and quarantine instead of delete (MOB-1565)

nesence-m:
Fixes [MOB-1836 — Exception on Android after updating from an old release](https://linear.app/zodl/issue/MOB-1836/exception-on-android-after-updating-from-an-old-release) and [MOB-1565 — Seed destroyed on any transient Keystore/encrypted-prefs exception](https://linear.app/zodl/issue/MOB-1565/security-seed-destroyed-on-any-transient-keystoreencrypted-prefs).

## Problem

The wallet seed lives in the encrypted preferences store; the onboarding flag lives in the plain store. When the encrypted store is provably unreadable (for example after a device-to-device transfer, or the `AEADBadTagException` from MOB-1452), `newEncrypted` deletes and recreates it (MOB-1452 in 3.7.1, narrowed by MOB-1691 in 3.10.1) but nothing resets the onboarding flag. `secretState` derived READY from the flag alone, so the app opened Home with no wallet behind it and the balance and transaction list spun forever. That is the state the MOB-1836 user is in on 3.10.2: the reported exception predates their update, and 3.10.2 wiped the store silently.

MOB-1565 additionally asked for a bounded retry before the store is condemned, and for the corrupted file to be set aside rather than deleted.

## Fix

**MOB-1836 (ui-lib)**
- `SecretState.READY` now requires both the onboarding flag and a stored wallet (`resolveSecretState`); READY with no wallet resolves to NONE, and `RootNavGraph` already routes NONE to onboarding. The flow deliberately never maps a store failure to null.
- Startup self-heal in `WalletRepositoryImpl.init()`: when the flag says READY but no wallet is stored, erase the stale SDK databases and persist the flag back to NONE. This is what recovers users who were already wiped by 3.10.1/3.10.2.
- `createNewWallet()` persists the wallet before marking onboarding READY, so a crash between the two writes can never leave "flag without wallet".
- "Reset Zodl" also purges the quarantine directory.
- A failed `Synchronizer.erase` on the self-heal path is reported through `GlobalCrashReporter` as well as logged: "the erase failed and we carried on anyway" is the half-reset state this ticket was reported for, and a `Twig.error` never leaves the device.

**MOB-1565 (preference-impl-android-lib)**
- Opening the encrypted store is retried three times with a short backoff before any failure is classified (the MOB-1452 trace showed `AEADBadTagException` caused by a Keystore operation error).
- Provable corruption now quarantines `shared_prefs/<file>.xml` and its `.bak` sibling under `no_backup/encrypted_prefs_quarantine/` instead of deleting them. Retention keeps the oldest copy of a filename unconditionally plus the two newest: only the first quarantine can set aside a store that ever held a wallet, because the recovery that follows recreates it empty. A quarantine that fails mid-move is rethrown with the original files left in place — there is no delete fallback. The Keystore master-key alias is deleted only for key-level failures, so a quarantined file stays decryptable by this device for data-level ones.
- A cause chain carrying AOSP's transient marker (`android.security.KeyStoreException`) is never classified as corruption *or* as a master-key failure — one shared veto used by both classifiers — so a wedged Keystore daemon rethrows and leaves the store where it is.
- The persisted recreated-marker, written only after a quarantine actually succeeded, bounds the immediate retry within a run: it is cleared as soon as a store opens successfully, so it does **not** bound repeat quarantines across launches. Surviving repeat quarantines is what the retention policy above is for.

Deliberately left for follow-up tickets: the `:crash` process running the full app init, a `SecretState.ERROR` for a persistently unreadable store (which is also where the erase-after-a-failed-`ensureReadable` question belongs), finer API 33+ `KeyStoreException` classification, and the pre-existing `Twig.error(e)` on a `PersistableWallet` parse failure, which can log decrypted wallet JSON through `JSONTokener.syntaxError`.

## Test plan

- `:preference-impl-android-lib:testDebugUnitTest` — 35/35 (retry ladder, classification incl. the transient-marker veto for both classifiers, quarantine/prune retention, quarantine-failure rethrow).
- `:ui-lib:testZcashmainnetStoreDebugUnitTest` — full suite 469/469, incl. new `WalletRepositoryImplTest` (resolver, real combine, create-wallet ordering, init self-heal).
- `:ui-lib:testZcashmainnetInternalDebugUnitTest --tests '*WalletRepository*'` — 11/11 after the review fixes, incl. the fail-safe hinge (a wallet read that throws writes no onboarding state and erases nothing) and the failed-erase crash report.
- `:preference-impl-android-lib:connectedDebugAndroidTest` on Pixel 9 Pro XL (API 35) — 12/12; the two recovery tests now assert the quarantine file, no leftover `.bak`, and alias retention.
- ktlint + detektAll clean.
- Manual, debug build against the pinned SDK on the emulator:
  - corrupted Tink keyset → three failed attempts logged → quarantine → onboarding, flag reset, SDK DB erased; creating a new wallet from there reaches Home;
  - encrypted store deleted with the flag still READY (the reported user's state) → self-heal → onboarding.

# Author

- [ ] **Self-review** your own code in GitHub's web interface
- [ ] Update the [**manual tests**](../blob/main/docs/testing/manual_testing) as appropriate
- [ ] Check the **code coverage** report for the automated tests
- [x] Update **documentation** as appropriate (CHANGELOG.md)
- [ ] Pull in the latest changes from the **main** branch and **squash** your commits before assigning a reviewer

# Reviewer

- [ ] Perform an **ad hoc review**
- [ ] Review the **documentation**

https://claude.ai/code/session_016rqq1yYyS5XofPcqE7bZ4r

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 - Support/error reports for a failed transaction no longer claim a fake `gRPC: false, code: -1` status when the failure actually happened before submission (e.g. a Sapling parameter download failure); such reports now correctly identify the real exception instead.
 - Opening Send immediately after a cold start no longer falls back to a loading screen while automatic server selection reconnects the wallet. The automatic winner waits for the local balance snapshot, which remains visible while the new server connection settles.
 - Fixed a crash loop some wallets hit at startup when the SDK reported the stored seed as no longer relevant to its database. The app now recovers automatically by clearing the mismatched local data and rescanning, instead of crashing repeatedly.
+- The app returns to onboarding so you can restore from your recovery phrase when its secret store is found empty while it still believes a
+  wallet was set up, instead of opening a home screen with no wallet behind it.
+- Opening the encrypted secret store is retried a few times before it is declared unreadable, and an unreadable store is set aside on the
+  device instead of being deleted.
 
 ## [3.10.1 (2512)] - 2026-08-25
 
--- a/app/src/main/java/co/electriccoin/zcash/app/ZcashApplication.kt
+++ b/app/src/main/java/co/electriccoin/zcash/app/ZcashApplication.kt
@@ -170,4 +170,7 @@ class ZcashApplication : CoroutineApplication() {
     }
 }
 
+/**
+ * Kept in sync with `GlobalCrashReporter`'s own copy of this suffix — see [isCrashProcess].
+ */
 private const val CRASH_PROCESS_NAME_SUFFIX = ":crash"
--- a/app/src/main/java/co/electriccoin/zcash/app/ZcashApplication.kt
+++ b/app/src/main/java/co/electriccoin/zcash/app/ZcashApplication.kt
@@ -17,6 +17,7 @@ import co.electriccoin.zcash.di.viewModelModule
 import co.electriccoin.zcash.migration.di.featureMigrationModule
 import co.electriccoin.zcash.spackle.StrictModeCompat
 import co.electriccoin.zcash.spackle.Twig
+import co.electriccoin.zcash.spackle.process.ProcessNameCompat
 import co.electriccoin.zcash.ui.common.provider.CrashReportingStorageProvider
 import co.electriccoin.zcash.ui.common.provider.MigrationNotifier
 import co.electriccoin.zcash.ui.common.provider.SynchronizerProvider
@@ -59,6 +60,11 @@ class ZcashApplication : CoroutineApplication() {
         super.onCreate()
 
         configureLogging()
+
+        if (isCrashProcess()) {
+            Twig.info { "Skipping app initialization in the crash-reporting process" }
+            return
+        }
 
         preloadSdkNativeLibrary()
 
@@ -133,6 +139,15 @@ class ZcashApplication : CoroutineApplication() {
         }
     }
 
+    /**
+     * The `:crash` secondary process only hosts `ExceptionReceiver` and the crash content
+     * provider, which are Koin-free — everything below this check assumes Koin is started.
+     * [co.electriccoin.zcash.crash.android.GlobalCrashReporter.register] has its own copy of the
+     * same `:crash` suffix check, using a constant `internal` to crash-android-lib, so the literal
+     * is deliberately repeated here rather than shared.
+     */
+    private fun isCrashProcess() = ProcessNameCompat.getProcessName(this).endsWith(CRASH_PROCESS_NAME_SUFFIX)
+
     private fun configureStrictMode() {
         if (BuildConfig.DEBUG) {
             StrictModeCompat.enableStrictMode(BuildConfig.IS_STRICT_MODE_CRASH_ENABLED)
@@ -154,3 +169,5 @@ class ZcashApplication : CoroutineApplication() {
         }
     }
 }
+
+private const val CRASH_PROCESS_NAME_SUFFIX = ":crash"
--- a/preference-impl-android-lib/build.gradle.kts
+++ b/preference-impl-android-lib/build.gradle.kts
@@ -21,6 +21,10 @@ android {
             execution = "ANDROIDX_TEST_ORCHESTRATOR"
         }
     }
+
+    testOptions {
+        unitTests.isReturnDefaultValues = true
+    }
 }
 
 dependencies {
--- a/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
+++ b/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
@@ -127,6 +127,7 @@ class EncryptedPreferenceProviderTest {
 
             assertFalse(sharedPrefsBakFile(context, RECOVERY_FILENAME).exists())
             assertTrue(androidKeyStoreAlias().containsAlias(MasterKey.DEFAULT_MASTER_KEY_ALIAS))
+            assertFalse(recreatedMarkerFile(quarantineDirectory(context), RECOVERY_FILENAME).exists())
         }
 
     @Test
@@ -169,6 +170,7 @@ class EncryptedPreferenceProviderTest {
             assertEquals(1, quarantined.size)
 
             assertTrue(androidKeyStoreAlias().containsAlias(MasterKey.DEFAULT_MASTER_KEY_ALIAS))
+            assertFalse(recreatedMarkerFile(quarantineDirectory(context), D2D_FILENAME).exists())
 
             val expectedValue = StringDefaultPreferenceFixture.DEFAULT_VALUE + "restored"
             restoredProvider.putString(StringDefaultPreferenceFixture.KEY, expectedValue)
@@ -200,13 +202,13 @@ class EncryptedPreferenceProviderTest {
         private fun sharedPrefsBakFile(
             context: Context,
             filename: String
-        ) = File(File(context.filesDir.parent, "shared_prefs"), "$filename.xml.bak")
+        ) = encryptedPreferencesBackupFile(sharedPreferencesDirectory(context), filename)
 
         private fun quarantinedFiles(
             context: Context,
             filename: String
         ): List<File> {
-            val quarantineDir = File(context.noBackupFilesDir, QUARANTINE_DIRECTORY)
+            val quarantineDir = quarantineDirectory(context)
             return quarantineDir
                 .listFiles { file -> file.name.startsWith("$filename-") && file.name.endsWith(".xml") }
                 ?.toList()
--- a/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
+++ b/preference-impl-android-lib/src/androidTest/java/co/electriccoin/zcash/preference/EncryptedPreferenceProviderTest.kt
@@ -115,12 +115,18 @@ class EncryptedPreferenceProviderTest {
                 .putString(VALUE_KEYSET, "corrupted_keyset_data")
                 .commit()
 
-            // Should NOT throw — fix catches the parse/decryption exception and recreates fresh prefs
             val restoredProvider =
                 AndroidPreferenceProvider.newEncrypted(context, RECOVERY_FILENAME)
 
             // Prefs are empty: user will re-enter seed phrase to recover wallet
             assertFalse(restoredProvider.hasKey(StringDefaultPreferenceFixture.KEY))
+
+            val quarantined = quarantinedFiles(context, RECOVERY_FILENAME)
+            assertEquals(1, quarantined.size)
+            assertTrue(quarantined.single().readText().contains("corrupted_keyset_data"))
+
+            assertFalse(sharedPrefsBakFile(context, RECOVERY_FILENAME).exists())
+            assertTrue(androidKeyStoreAlias().containsAlias(MasterKey.DEFAULT_MASTER_KEY_ALIAS))
         }
 
     @Test
@@ -154,11 +160,19 @@ class EncryptedPreferenceProviderTest {
                 deleteEntry(MasterKey.DEFAULT_MASTER_KEY_ALIAS)
             }
 
-            // Should NOT throw — the orphaned file is provably undecryptable, so it is recreated
             val restoredProvider = AndroidPreferenceProvider.newEncrypted(context, D2D_FILENAME)
 
             // Prefs are empty: user will re-enter seed phrase to recover wallet
             assertFalse(restoredProvider.hasKey(StringDefaultPreferenceFixture.KEY))
+
+            val quarantined = quarantinedFiles(context, D2D_FILENAME)
+            assertEquals(1, quarantined.size)
+
+            assertTrue(androidKeyStoreAlias().containsAlias(MasterKey.DEFAULT_MASTER_KEY_ALIAS))
+
+            val expectedValue = StringDefaultPreferenceFixture.DEFAULT_VALUE + "restored"
+            restoredProvider.putString(StringDefaultPreferenceFixture.KEY, expectedValue)
+            assertEquals(expectedValue, StringDefaultPreferenceFixture.new().getValue(restoredProvider))
         }
 
     companion object {
@@ -179,5 +193,24 @@ class EncryptedPreferenceProviderTest {
                 ApplicationProvider.getApplicationContext(),
                 FILENAME
             )
+
+        private fun androidKeyStoreAlias(): KeyStore =
+            KeyStore.getInstance("AndroidKeyStore").apply { load(null) }
+
+        private fun sharedPrefsBakFile(
+            context: Context,
+            filename: String
+        ) = File(File(context.filesDir.parent, "shared_prefs"), "$filename.xml.bak")
+
+        private fun quarantinedFiles(
+            context: Context,
+            filename: String
+        ): List<File> {
+            val quarantineDir = File(context.noBackupFilesDir, QUARANTINE_DIRECTORY)
+            return quarantineDir
+                .listFiles { file -> file.name.startsWith("$filename-") && file.name.endsWith(".xml") }
+                ?.toList()
+                .orEmpty()
+        }
     }
 }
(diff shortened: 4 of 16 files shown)


LukasKorba (commented):
Reviewed all three passes and the deferred items. Answers to your three maintainer calls at the bottom — but first, two things the bot missed across all three passes that I do want fixed here. Both are inside code this PR writes, both on the "can this orphan the only copy of a spending key" axis. Neither is a regression against `maint/v3.10.x` (which deletes the store outright), so this is still strictly safer than what's on the branch — but both fixes are small.

## Blocking

### 1. `isUnrecoverableCorruption` doesn't apply the transient-HAL veto that `isMasterKeyFailure` does

`AndroidPreferenceProvider.kt:365-374` vs `:397-403`.

```kotlin
internal fun isUnrecoverableCorruption(exception: Exception): Boolean =
    generateSequence<Throwable>(exception) { it.cause }
        .take(CAUSE_CHAIN_LIMIT)
        .any { ... || it is InvalidKeyException || ... }        // :372 — no veto

internal fun isMasterKeyFailure(exception: Exception): Boolean {
    val chain = ...
    val hasAndroidKeyStoreFailure =
        chain.any { it !is KeyStoreException && it.javaClass.simpleName == "KeyStoreException" }
    if (hasAndroidKeyStoreFailure) return false                  // :401 — veto present
    ...
}
```

Your own KDoc at `:376-395` says AOSP wraps a **transient** HAL error as `InvalidKeyException("Keystore operation failed")` carrying `android.security.KeyStoreException`. The code acts on that for key deletion and ignores it for the quarantine decision.

Sequence — Keystore daemon wedged >600ms after an OS update or a StrongBox hiccup, i.e. the MOB-1565 scenario:

1. `create()` throws that shape on all three attempts (`:431-445`, 200+400ms total).
2. `isUnrecoverableCorruption` → **true** (chain contains `InvalidKeyException`).
3. `:451 quarantine(cause)` — the seed blob leaves `shared_prefs`.
4. `isMasterKeyFailure` → false, so the key survives. Good, but too late for reachability.
5. `getPersistableWallet()` → null, onboarding is READY → `WalletRepository.kt:299` writes NONE, `:309` erases SDK databases.
6. User lands on onboarding. Seed is in `no_backup/encrypted_prefs_quarantine/`, unreachable from any UI, excluded from Auto Backup, and one **Reset Zodl** away from `purgeQuarantine()`.

Failing safe here means rethrowing and leaving the store in place: the user gets the (already-deferred) splash hang, recoverable by relaunching once the Keystore settles. Quarantine is not user-recoverable. Please extract the veto into one shared `hasTransientAndroidKeyStoreMarker(chain)` so the two classifiers can't drift again. `EncryptedPreferenceRecoveryTest.kt:83-90` pins the AOSP shape for `isMasterKeyFailure` and has no counterpart for `isUnrecoverableCorruption` — that's why this survived three passes.

### 2. `pruneQuarantine` deletes oldest-first, i.e. the only copy that has a wallet in it

`EncryptedPreferenceQuarantine.kt:110-116`, `QUARANTINE_KEEP_NEWEST = 3`:

```kotlin
xmlFiles.sortedByDescending { it.name }.drop(keepNewest).forEach { ... delete() }
```

After the first quarantine the live store is a **fresh empty** one, so every subsequent quarantine of that filename sets aside an empty or newly-created store — while the first entry, the one holding the seed, is the oldest and therefore the first thing `drop(3)` deletes.

Flaky hardware Keystore: launch 1 quarantines the real seed (T1). The recreated-marker is cleared the same run (`:260` deletes it on any successful open, and the post-quarantine `create()` on an empty store basically always succeeds), so nothing prevents recurrence. Launches 2-4 quarantine T2/T3/T4. On the fourth, `sortedByDescending` keeps T4/T3/T2 and **deletes T1**. Four app launches to total loss.

Cheapest defensible fix: keep the oldest entry for a filename unconditionally, plus the newest N-1. A few KB of XML against total loss.

Related — the PR description says "at most one recreate per filename per process", but because `:260` clears the marker on any successful open, it bounds the immediate retry, not repeat quarantine across launches. Worth correcting the description.

### 3. The quarantine-failure fallback deletes the user's ciphertext

`AndroidPreferenceProvider.kt:316-319`:

```kotlin
.onFailure { failure ->
    Twig.error(failure) { "Quarantining encrypted preferences $filename failed; deleting instead" }
    runCatching { deleteEncryptedPreferencesFiles(sharedPrefsDir, filename) }
}
```

`moveFile` (`EncryptedPreferenceQuarantine.kt:87-90`) falls back to `copyTo(overwrite = true)` + `delete()` when `renameTo` fails. If `copyTo` throws mid-write — low disk being the obvious trigger, and low disk is also a plausible reason `renameTo` failed — the exception unwinds into this `onFailure`, which then deletes the original, leaving a truncated partial in quarantine. Seed gone, log line cheerful about it.

Three lines: when quarantine fails, don't delete — rethrow so `createEncryptedPreferencesWithRecovery` propagates and the store stays put for a later attempt. "Recovery still completes" (KDoc at `:286`) isn't worth an unrecoverable seed. If you want to keep a fallback, gate it on `quarantinedXml.exists() && quarantinedXml.length() == originalLength`.

(The bot flagged `moveFile` in pass 3 but chased the `delete()`-returns-false hazard, which is harmless since `:335` deletes that path anyway.)

## Your three deferred items

**Erase after a failed `ensureReadable` — agreed, follow-up ticket.** By `WalletRepository.kt:301` we've established twice (`:296`, `:304`) that no wallet is stored, and `Synchronizer.erase` destroys SDK *databases* — block cache and derived note data, all reconstructible from the seed, which isn't in there. Skipping the erase would be no safer, just the crash-loop MOB-1836 was chasing. File it with the `SecretState.ERROR` ticket, since the honest fix is surfacing "your secret store is unreadable, do not reset" rather than routing silently to onboarding. **One thing before merge though:** the erase's `.onFailure` should log at a level that reaches crash reporting — "erase failed and we continued anyway" is the state that produces the next MOB-1836.

**Test `scope` replacement ordering — genuinely fine, close it.** A test calling `init()` before `useTestScope` would run the self-heal on real `Dispatchers.IO`, `advanceUntilIdle()` would no-op, and it'd pass vacuously. Test-integrity risk only, all current tests comply, and the bot offered no concrete fix. Not refactoring a passing test file on a maint branch.

**`sharingScope` never cancelled in production — fine.** `singleOf` + `init()` called once from `ZcashApplication.onCreate`. Process-lifetime scope on a process-lifetime singleton. Your reasoning is right.

## What I checked and found safe

The **retry-then-rethrow default is genuinely fail-safe** (`:449` requires both `isRecreateAllowed` and `isUnrecoverableCorruption`, else `:455 throw cause`), and `PreferenceHolder`/`PreferenceProviderCache` cache nothing on throw so a later launch retries cleanly. A read failure **cannot** surface as `hasWallet = false` — there's no `try`/`runCatching` anywhere in `getPersistableWallet()` → `BaseNullableStorageProvider.get()` → `PersistableWalletPreferenceDefault.getValue`, so `null` always means "key absent". That's the most important property in `init()` and it rests entirely on the *absence* of a catch three frames down — one `coEvery { … } throws` test would pin it against a future well-meaning `runCatching`. Device-locked/`UserNotAuthenticatedException` isn't reachable (`MasterKey.Builder:267-271` never sets `setUserAuthenticationRequired`). Quarantine atomicity is sound: same filesystem so `renameTo` is an atomic rename, the `.bak` sibling is taken along (`:76-78`, correct and non-obvious — leaving it would let `loadFromDisk` resurrect the corrupt data), and I walked both crash windows (die after move before marker; die after marker before recreate) — neither loops and neither erases a usable store. Master-key deletion is correctly narrowed by the veto at `:399-401`, which is what makes quarantine meaningfully non-destructive at all.

TOCTOU (finding 1 pass 1) is over-delivered — single read at `:257-260`, `walletMutation` over init/create/restore, re-read before erase at `:304-307`, pinned by `initSkipsEraseWhenAWalletIsStoredWhileSelfHealing`. Findings 2, 4, 5 and both minors check out; I verified the `resolveSecretState` one against the pre-image and the old code mapped both ordinals to `NONE` already, so there's genuinely no behaviour change.

No log statement this PR adds leaks anything — `:317`, `:325`, `:440`, `:450`, `:454` take only `filename` and counters. Separate pre-existing ticket though: `WalletRepository.kt:262-264` does `Twig.error(e)` on a `PersistableWallet` parse failure, and Android's `JSONTokener.syntaxError` appends the entire input to the message — which at that point is the **decrypted** wallet JSON. Same catch existed before this PR; flagging it because this PR makes malformed-decrypt a routine path rather than an anomaly.

## Test gaps

Three, and they map exactly onto the above: no `isUnrecoverableCorruption` assertion for the AOSP-wrapped transient shape; `pruneQuarantine` is tested for "keeps the newest" but nothing asserts which copy holds a wallet, so the retention policy's actual purpose is untested; and nothing tests the fail-safe hinge (throw → no onboarding write, no erase). The `WalletRepositoryImplTest` decision table is otherwise complete — all four quadrants, create-ordering, the race shape, the not-READY short-circuit.

## One process note

The bot's three passes converged on nits while the two highest-consequence issues in the diff went unraised in all three. Both are only visible by reading the two classifiers *against each other* and by asking what `pruneQuarantine` is retaining rather than whether it retains correctly. Worth calibrating how much weight its "Risk: Low" carries on this kind of change.

Also: the user is told nothing (`:298` is a `Twig.warn`) before landing on onboarding. Not a blocker — but a user who assumes a reinstall and taps **Reset Zodl** hits `ResetZashiUseCase.kt:94` → `purgeQuarantine()` → `deleteRecursively()`, which is the last mile of the loss chain in items 1 and 2.

LukasKorba (commented):
All three blocking items are properly closed, and (2) plus the new tests are better than the minimum ask. One thing needs to be on the record rather than riding in on a fix round — see the predicate note below. Good to merge once that's acknowledged.

## The predicate change — my call, recorded

`2de9a76` narrowed the veto I asked for. I asked for the bare `it !is java.security.KeyStoreException && simpleName == "KeyStoreException"`; head has a third conjunct, `!isPermanentAndroidKeyStoreFailure(it)` (`AndroidPreferenceProvider.kt:457-462`).

**You were right to push back and I'm accepting it.** My predicate would have stranded the device-to-device transfer case — the thing this recovery exists for — permanently: the real chain is `AEADBadTagException` ← `android.security.KeyStoreException: Signature/MAC verification failed`, `MasterKey.Builder.build()` mints a replacement under the old alias so the ciphertext can never authenticate again, and a bare marker veto reads that as transient forever. Ladder exhausts, rethrow, splash with no recourse. I'd have traded a loss path for a brick, which is not obviously the better trade. Reproducing it on API 31 and 35 and pinning it with `graceful_recovery_when_master_key_is_lost_in_migration` is the right way to have made that argument.

The structure is also what I asked for: one `hasTransientAndroidKeyStoreMarker` at `:457`, called as the first statement of both classifiers (`:486`, `:516`), with `causeChain()` shared too, so `CAUSE_CHAIN_LIMIT` can't drift either. And the ordering inside `isPermanentAndroidKeyStoreFailure` is the safe one — `isTransientPerAndroidKeyStore` first at `:430`, so when AOSP says transient that outranks whatever the message string says, and anything unclassifiable stays vetoed.

I checked the veto didn't over-apply, which was my worry: `BadPaddingException`, `AEADBadTagException`, `CharConversionException` and `InvalidProtocolBufferException` with no marker still classify as corruption and still recover. Genuine corruption isn't stuck.

**The residual I'm accepting with open eyes**, so it's written down: on API 27-32 both reflective signals are absent (`isTransientFailure()` and the numeric code are API 33+), so `PERMANENT_ANDROID_KEY_STORE_MESSAGES` (`:383-390`) is the *only* guard there. A transient failure that happened to carry `"Key not found"` — a keystore daemon whose DB isn't up yet, say — would release the veto, and then `InvalidKeyException` in the chain makes both classifiers true: quarantine the seed *and* delete the shared master key. I think that's unlikely (AOSP returns `SYSTEM_ERROR`, not `KEY_NOT_FOUND`, for daemon-level trouble) and the ~1.4s ladder covers the common case, but it's the one place the narrowing costs loss-axis margin, and it's the thing to revisit first if a pre-33 report ever comes in.

## The rest

**(2) retention — fixed, and the shape is right.** `EncryptedPreferenceQuarantine.kt:150-170`, `drop(keepNewest - 1).filterNot { it == oldest }`. I walked it out to five quarantines: Q1 survives every round, steady state is exactly 3 entries, no unbounded growth. `keepNewest = 1` and even `0` still keep the oldest, so there's no argument for which prune deletes the last copy. `.bak` pairing is correct — the filter requires `.xml` so `.bak` files are never candidates themselves, and `EncryptedPreferenceQuarantineTest.kt:83-94` asserts the three survivors' `.bak` siblings all remain.

**(3) — fixed.** The `deleting instead` block is gone; `:318-322` calls quarantine unguarded, so a throw propagates with the originals in place. The bonus is the better half: the marker now writes only after the move succeeded (`EncryptedPreferenceQuarantine.kt:99-121`), so a store that was never set aside can still recover next launch. `a failed quarantine keeps the original files and writes no marker` pins both.

**Crash reporting — fixed**, `WalletRepository.kt:314-318`, pinned with `verify(exactly = 1)` on the instance.

**All three test gaps genuinely pinned**, not adjacent: `EncryptedPreferenceQuarantineTest.kt:103-133` writes `"seed ciphertext"` then asserts by **content** that `target-1700000000001.xml` still holds it after five quarantines — that's the property, not a file count. `WalletRepositoryImplTest.kt:198-221` asserts the fail-safe hinge with an empty event list, the onboarding pref unchanged at READY, and `coVerify(exactly = 0)` on erase. And `EncryptedPreferenceRecoveryTest.kt:71-201` now covers both API tiers with stand-ins, including transient-outranks-message.

## Nits, not blocking

- `EncryptedPreferenceQuarantine.kt:150-158` — if a quarantine ever moves only a `.bak` (xml absent mid-commit), the resulting `<filename>-<millis>.xml.bak` has no `.xml` sibling, so the `endsWith(".xml")` filter never sees it and it's never pruned. Genuinely unbounded, though it's a few KB and `purgeQuarantine()` clears it.
- Worth naming in the KDoc: on an unclassifiable permanent failure the user now sits on the splash with **no in-app recourse**, since Reset Zodl is behind it. That's the deliberate price of failing safe and it's the right default — but it should be written down next to `WalletRepository.kt:146-153` rather than discovered later.

LukasKorba (commented):
Both nits done, and the orphan fix is the right shape — `EncryptedPreferenceQuarantine.kt:160-165` collects base names from `.xml` and `.xml.bak` into a set and deletes both siblings per entry, so a `.bak` with no `.xml` is now reachable by the pruner while the oldest entry stays pinned. Retention semantics are unchanged.

The KDoc is better than what I asked for. I wanted the no-in-app-recourse consequence written down; you also supplied the fact that makes it defensible — that the crashed sharing coroutine is *the same terminal outcome the app already reaches today* via `SynchronizerProvider` collecting the same underlying store. That reframes it: the splash hang isn't a new failure mode this PR introduces, it's the existing one, and the alternative would be misrouting an intact-but-unreadable store to onboarding. That's the stronger argument and I hadn't had it in front of me.

Nothing further from me on this one. Signed off on the API 27-32 residual as discussed — the pre-33 message list is the thing to revisit first if a report ever comes in from that range.

LukasKorba (approved):
Approving.

Summary of what this sign-off covers, for the record: the three blocking items I raised are closed in code — the transient-Keystore veto is shared by both classifiers through one helper so they can't drift, `pruneQuarantine` pins the oldest (seed-bearing) entry indefinitely with no unbounded growth, and the quarantine-failure path no longer deletes the original. The erase failure reaches crash reporting, and all three test gaps are pinned by tests that assert the property rather than something adjacent — the quarantine one asserts by content that the seed survives five rounds.

Two things I'm explicitly signing off rather than silently accepting:

1. **The narrowed veto** (`hasTransientAndroidKeyStoreMarker` + `isPermanentAndroidKeyStoreFailure`). This is not the predicate I originally asked for. You were right that the bare version would have stranded the device-to-device transfer case on the splash forever, and reproducing it on API 31 and 35 was the right way to make that argument.
2. **The API 27-32 residual.** Below API 33 both reflective signals are absent, so the five-string message list is the only guard, and a transient failure carrying one of those messages would release the veto and both quarantine the seed and delete the master key. I judge this unlikely and the ~1.4s ladder covers the common case. If a report ever comes in from that API range, this is the first thing to look at.

PR #2500, opened 2026-09-08
Consume the Zodl SDK snapshot from Maven instead of an SDK source pin

nesence-m:
## What

Point the app's SDK dependencies at the Zodl fork's own Maven group and drop the SDK source pin.

- `settings.gradle.kts`
  - `zcash-sdk`, `zcash-sdk-backend`, `zcash-sdk-incubator` version-catalog aliases: `cash.z.ecc.android:*` → `com.zodl.android:*` (version still `$zcashSdkVersion`).
  - The three `substitute(module(...))` targets for the `SDK_INCLUDED_BUILD_PATH` composite build: same group switch, so the included-build path keeps substituting.
  - Both snapshot repositories (`oss.sonatype.org/content/repositories/snapshots`, `central.sonatype.com/repository/maven-snapshots`) gain `includeGroup("com.zodl.android")` alongside the existing `cash.z.ecc.android`.
- `gradle.properties`
  - `SDK_COMMIT_PIN=66701636e2ab9b28dd95807365135f7eb097e75f` → `SDK_COMMIT_PIN=` (blank).
  - Comment block above the pin gains a line saying what "blank" now resolves to.

No lockfile changes — see below. No CHANGELOG entry: nothing user-facing, and the file only carries user-facing prose.

## Why

The fork `zodl-inc/zodl-android-wallet-sdk` publishes under the Maven group `com.zodl.android` (hardcoded in its `zcash-sdk.publishing-conventions.gradle.kts`), not upstream's `cash.z.ecc.android`. The app still declared the upstream coordinates, so the upstream `3.1.1-SNAPSHOT` it resolved was stale and the only way to build against current SDK work was `SDK_COMMIT_PIN` — CI checking out the SDK at a commit and compiling it from source, Rust/NDK toolchain and all, on every job.

A snapshot `3.1.1-20260908.055811-1` of all five fork artifacts (`zcash-android-sdk`, `zcash-android-backend`, `zcash-android-sdk-incubator`, `lightwallet-client`, `zcash-android-sdk-slipstream`) was published today from `maint/v3.1.x` @ `66701636e2ab9b28dd95807365135f7eb097e75f` — the exact commit the pin named. The published snapshot therefore carries what the source build carried, and the pin has nothing left to add.

## Deliberately left on `cash.z.ecc.android`

[...]

PR #2503, opened 2026-09-08
Release 3.11.0

### LINKED ISSUE #2502: App Release 3.11.0
## Release checklist (zodl-android 3.11.0 from maint/v3.10.x)

- [ ] SDK 3.2.0-SNAPSHOT published and verified (content + Rust check)
- [ ] ZCASH_SDK_VERSION=3.2.0-SNAPSHOT, SDK_COMMIT_PIN blank
- [ ] Version bump, CHANGELOG, WHATS_NEW EN/ES, fastlane changelogs (notes from Andrea, approved in #wallet-team)
- [ ] candidate/v3.11.0 → release/v3.11.0 PR opened (merge only after Milan's double check)
- [ ] Tag 3.11.0-<versionCode> on the release/v3.11.0 merge commit
- [ ] release/v3.11.0 merged back into maint/v3.10.x

nesence-m:
Closes #2502

### Added:

- App Theme setting in Settings, letting you choose System, Light, or Dark — with a bottom sheet to pick
  Classic Dark or Pure Black (OLED) whenever a dark appearance applies.
- The in-app browser now matches the app theme, including the Pure Black theme.

### Changed:

- Automatic server selection now switches only to a server that is meaningfully faster than the one you're on (at least 200 ms and 25 % faster), or when the current server fails its health check twice in a row, so the wallet no longer flips between near-equal or briefly slow servers. A switch is never followed by another within thirty minutes, and the servers are re-checked at most once every ten minutes instead of on every return to the app.

### Fixed:

- Support/error reports for a failed transaction no longer claim a fake `gRPC: false, code: -1` status when the failure actually happened before submission (e.g. a Sapling parameter download failure); such reports now correctly identify the real exception instead.
- Opening Send immediately after a cold start no longer falls back to a loading screen while automatic server selection reconnects the wallet. The automatic winner waits for the local balance snapshot, which remains visible while the new server connection settles.
- Fixed a crash loop some wallets hit at startup when the SDK reported the stored seed as no longer relevant to its database. The app now recovers automatically by clearing the mismatched local data and rescanning, instead of crashing repeatedly.
- The app returns to onboarding so you can restore from your recovery phrase when its secret store is found empty while it still believes a
  wallet was set up, instead of opening a home screen with no wallet behind it.
- Opening the encrypted secret store is retried a few times before it is declared unreadable, and an unreadable store is set aside on the
  device instead of being deleted.
- Swap quotes that fail the request-vs-response validation now show a dedicated "Swap details didn't match" sheet with a Report button that pre-fills a support email (MOB-1340).


LukasKorba (commented):
Release mechanics are in good shape — the two things I'd have gone looking for first are already right. One defect in the CHANGELOG, and one timing hazard worth understanding before you merge.

## Fix before merge

**`CHANGELOG.md:10-45` — the 3.11.0 section has two sets of `### Added/Changed/Fixed`.** Lines 12-25 are the user-facing summary copied from whatsNew; lines 26-45 are the accumulated developer entries. Every item is stated twice in two registers under one heading.

Compare `f7e18d8a`, the 3.10.2 release-prep commit: it added only `## [3.10.2 (2569)] - 2026-08-27` above the entries that were already there, and the user-facing prose lived solely in `WHATS_NEW_*` and fastlane. Same treatment here means deleting the first `### Added` through the first `### Fixed` block. The PR body mirrors the duplication and wants the same trim.

## Worth knowing before you merge — the fastlane filename is timing-dependent

`app/build.gradle.kts:287` is:

```kotlin
output.versionCode.set(if (versionCodeOverride == 1) gitInfo.commitCount else versionCodeOverride)
```

With `ZCASH_VERSION_CODE=1`, the APK's versionCode is the **git commit count**, and the release notes Play reads are keyed by that number. This isn't inference — tag `3.10.2-2569` has a commit count of exactly 2569 and ships `2569.txt`. (The `gradle.properties:47-48` comment about Play ignoring VERSION_CODE refers to the *property*, not to the computed value the APK carries.)

`candidate/v3.11.0` is at **2652** today. Merging adds one commit → 2653, which is why `2653.txt` is right. But that holds only if:

1. **The merge produces exactly one commit.** A merge commit or a squash both land on 2653. A **rebase merge** replays all 83 commits onto the base and lands on 2652 — the APK would then look for `2652.txt`, find nothing, and ship with empty release notes.
2. **Nothing further is pushed to the candidate branch.** Since you're double-checking and merge is on hold, this is live: one more commit makes head 2653 and the merged result 2654, orphaning `2653.txt`.

Nothing in CI catches either case — the filename is only consulted by Play at publish time. If anything does land, rename both `2653.txt` files to match and update the `(2653)` in the CHANGELOG and whatsNew headings.

## Nit

The user-facing notes fold MOB-1340, MOB-1836 and MOB-1565 into "We made a range of security and privacy improvements." The swap-mismatch sheet is a new visible screen with a Report button — users who hit it will wonder what it is. Worth its own line in the Play notes.

## Checked and clean

**Base premise holds:** `compare 3.10.2-2569...release/v3.11.0` returns `identical`, so the 165-file diff really is the full delta since 3.10.2 and there's no surprise sitting in the base.

**Attribution is clean, which is where a real problem would have hidden.** The union of file lists across all 70 non-merge commits covers all 165 files with nothing unaccounted for, and the candidate branch's first-parent chain is nine PR merges plus exactly one release-prep commit — no code pushed straight to `candidate/` or `release/`. All nine PRs verified merged into `maint/v3.10.x` (#2444, #2351, #2489, #2490, #2494, #2495, #2497, #2499, #2500). `8156b185` touches only release-prep files.

**The two release gates I'd have checked first are both right:** `SDK_COMMIT_PIN` is blank — cleared by `89a94ded` in #2500, which is exactly the unpin I asked for on #2495 — and `check_properties` is green, which is the hard-fail gate the comment block describes. `ZCASH_SDK_VERSION=3.2.0-SNAPSHOT` matches what the SDK actually publishes (verified at SDK tag `v3.2.0`: `IS_SNAPSHOT=true`, `LIBRARY_VERSION=3.2.0`).

**The SDK groupId change was followed properly**, which was my main worry given it landed in the same SDK release: `settings.gradle.kts:251-253` declares `com.zodl.android:zcash-android-{sdk,backend,sdk-incubator}`, `:378-380` substitutes the same coordinates for the composite build, and `includeGroup("com.zodl.android")` is added to **both** snapshot repos at `:119` and `:130` — that second one matters, since a missed filter fails resolution silently. The only `cash.z.ecc.android` left is `kotlin-bip39`, a genuinely separate upstream artifact, correctly untouched.

No misfiled entries this time — `[3.10.2 (2569)]` at head is byte-identical to the same section at its tag, and everything from line 47 to EOF matches. `[Unreleased]` is correctly empty. EN and ES correspond item-for-item and the Spanish is genuinely translated, not stale from 3.10.2. No stale `3.10.x` strings anywhere, no debug flags, staging endpoints or release-gating TODOs in the delta.

`.github/actions/setup-sdk-companion/action.yml` repoints the companion checkout from `zcash/zcash-android-wallet-sdk` to `zodl-inc/zodl-android-wallet-sdk`; it's here because it landed via #2351 during the cycle, not as release-prep.

**CI:** 13 green including both `build`s, `check_properties`, all three static-analysis jobs, `test_kotlin_modules` and `e2e-android (phase 1)`. The three reds are the known infra failures — `test_android_modules_emulator (app)` shows `device offline` / `Failed to find ColorBuffer` / "Expected 12 tests, received 4", and the same job is red right now on `chore/consume-zodl-sdk-snapshot` (a properties-only chore) and on a pure merge branch; `Keystone` fails at `Unable to connect to adb daemon`; `e2e-android (phase 2)` is cancelled downstream of it. Nothing attributable to this release.

LukasKorba (approved):
Approving. Amending in place rather than adding a commit was the right call, and I checked that it actually worked rather than taking it on trust: head `447903495` still has a commit count of **2652**, so the merge lands on 2653 and `2653.txt` stays valid.

The release-prep commit's CHANGELOG contribution went `+16-0` → `+2-0`, which is now just the heading and a blank line — the same shape as `f7e18d8a` for 3.10.2. Every other file in that commit is unchanged (`WHATS_NEW_EN`/`_ES` +16-0 each, both `2653.txt` +10-0, `gradle.properties` +2-2), so CHANGELOG really was the only thing touched. The 3.11.0 section now carries a single set of `### Added/Changed/Fixed`.

**Two conditions this approval rests on, since neither is enforced by CI:**

1. **Merge with a merge commit or a squash — not a rebase.** A rebase merge replays all 83 commits and lands on 2652, so the APK would look for `2652.txt`, find nothing, and 3.11.0 would ship with empty release notes in Play.
2. **Nothing further pushed to `candidate/v3.11.0`.** One more commit makes head 2653 and the merged result 2654, orphaning `2653.txt`. If anything does land, rename both fastlane files and update the `(2653)` in the CHANGELOG and both whatsNew headings — and amending again, as you did here, avoids the problem entirely.

Everything from my first pass stands: base identical to `3.10.2-2569`, attribution clean with no code pushed straight to the release branches, `SDK_COMMIT_PIN` blank with `check_properties` green, `ZCASH_SDK_VERSION=3.2.0-SNAPSHOT` matching what the SDK publishes, and the `com.zodl.android` groupId followed through both snapshot-repo `includeGroup` filters.

CI at the time of writing: 7 green including `check_properties`, 0 failures, 8 still running. Given this round was markdown only I'm not holding for them, but confirm they land — and expect `Keystone`, `test_android_modules_emulator (app)` and `e2e-android (phase 2)` to come back red on the known emulator/adb infra problem, as they are across unrelated branches.

The nit from my first pass is still open and still optional: the Play notes fold the new "Swap details didn't match" sheet into "a range of security and privacy improvements". It's a visible new screen with a Report button — a line of its own would help users recognise it. Your call, not a blocker.

PR #2505, opened 2026-09-09
Merge maint/v3.8.x into maint/v3.9.x (MOB-1501 swap assets cascade)

nesence-m:
Forward-merge of `maint/v3.8.x` into `maint/v3.9.x`. True two-parent merge commit — please land with **Create a merge commit**, never squash.

**What it carries:** exactly the four commits of [#2363 [MOB-1501] feat: add DASH, BCH, ZEC (SOL, NEAR) to assets list](https://github.com/zodl-inc/zodl-android/pull/2363), which merged into `maint/v3.8.x` on 2026-09-04 after the 3.9 and 3.10 lines were already cut and was never cascaded forward. As a result v3.11.0 shipped without DASH, BCH, ZEC-on-SOL and ZEC-on-NEAR in the curated swap list (reported by Neal, [MOB-1501](https://linear.app/zodl/issue/MOB-1501/add-dash-bch-and-synthetic-zecs-to-swap-lists)).

`git log maint/v3.9.x..maint/v3.8.x` is empty after this merge; the merge is conflict-free. First link of the strict chain maint/v3.8.x → v3.9.x → v3.10.x → v3.11.x; once this lands, the next PR merges the updated `maint/v3.9.x` into `maint/v3.10.x`, then `maint/v3.10.x` into the newly cut `maint/v3.11.x`.

https://claude.ai/code/session_018siBR7Utdc9Sgd9BXxa4DB


LukasKorba (approved):
utACK

LukasKorba (commented):
Recording what I checked, since three more links follow this one and the same checks apply to each.

The branches are genuinely diverged — `compare maint/v3.8.x...maint/v3.9.x` is `ahead_by: 359, behind_by: 4` — so this was a real merge with real resolution, not a fast-forward. That's what makes it worth verifying rather than eyeballing:

- **Real merge commit**, two parents: `03593611a` (v3.9.x tip) and `1e28108e9` (#2363).
- **Nothing added by the resolution.** The merge commit's diff against its first parent is exactly the 4 files in the PR — no evil-merge extras.
- **Nothing dropped from #2363.** Its own net diff is the same 4 files with the same `+52-0`, so the cascade carried all of it.
- **Nothing on v3.9.x reverted.** Zero deletions, and every modification is `+N-0`. This is the failure mode I'd worry about most in a forward-merge from an older line.
- The DASH icons are absent from the diff because v3.9.x already carries them (`ic_chain_dash.png`/`ic_token_dash.png`, blob `b59cd5c44`) from an independent path — not a dropped file.
- `"ZEC" to "sol"` and `"ZEC" to "near"` looked like a duplicate-key hazard, but `allowedAssets` is `List<Pair<String, String>>` (`SimpleSwapAssetProvider.kt:18`), so repeated first elements are intended — `"USDC"` already appears four times across arb/sui/near/pol. Non-issue, noting it so the next reviewer doesn't re-derive it.
- Tests cover DASH, BCH, and both new ZEC chains.

**Agreed on merge-commit rather than squash, and it matters more than style.** Squashing collapses the two parents into one, so v3.8.x would never be recorded as an ancestor of v3.9.x — the next link (v3.9.x → v3.10.x) would then re-present these same changes as unmerged work, and the "tagged release branches are merged back" check (#2488) would keep flagging the line. The two-parent ancestry is the thing the cascade actually runs on.

**CI:** one `build` job failed on `The action 'Build Store APK' has timed out after 60 minutes` — not a compile error, and the identical tree passed in the sibling run (`34309024408` green vs `34309024412` timed out). Worth noting the emulator jobs are genuinely green here, including `test_android_modules_emulator (app)`, which has been red across the repo for days — so that infra appears to have recovered and this timeout is isolated rather than part of it.

PR #2508, opened 2026-09-09
MOB-1890: Warn before a sub-$300 swap or CrossPay payment

nesence-m:
Implements [MOB-1890](https://linear.app/zodl/issue/MOB-1890/implement-sub-dollar300-refund-warning-for-swap-and-crosspay). Android counterpart of zodl-ios#2107.

## Summary

NEAR does not refund a swap or payment under $300 when the loss is user error (wrong address, wrong network), and nothing in the app said so.

- **Warning sheet.** Tapping the primary CTA on Swap-to-ZEC, Swap-from-ZEC or CrossPay with an estimated amount-in under $300 opens a "No Refunds Under $300" sheet with Cancel, Continue and a "Don't show this message again" checkbox. The threshold is evaluated at tap time on the pre-quote USD estimate (agreed with product in Slack: instant explainer, boundary cases handled by NEAR support); an unknown price fails safe and shows the sheet. Continue re-enters the existing quote flow unchanged, Cancel persists nothing.
- **Suppression.** One flag per surface (as product asked, same as iOS): silencing it on Swap leaves it armed for CrossPay. The flag is written only after a Continue reaches the awaiting caller, so a sheet restored after process death or a Cancel racing the write cannot silence the warning. Reset Zodl clears the flags.
- **Explainers.** The Swap, CrossPay and Refund Address explainers state the threshold. Swap and CrossPay gain a Learn more button that closes the sheet and opens the support article in the in-app browser. All three close with Dismiss. English and Spanish copy match iOS.
- **Design.** The shared info-sheet shell now draws its secondary button with the theme border and the 8dp gap from Figma (also reaches the swap mismatch sheet), and the warning uses the Figma alert icon with a night variant.
- **Hardening.** Re-entrancy guard in the use case instead of raising the requesting flag while the sheet is open, swap direction snapshotted once per request and the direction toggle disabled while a quote is in flight, remaining nullable USD price sites in the swap mapper null-guarded.

Implementation: `NavigateToSwapRefundWarningUseCase` (gate + suspending result pipeline, same shape as `NavigateToSlippageUseCase`), `screen/swap/refundwarning/` sheet, `IsSwapRefundWarningDismissedStorageProvider`, one-line hooks in `SwapVM` and `PayVM`.

## Test plan

- [x] `ui-lib` unit tests: 560 pass (new `NavigateToSwapRefundWarningUseCaseTest`, `SwapRefundWarningVMTest`, plus gate/mode tests in `SwapVMTest` and `PayVMTest`)
- [x] `ktlint` and `detektAll` clean
- [x] `assembleZcashmainnetStoreDebug` builds; APK native-lib check passes
- [x] Emulator (Pixel 9 Pro XL, funded mainnet test wallet): sheet on all three surfaces under $300 with "swaps"/"payments" wording; Cancel keeps the entered values and persists nothing even with the box ticked; Continue reaches Review Quote / the quote request; box + Continue silences that surface only and survives an app restart; explainers show the new copy, Learn more opens the correct support articles; light + dark, English + es-419
- [ ] Reviewer: run the app and try the changes

> **Note**

# Author

> **Note**

# Reviewer

https://claude.ai/code/session_013K5WPsS4evGUbPTdyxZCK7


ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/common/InfoBottomSheetView.kt
```diff
@@ -75,9 +76,11 @@ fun InfoBottomSheetView(
...
                             state = it,
                             modifier = Modifier.fillMaxWidth(),
-                            defaultPrimaryColors = ZashiButtonDefaults.secondaryColors(),
+                            defaultPrimaryColors =
+                                ZashiButtonDefaults.secondaryColors(
+                                    borderColor = ZashiColors.Btns.Secondary.btnSecondaryBorder
+                                ),
                         )
-                        Spacer(12.dp)
                     }
```
noop-sk, line 84: The `Spacer(12.dp)` that used to sit between the secondary and primary buttons was deleted and nothing replaced it. The PR description says this change should give "the 8dp gap from Figma," but as written the two buttons now render flush with a 0dp gap — no `Spacer` here and no `verticalArrangement` spacing on the enclosing `Column`. This hits every sheet with both a primary and secondary button through `InfoBottomSheetView`, including the two screens this same PR adds a secondary "Learn more" button to (`SwapInfoView`, `PayInfoView`). Please add `Spacer(8.dp)` back in (or equivalent `Arrangement.spacedBy(8.dp)`) before merge.

noop-sk (changes requested):
Reviewed the diff (threshold/currency handling, boundary case, null-price fail-safe, bypass surface, re-entrancy guard, suppression-flag scoping, i18n, test coverage) — all sound. One concrete blocking regression found, inline below.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/common/InfoBottomSheetView.kt
nesence-m: Fixed in 5b8860a48: `Spacer(8.dp)` is back between the secondary and primary buttons, scoped inside the `secondaryButton?.let` block so primary-only sheets are unchanged. Also merged maint/v3.11.x in (ce903d908) to clear the CHANGELOG conflict from #2509.

noop-sk (dismissed):
Re-verified the fix. `5b8860a48` restores the spacer correctly: `Spacer(8.dp)` at `InfoBottomSheetView.kt:84`, exactly 8dp as claimed, correctly positioned between the secondary and primary buttons and scoped to sheets that have a secondary button, so primary-only sheets are unaffected. Both `SwapInfoView` and `PayInfoView` inherit the fix through the shared component. The `maint/v3.11.x` merge resolved the CHANGELOG conflict cleanly — no conflict markers, no duplicated entries. Nothing new introduced by either the fix commit or the merge. CI green on everything completed.

ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/common/InfoBottomSheetView.kt
nesence-m: Reverted that in 2e9398d47 after checking with Milan: the buttons space themselves, so no explicit `Spacer` is needed between the secondary and primary button. `InfoBottomSheetView` is back to the state originally in this PR.

noop-sk (changes requested):
The one new commit (2e9398d47, "Revert MOB-1890: restore the 8 dp gap...") reverts the exact spacer fix from the previous round, which was already confirmed and approved. This is a real regression, not a trivial change — flagging it as blocking.

The revert's stated rationale ("the buttons already space themselves; no explicit spacer is needed") does not hold up. I checked the actual layout: the two `ZashiButton`s sit as direct siblings in a plain `Column` with no `verticalArrangement` spacing (defaults to `Arrangement.Top`, i.e. zero gap), and `ZashiButton` itself applies no default outer margin. With the `Spacer(8.dp)` removed, the buttons render flush again — the same visual bug from round 1, now reintroduced.

Please restore the `Spacer(8.dp)` in `ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/common/InfoBottomSheetView.kt` (right after the secondary button block, before the primary button). If there's a reason to believe the buttons do space themselves, happy to take another look at a repro, but the Compose code as written doesn't support that.

Also, unrelated but worth a look before merge: three CI checks are currently red on this head — `Keystone`, `test_android_modules_emulator (app)`, `test_robo_release`.

noop-sk (changes requested):
Checked the touch-target claim against the actual source, not taking it at face value given the prior round's rationale also didn't hold.

The mechanism described (Material3's 48dp invisible touch-target padding) is real, but it's not what `Button` uses. That's `minimumInteractiveComponentSize()`/`LocalMinimumInteractiveComponentEnforcement`, used by compact controls like `IconButton`, `Checkbox`, `RadioButton`, `Switch` — not `Button`. Reading `Button.kt` from this repo's pinned M3 version directly:

```kotlin
Surface(onClick = onClick, modifier = modifier.semantics { role = Role.Button }, ...) {
    ProvideContentColorTextStyle(...) {
        Row(
            Modifier.defaultMinSize(minWidth = ButtonDefaults.MinWidth, minHeight = ButtonDefaults.MinHeight)
                .padding(contentPadding),
            ...
        )
    }
}
```

`defaultMinSize(minHeight = ...)` is applied to the Row that's the direct, only child of `Surface` — it sets the actual *visible* bounds of the button's surface, not an invisible buffer around a smaller visible element. There's no `minimumInteractiveComponentSize()` call anywhere in `Button.kt`.

On top of that, `ZashiButtonDefaults.contentPadding` overrides Material3's default with `PaddingValues(horizontal = 10.dp)` — vertical padding is explicitly zero, and `ZashiButton.kt` adds no height/min-size modifier of its own beyond `.pressMorph(...)` (confirmed to be a visual-only `graphicsLayer` scale transform, no layout effect).

So with two `ZashiButton`s as siblings in a `Column` with no `verticalArrangement` spacing (still the case at this head) and no `Spacer` between them, they render with their visible bounds genuinely flush — same 0dp-gap defect as round 1. No screenshot or UI test in this PR pins the claimed 8dp gap either.

Please restore `Spacer(8.dp)` in `InfoBottomSheetView.kt` between the secondary and primary buttons. If there's a screenshot or repro showing an actual 8dp visual gap without it, genuinely happy to take another look — but the component source as written doesn't produce that.

noop-sk (approved):
Approving based on manual verification on a real device by the app's maintainer — confirmed in person with Milan that the buttons render with correct spacing at this head, without the Spacer.

For the record: my source-level read of Button.kt/ZashiButton.kt suggested the visible bounds should be flush with no spacer, and I couldn't find a mechanism in the M3 Button/ZashiButton code that would produce a gap. That didn't match what real-device testing shows, so there's something about the actual rendered/measured layout (ripple/indication bounds, density-driven touch-target behavior, or something else not visible from static source reading) that the code-only analysis missed. Real-device verification is the more reliable signal here — deferring to it.

Given how much back-and-forth this one line caused, a screenshot test pinning the current (no-spacer) button layout in InfoBottomSheetView would be a cheap way to prevent this from flip-flopping again, if convenient.

PR #2509, opened 2026-09-09
Add Tor state to the support email

nesence-m:
Closes [MOB-1323](https://linear.app/zodl/issue/MOB-1323/add-tor-state-into-debug-email).

## Summary

- The support/feedback email now includes a `Tor enabled: Yes` / `No` / `Not set` line, read from `IsTorEnabledStorageProvider`, placed after the environment block and before permissions — the same position as the iOS `SupportDataGenerator` `TorItem`.
- New `TorInfo` support model; `SupportInfo` gained a `Tor` type and field; `GetSupportUseCase` and `SupportViewModel` take the provider via constructor (resolved by the existing `factoryOf` / `viewModelOf` registrations, no DI module changes).
- `Not set` is reported when the user has never made the Tor choice (iOS collapses that case to `No`); the app treats it as Tor off.
- CHANGELOG entry under Unreleased.

## Test plan

- `./gradlew :zashi-android:app:compileZcashmainnetStoreDebugKotlin :zashi-android:ktlint :zashi-android:detektAll` from the task workspace root — BUILD SUCCESSFUL.
- `SupportFinancialInfoStateTest` (instrumented) updated with a fake Tor provider and four new tests: the include/exclude filter for the new type plus one per rendered value. Not run on a device; note the ui-lib androidTest source set currently fails to compile on `maint/v3.11.x` in unrelated fixtures (`MockSynchronizer`, `ChooseServerSelectionTest`) that lag the SDK interfaces — that predates this branch.

https://claude.ai/code/session_01MLpPj2TxQ8FnaLFRQ2kXxf

### DIFF
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetSupportUseCase.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetSupportUseCase.kt
@@ -2,11 +2,14 @@ package co.electriccoin.zcash.ui.common.usecase
 
 import android.content.Context
 import co.electriccoin.zcash.configuration.api.ConfigurationProvider
+import co.electriccoin.zcash.ui.common.provider.IsTorEnabledStorageProvider
 import co.electriccoin.zcash.ui.screen.support.model.SupportInfo
 
 class GetSupportUseCase(
     private val context: Context,
-    private val androidConfigurationProvider: ConfigurationProvider
+    private val androidConfigurationProvider: ConfigurationProvider,
+    private val isTorEnabledStorageProvider: IsTorEnabledStorageProvider
 ) {
-    suspend operator fun invoke() = SupportInfo.new(context, androidConfigurationProvider)
+    suspend operator fun invoke() =
+        SupportInfo.new(context, androidConfigurationProvider, isTorEnabledStorageProvider)
 }
--- a/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetSupportUseCase.kt
+++ b/ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetSupportUseCase.kt
@@ -11,5 +11,5 @@ class GetSupportUseCase(
     private val isTorEnabledStorageProvider: IsTorEnabledStorageProvider
 ) {
     suspend operator fun invoke() =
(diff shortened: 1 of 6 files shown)


ui-lib/src/main/java/co/electriccoin/zcash/ui/common/usecase/GetSupportUseCase.kt
```diff
@@ -2,11 +2,14 @@ package co.electriccoin.zcash.ui.common.usecase
...
 
 class GetSupportUseCase(
     private val context: Context,
-    private val androidConfigurationProvider: ConfigurationProvider
+    private val androidConfigurationProvider: ConfigurationProvider,
+    private val isTorEnabledStorageProvider: IsTorEnabledStorageProvider
 ) {
-    suspend operator fun invoke() = SupportInfo.new(context, androidConfigurationProvider)
+    suspend operator fun invoke() =
+        SupportInfo.new(context, androidConfigurationProvider, isTorEnabledStorageProvider)
```
nesence-m: never pass IsTorEnabledStorageProvider like this -> pass boolean instead

nesence-m: Fixed in 06f75dfde: SupportInfo.new and TorInfo now take a Boolean?; GetSupportUseCase and SupportViewModel read the provider and pass the resolved value down.

noop-sk (approved):
Reviewed the diff — Tor-state read is a one-shot suspend read of a persisted user setting, not a live bootstrap/race-prone value, so no staleness concern. DI wiring picks up the new constructor param automatically, no other call sites break. `TorInfo.toSupportString()` only emits a tri-state Enabled/Disabled/Not-set string — no PII leak.

Two non-blocking notes, not gating:
1. `SendEmailUseCase.kt:184-191` — the swap-issue support email hardcodes its `SupportInfoType` set instead of using `.entries.toSet()` like the other 4 call sites, so `Tor` won't appear there. Likely intentional/out of scope, but worth a quick confirm.
2. New test helper `newSupportInfo`'s `isTorEnabled: Boolean? = true` default is a bit arbitrary but harmless — all three Tor states are covered by dedicated tests.

CI failures on this head (Keystone, e2e phase 2, test_android_modules_emulator) trace back to unrelated emulator/instrumentation infra flake, not this diff.

PR #2512, opened 2026-09-09
MOB-1707: Offer all CMC-supported fiat currencies in Currency Conversion

nesence-m:
Closes [MOB-1707](https://linear.app/zodl/issue/MOB-1707/add-all-supported-currencies-to-the-currency-conversion-feature-except).

## Summary

- The Currency Conversion picker no longer shows a hard-coded list of 23 currencies. It fetches CoinMarketCap's `GET /v1/fiat/map` over the Tor-only client when the sheet opens (93 currencies today, kept in the provider's order, USD first) and caches the result in memory for the process.
- **CUP, IRR and RUB are excluded** (`EXCLUDED_FIAT_CURRENCY_CODES` in `FiatCurrencyDataSource`). Codes the platform currency table cannot name are dropped as well, so every row can render a localized name.
- While the list loads the sheet shows the same shimmer as the swap asset picker; on failure it shows the common error screen with a **Try again** button that refetches. A failed retry never wipes an already loaded list.
- New layers follow the existing ones: `CMCApiProvider.getFiatMap` → `FiatCurrencyDataSource` → `FiatCurrencyRepository` → `CurrencyConversionPickerVM`. The picker state gained a sealed `Loading / Success / Error` slot; rows, selection and the settings/opt-in screens are unchanged. No search field, no prefetch (decided with Milan).
- New EN/ES string `currencyConversion_pickerErrorDesc`; `FiatCurrencyExt.kt` and its test are removed; changelog entry added.

## Notes for reviewers

- The shimmer rows and the sheet background are both `bgSecondary` since #2436, so the loading state is very low-contrast on every sheet (the swap asset picker included). Left as is here; worth a small follow-up on `CommonShimmerLoadingScreen`.
- The Tor client has no HTTP timeout, so offline the sheet stays in the loading state until the Tor connect times out (about one to two minutes on the emulator) before the error appears.

## Test plan

- [x] `ktlint` and `detektAll` clean.
- [x] `:ui-lib:testZcashmainnetStoreDebugUnitTest`: 456 tests pass, including the new `CMCApiProviderTest.getFiatMapUsesTorClientAndNeverTheDirectClient`, `FiatCurrencyDataSourceImplTest` (exclusions, ordering, validity filter, blank key, empty response, error propagation), `FiatCurrencyRepositoryImplTest` (load once, in-flight dedupe, loading flag, error/cache semantics) and the rewritten `CurrencyConversionPickerVMTest`.
- [x] `assembleZcashmainnetStoreDebug` built against the pinned companion SDK (fbacf6f6).
- [x] Emulator (Pixel 9 Pro XL, Tor Protection on): Settings → Currency Conversion → Select currency shows the loading sheet, then all 90 currencies in CMC order matching the live API minus the three exclusions; reopening is instant from cache; picking CZK and saving shows the balance in CZK on Home.
- [x] Emulator offline: the sheet reaches the error state; after reconnecting, Try again reloads the list.

https://claude.ai/code/session_018R7BTgKGo5ZtY94Fhi72Th


noop-sk (approved):
Reviewed the diff — threshold/currency-count expansion (23 → ~90 fiat currencies) is correctly wired: no dangling references to the deleted `supportedFiatCurrencies`/`FiatCurrencyExt.kt`, `FiatCurrencyRepositoryImpl`/`CurrencyConversionPickerVM` dedupe/refresh/error-mapping logic checked for races and found sound, i18n strings present in both locales. No search/filter on the longer list is a documented, deliberate product decision (no search field, no prefetch — decided with Milan), and the LazyColumn implementation makes that a non-issue either way.

Non-blocking notes only:
1. `CurrencyFormatterExt.kt:22-38` — the generic `DecimalFormat` formatter doesn't follow each currency's actual ISO-4217 fraction-digit convention (e.g. 0 for JPY, 3 for KWD/BHD). Pre-existing, not touched by this PR, but expanding to ~90 currencies makes hitting one of these more likely — worth a follow-up ticket.
2. `FiatCurrencyDataSource.kt:132-136` — `isAlpha3Code` check alongside `Currency.getInstance(...).isSuccess` is likely redundant (the latter already throws for non-ISO codes), harmless as-is.

CI failures on this head (Keystone, test_android_modules_emulator, enable-auto-merge) are unrelated emulator/instrumentation infra flake and the expected pre-approval auto-merge gate, not caused by this diff.

PR #2519, opened 2026-09-15
Release 3.13.0

nesence-m:
Closes #2518

### Added:

- The support email now includes whether Tor is enabled.
- Swap and CrossPay now warn you before requesting a quote when the amount you are sending is worth less than
  $300, because NEAR does not refund a swap or payment under $300 that is lost to a wrong address or a wrong
  network. The warning carries a "Don't show this message again" checkbox that silences it for that flow only
  (MOB-1890).

### Changed:

- The Swap, CrossPay and refund-address explainers no longer describe NEAR refunds as unconditional; they now
  state the $300 threshold. The Swap and CrossPay explainers also gained a "Learn more" button that opens the
  matching support article, and all three explainers now close with a "Dismiss" button instead of "OK", matching
  iOS (MOB-1890).
- Currency Conversion now offers every fiat currency supported by the exchange-rate provider except CUP, IRR and RUB. The currency picker loads
  the list on demand and shows a loading state, or an error with retry when the list cannot be fetched.
- Currency Conversion now reaches the exchange-rate provider over Tor even when Tor is turned off in Settings
  and Currency Conversion has not been enabled yet, so the currency picker loads its list on first use instead
  of showing an error.

### Fixed:

- The seed backup prompt now always takes priority over the migration banner and any sync/connectivity error banner once your wallet has received a balance and you haven't backed up your recovery phrase yet, so it can no longer be hidden behind other home screen messages for days.
- The loading placeholders on the currency and swap-asset picker sheets are visible again; they used the same
  colour as the sheet background.

### Removed:

- The "Disable Tor?" dialog for a failed Tor start-up. Tor starts lazily, so the condition it reported could
  never occur.

### Not in this release

- MOB-1751 cross-chain payment URIs (#2457) is reverted on this line only (62214c2da): product wants more testing first, so it stays on main for a later release. The cascade back into main must revert the revert.

### Rebuilt at 2705

The candidate was rebuilt once more on top of the revert with commit 58ab0f164: SDK bumped to 3.2.1-SNAPSHOT (zodl-inc/zodl-android-wallet-sdk#19, the Slipstream engine always provides a lazy Tor client so the currency picker loads with Tor off), the picker sheets' loading placeholders given a visible colour, and the unreachable "Disable Tor?" dialog removed. Version code moved 2703 -> 2704 -> 2705 with these rebuilds. Firebase internal build 2705 already carries this exact content.

PR #2516, opened 2026-09-11
Merge maint/v3.11.x into main

nesence-m:
Forward cascade of the whole 3.11.x line into main (release/3.10.x, 3.11.0 and everything merged into maint/v3.11.x since, currently 203 commits). True merge commit, not a squash, so the next cascade does not replay phantom conflicts.

Opened as a draft on purpose: the draft shield keeps the auto-squash workflow from arming. It will be marked ready and merged with a merge commit in one step once #2508 (MOB-1890) has landed on maint/v3.11.x and been folded in here.

## Conflict resolution

- `gradle.properties`, `.github/actions/setup-sdk-companion/action.yml`, `.github/workflows/pull-request.yml`: maint side taken. `SDK_COMMIT_PIN` is now blank on main and `ZCASH_SDK_VERSION` is `3.2.0-SNAPSHOT`; the commit main was pinned to (fbacf6f6, the MOB-1751 URI parser) is an ancestor of the SDK's `v3.2.0` tag, so main's cross-chain URI code builds against the published snapshot.
- `AutomaticServerRepository.kt` and its test: both sides added code at the same spot. main's `isServerCustom` (from #2448) is kept next to maint's server-switch hysteresis lane (MOB-1832); both test groups are kept.
- `CHANGELOG.md`: git filed main's MOB-1707 entry under the shipped 3.10.0 heading; moved back under Unreleased inside the merge commit. The other main Unreleased lines had shipped in 3.10.x and now sit under those headings.

## Gates on the merge commit

- `ktlint` + `detektAll`: exit 0.
- `:ui-lib:testZcashmainnetStoreDebugUnitTest`: 585 tests, 0 failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016YiMjC34rcp1cTZvex8n3q

PR #2522, opened 2026-09-15
Merge maint/v3.13.x into main

nesence-m:
Forward cascade after release 3.13.0 (2705): `release/v3.13.0` = `maint/v3.13.x` (tip bce64e3b2) → `main`.

## What it carries

- `89ec493ba` true merge of `maint/v3.13.x`: the release prep commit (version 3.13.0, code 2705, changelog promotion, WHATS_NEW and fastlane notes), the follow-up commit `58ab0f164` (SDK 3.2.1-SNAPSHOT, visible shimmer on the picker sheets, dead "Disable Tor?" dialog removed), and the MOB-1751 revert that kept the payment URI parser out of the release.
- `bad6942cf` reverts that revert so **main keeps MOB-1751** (#2457), and files the CHANGELOG entry the feature never had under `[Unreleased]`.

Net against `main`: 17 files, only the follow-ups, the prep commit and the new changelog entry; the MOB-1751 sources are byte-identical to main. `maint/v3.13.x` is an ancestor of the head.

## Landing

Opened as a **draft on purpose**: `auto-merge.yml` on `main` would arm a squash merge, and a squashed cascade breaks the next one (#2408, #2423). After approval: `gh pr ready` then immediately `gh pr merge --merge`. Do not push to this branch after opening.

## Gates

ktlint, detekt and the app debug compile pass on the head (exit 0).


noop-sk (commented):
Reviewed the forward cascade maint/v3.13.x -> main. Everything checks out:

- True merge confirmed: merge commit 89ec493ba6 has two parents (prior main tip + maint/v3.13.x's actual current tip, matches the branch API exactly) - real merge history preserved, not a squash/rebase.
- MOB-1751 revert-of-revert verified genuinely symmetric: diffed the restoration commit against the original revert, all 14 files are inverted add/remove exactly - CrossPayRequest.kt and its tests re-added, NearSwapDataSource/SwapAsset/PayVM/ScanGenericAddressVM/SwapVM and their tests all modified back. Nothing still missing.
- CHANGELOG Unreleased entry is real and substantive, correctly positioned above the already-released 3.13.0 section, accurately describes the restored payment-URI feature.
- Diff vs main is exactly as narrow as described - 17 files changed, 6 commits ahead, 0 behind, matching the PR body precisely. None of the MOB-1751/CrossPay files appear in the diff (confirms the two reverts cancel out cleanly against main).
- Draft-as-auto-squash-shield is a sensible, deliberate practice, not just asserted - the PR body cites the specific auto-merge.yml hazard and prior incidents.

No blocking issues found. CI: all completed checks pass, a handful (build, both e2e-android phases, Keystone, emulator app test) still in progress with nothing red yet - worth letting finish before this goes ready+merge, but nothing here to flag as broken. Still draft, staying at COMMENT.

noop-sk (commented):
Checked the CI-settled claim carefully since it's specifically compared to #2519's profile, and it doesn't hold up for 2 of the 3 red checks - flagging this rather than approving, since it needs your eyes before we call it "same as before":

- **test_android_modules_emulator (app)**: genuinely matches #2519 - same ColorBuffer GPU errors, same 4/12-tests-completed, same device-offline crash.
- **Keystone**: does NOT match #2519's failure. #2519 failed with a Maestro "Create new wallet is visible" assertion on the onboarding screen. This PR's Keystone run gets past the entire connect flow cleanly and instead fails much later waiting for sync: "[Sync] Wait for wallet sync to complete" times out after a ~10.5 minute wait loop, "home.syncComplete" never becomes visible. Different failure point, different assertion.
- **e2e-android (phase 2)**: does NOT match the established 45-minute-timeout pattern either. This run's conclusion is `failure`, not `cancelled`, and it finished in 38m51s - never hit the 45-min cap. The actual failures are real assertion failures deep in the transaction flows: "Confirm" not visible and "Amount" not visible, repeated across crosspay-transaction.yaml, swap-transaction.yaml, and swap-receive-usdc-to-zec.yaml, running to completion before Maestro exits with status 1.

Worth noting: the two genuinely-different failures (Keystone sync stall, and the e2e Confirm/Amount assertion failures) are both in the CrossPay/swap payment-flow area - exactly what this PR restores via the MOB-1751 revert-of-revert. That could be coincidental infra flakiness on a heavier test run, or it could be a real regression surfacing now that the feature is back on main. Worth a look before merging rather than assuming it's the known backlog.

nesence-m:
Checked the two failures you flagged against runs on both sides of the MOB-1751 line. Neither correlates with the feature being present.

| E2E Smoke Tests run | Branch | MOB-1751 on branch | Keystone failure point | e2e phase 2 |
|---|---|---|---|---|
| 34935601752, 09-15 06:08 | candidate/v3.13.0 | **absent** (reverted) | `home.syncComplete is visible` | cancelled at cap |
| 34942534632, 09-15 07:36 | candidate/v3.13.0 | **absent** (reverted) | `home.syncComplete is visible` | failure: `"Confirm" is visible` ×5 |
| 34956499087, 09-15 10:10 | feature/chp-submission-speed | present (main-based) | `home.syncComplete is visible` | cancelled at cap |
| 34971816840, 09-15 12:55 | candidate/v3.13.0 | **absent** (reverted) | `"Create new wallet" is visible` | cancelled at cap |
| 34990012584, 09-15 15:40 | this PR | present | `home.syncComplete is visible` | failure: `"Confirm" is visible` ×5, `"Amount" is visible` ×1 |

- **Keystone:** the sync-wait stall on `home.syncComplete` is the dominant Keystone failure this week. It hit two candidate runs that had MOB-1751 reverted (06:08 and 07:36) and a main-based branch where the feature has been since 27 August. The onboarding-screen `Create new wallet` failure you compared against is the 12:55 run, the outlier, not the baseline.
- **e2e phase 2:** the five `"Confirm" is visible` failures across the CrossPay and swap flows are identical on the 07:36 candidate run, which did not contain MOB-1751 at all. The only delta on this PR is one additional `"Amount" is visible` assertion inside flows that were already failing on `Confirm`; the CrossPay e2e has been red on every branch since 2026-09-08. Whether a run reaches assertion failures or is cancelled at the 45-minute cap has varied run to run all week.
- Structurally, the MOB-1751 sources on this head are byte-identical to `main`, where they have been exercised by every PR run since 27 August, including the #2516 cascade that produced the `main` this branch starts from.

So the failures are the existing backlog, not a regression surfacing with the restore. The one thing I would not claim is that the `Amount` assertion is fully explained: it is one extra step inside an already-failing flow, and the flows deserve a real fix on main regardless.

noop-sk (commented):
Thanks for the run-by-run table — independently re-pulled the job logs to verify rather than take the comparison at face value, and the core conclusion holds: **the Keystone and e2e-phase2 failures are pre-existing backlog, not a regression from this PR's MOB-1751 restore.**

Two things worth flagging before merge, neither blocking my conclusion above:

1. **The evidence table has the MOB-1751 presence/absence flipped for two rows.** Checked the actual source tree at each cited commit:
   - `34935601752` (06:08, `ea69a53a`) — table says "absent", tree actually **has** `CrossPayRequestParser` present.
   - `34956499087` (10:10, `260dd3b6`, feature/chp-submission-speed) — table says "present", tree actually **lacks** it entirely.
   
   This doesn't change the verdict — if anything it strengthens it: the 06:08 run shows the identical ~10.5min `home.syncComplete` Keystone stall *with* MOB-1751 present on a commit unrelated to this PR, which is better "feature-independent" evidence than the table as written. But worth correcting the table so it doesn't mislead anyone re-deriving this later.

2. **The `Assert that "Amount" is visible"` failure is a real, unexplained delta.** Confirmed via job logs: the 07:36 baseline run (no MOB-1751) completes that assertion; this PR's run (34990012584) fails it once before passing on retry. The 5× `"Confirm" is visible"` failures match signature-for-signature across both runs (same steps, same ~5min elapsed), so that part is solid pre-existing flake. The `Amount` blip is small and inside an already-flaky flow, but it's not proven harmless — agree with your own hedge on this one.

Also minor: "present since 27 August" — `CrossPayRequest.kt`'s oldest commit on `main` is 2026-08-20, and #2516 (the cascade that brought it to `main`) merged 2026-09-11, not late August. Doesn't affect the substance.

Still draft, so leaving this as COMMENT rather than approving — but the regression concern itself is resolved. Once ready for review (or if you want a second pass on the `Amount` blip specifically), happy to take another look.

nesence-m:
Both corrections are right, thank you. Corrected table:

| Run | Branch @ head | MOB-1751 on branch | Keystone failure point | e2e phase 2 |
|---|---|---|---|---|
| 34935601752, 06:08 | candidate/v3.13.0 @ ea69a53a7 (first cut, before the revert existed) | **present** | `home.syncComplete` stall | cancelled at cap |
| 34942534632, 07:36 | candidate/v3.13.0 @ 4d2d1f9b2 (rebuilt with the revert) | **absent** | `home.syncComplete` stall | `"Confirm"` ×5 |
| 34956499087, 10:10 | feature/chp-submission-speed @ 260dd3b6 (based on maint/v3.11.x) | **absent** | `home.syncComplete` stall | cancelled at cap |
| 34971816840, 12:55 | candidate/v3.13.0 @ 1817046f6 | **absent** | `"Create new wallet"` | cancelled at cap |
| 34990012584, 15:40 | this PR @ bad6942cf | **present** | `home.syncComplete` stall | `"Confirm"` ×5, `"Amount"` ×1 |

So the Keystone sync stall appears with and without the feature, on three different branches, and the five `Confirm` failures appear without it. On the date: #2457 merged directly into `main` on 2026-08-27 (`e696f608b`); the file's older commits are from the feature branch. #2516 was the maint cascade, unrelated to how MOB-1751 reached main.

On the `Amount` blip: agreed it is not proven harmless. It failed once and passed on retry inside a flow that already fails on `Confirm` every run, so I would not hold the cascade on it, but it goes on the list for the e2e fix on main alongside the `Confirm` failures and the emulator host crash.

Marking the PR ready for review now so it can be approved; no further pushes will land on this branch, and the merge will be a merge commit immediately after approval.

noop-sk (approved):
Corrected table verified — both flipped rows now match the source tree (06:08 = present, 10:10 = absent), and the corrected data makes the case even more clearly: the `home.syncComplete` Keystone stall reproduces across three distinct branches regardless of MOB-1751's presence, and the 5x `"Confirm" is visible` e2e failures reproduce identically without it either. Regression concern fully resolved.

No longer draft, head sha unchanged since the last round (bad6942c), no new pushes, mergeable. The 3 red checks (`test_android_modules_emulator (app)`, `Keystone`, `e2e-android (phase 2)`) are the same confirmed pre-existing backlog, nothing new.

One acknowledged, non-blocking loose end carried forward per your own comment: the `Assert that "Amount" is visible"` blip is still unexplained, but you've already flagged it for the e2e fix on `main` rather than holding this cascade on it — agreed that's the right call, it's inside an already-flaky flow and not specific to this PR's diff.

PR #2526, opened 2026-09-16
Render ZAP1 attestation memos as typed event cards

### LINKED ISSUE #2172: Feature: render ZAP1 attestation memos as typed event cards
## What

ZAP1 is an open attestation protocol that writes structured lifecycle events to Zcash shielded memos. There are currently 5 Merkle roots anchored on mainnet, 23 leaves, and 15 event types deployed.

Zodl already decrypts and displays these memos - but they show as raw strings like `ZAP1:04:f265b9a06a61b2b8...`. This proposal adds formatted rendering so users see what the attestation actually means.

## Before / After

| Current | Proposed |
|---|---|
| `ZAP1:01:075b00df2860...` | **PROGRAM_ENTRY** - participant enrolled `075b00df...` \[Verify\] |
| `ZAP1:04:f265b9a06a61...` | **DEPLOYMENT** - machine activated `f265b9a0...` \[Verify\] |
| `ZAP1:09:024e36515ea3...` | **MERKLE_ROOT** - tree anchored at block 3,286,631 `024e3651...` \[Verify\] |
| `ZAP1:0d:a487c25f5867...` | **GOVERNANCE_PROPOSAL** - proposal submitted `a487c25f...` \[Verify\] |
| Non-ZAP1 memos | Unchanged |

The Verify link opens the proof page where anyone can independently check the Merkle path back to the Zcash anchor transaction.

## Scope

Transaction detail rendering only. No send flow changes. No sync changes. No new permissions.

**Detection:** check if the decrypted memo string matches `^(ZAP1|NSM1):[0-9a-f]{2}:[0-9a-f]{64}$`

**Rendering:** replace the raw string with an attestation card showing event label, short hash, copy action, and verify link.

**Fallback:** malformed ZAP1-like strings render as plain text (current behavior).

`NSM1:` is the legacy prefix - same format, accepted during decode for backward compatibility.

## Integration point

The memo string is already resolved before it reaches the UI. The change is at the rendering boundary:

- `TransactionDetailVM.kt` lines 273-275 and 315-317: `content = stringRes(memo)` - intercept here
- New file `Zap1MemoFormatter.kt`: prefix detection + type lookup (~30 lines)
- `TransactionDetailInfoMemo.kt`: attestation card variant

## Parser (complete, 30 lines)

```kotlin
object Zap1MemoFormatter {
    private val PATTERN = Regex("^(ZAP1|NSM1):([0-9a-f]{2}):([0-9a-f]{64})$")

    private val EVENTS = mapOf(
        "01" to "PROGRAM_ENTRY", "02" to "OWNERSHIP_ATTEST",
        "03" to "CONTRACT_ANCHOR", "04" to "DEPLOYMENT",
        "05" to "HOSTING_PAYMENT", "06" to "SHIELD_RENEWAL",
        "07" to "TRANSFER", "08" to "EXIT",
        "09" to "MERKLE_ROOT", "0a" to "STAKING_DEPOSIT",
        "0b" to "STAKING_WITHDRAW", "0c" to "STAKING_REWARD",
        "0d" to "GOVERNANCE_PROPOSAL", "0e" to "GOVERNANCE_VOTE",
        "0f" to "GOVERNANCE_RESULT"
    )

    fun parse(memo: String): Attestation? {
        val m = PATTERN.matchEntire(memo.trim()) ?: return null
        return Attestation(
            prefix = m.groupValues[1],
            typeHex = m.groupValues[2],
            event = EVENTS[m.groupValues[2]] ?: "TYPE_0x${m.groupValues[2]}",
            hash = m.groupValues[3]
        )
    }

    data class Attestation(
        val prefix: String,
        val typeHex: String,
        val event: String,
        val hash: String
    ) {
        val verifyUrl get() = "https://pay.frontiercompute.io/verify/$hash"
        val shortHash get() = hash.take(12) + "..."
        val isLegacy get() = prefix == "NSM1"
    }
}
```

## No new dependencies

The parser is pure Kotlin string matching. No crate bindings, no FFI, no network calls. For projects that want deeper integration, the [zcash-memo-decode](https://crates.io/crates/zcash-memo-decode) Rust crate (0 dependencies) classifies all Zcash memo formats including ZAP1, ZIP 302, text, binary, and empty.

## Context

- Protocol: [github.com/Frontier-Compute/zap1](https://github.com/Frontier-Compute/zap1)
- Verification SDK: [crates.io/crates/zap1-verify](https://crates.io/crates/zap1-verify) (Rust + WASM)
- JS SDK: [npmjs.com/package/@frontiercompute/zap1](https://www.npmjs.com/package/@frontiercompute/zap1)
- Solidity verifier: [github.com/Frontier-Compute/zap1-verify-sol](https://github.com/Frontier-Compute/zap1-verify-sol) (deployed on Sepolia, Etherscan verified)
- ZIP draft: [zcash/zips PR #1243](https://github.com/zcash/zips/pull/1243)
- ZIP 302 partType request: [zcash/zips #1250](https://github.com/zcash/zips/issues/1250)
- Spec: [ONCHAIN_PROTOCOL.md](https://github.com/Frontier-Compute/zap1/blob/main/ONCHAIN_PROTOCOL.md) (v3.0.0)
- Full integration roadmap: [WALLET_INTEGRATION.md](https://github.com/Frontier-Compute/zap1/blob/main/WALLET_INTEGRATION.md)

Happy to open the PR if there's interest.

benjoshyo-create:
Closes #2172

Structured ZAP1 memos currently render as the raw `ZAP1:{TYPE}:{HASH}`
marker, which tells the reader nothing. This parses them and renders a
typed card showing the event type and an abbreviated commitment hash.

## What changed

- `ZapAttestationMemo` — a dependency-free parser for the
  `ZAP1:{EVENT_TYPE}:{LEAF_HASH}` grammar, also recognising legacy `NSM1`
  markers, with unit tests
- `TransactionDetailZapMemo` — the card, built from the existing design
  system tokens and mirroring `TransactionDetailInfoMemo`'s structure
- One branch added to `TransactionDetailMemo` to route parsed
  attestations to the card
- Three strings in `ui-design-lib`'s common resources

## Notes for review

**Ordinary memos are unaffected.** Parsing is conservative and never
throws — anything that isn't an exact grammar match returns null and
falls through to the existing plain-text path untouched. Tests cover
this explicitly, including text that merely mentions the prefix.

**Event labels are derived, not mapped.** ZAP1 defines 18 event types
and the registry is still moving (draft ZIP at `zcash/zips#1243`).
Hard-coding a label per type means anything added after this build
renders as raw `SCREAMING_SNAKE_CASE`; deriving it (`AGENT_ACTION` →
"Agent action") degrades gracefully. Trade-off is that type names aren't
localised — happy to switch to per-type strings if you'd prefer, once
the registry settles.

**No network, no verification claim.** The card shows only what the memo
already contains. Resolving the leaf hash against the ZAP1 API would
leak wallet interest in a specific transaction to a third party, so any
verification state seems like it warrants its own privacy review.

**Parsing sits in the UI layer**, wrapped in `remember`. Doing it in
`CommonTransactionDetailMapper` is arguably cleaner layering, but
`TransactionDetailMemoState.content` is a `StringResource` rather than a
raw string, so classification there would need the memo text threaded
through, plus changes to fixtures and the mapper. Glad to move it if you
prefer the domain-layer placement.

**Spanish strings not added** — Android falls back to English for missing
keys, and I'd rather leave that to your localisation process than
machine-translate into a shipped locale.

**No new drawables.** An event-type icon would suit the card but needs
design input; hierarchy is carried by weight and colour for now. Happy
to iterate on the visual treatment with your designer.

## Testing

Unit tests cover the parser: valid ZAP1 and NSM1, whitespace/padding
tolerance, label derivation, unknown-but-well-formed event types, and
rejection of plain text, malformed types, malformed hashes, wrong field
counts and unknown prefixes.

I don't have an Android SDK set up locally, so I'm relying on CI for the
compile and instrumentation checks — flagging that rather than claiming
a local build I didn't run.


Zk-nd3r (changes requested):
Thanks for picking up #2172. The parser needs one protocol correction: the wire type is a two-digit lowercase hex byte, not a name such as `AGENT_ACTION`, and the hash is exactly 64 lowercase hex characters. The current parser rejects canonical `ZAP1:09:<hash>` and `NSM1:04:<hash>` memos while accepting the invented named format.

I submitted a correction against your branch at https://github.com/benjoshyo-create/zodl-android/pull/1. It covers the 18 defined event types, explicit unknown-type labels and malformed-input fallback. It also labels the commitment as unverified and preserves full-memo copying. Three native Kotlin regression cases fail on this PR's current head; all 13 tests pass with the correction. Ktlint passes. Full Android validation remains pending: the Gradle test task stopped at a missing local Android SDK.

Keeping rendering local makes sense. The old Verify URL in my April issue currently serves the company homepage, including for the public root documented in the spec, so it should stay out of this rendering patch. Recognizing a memo does not verify its event claim or its anchor binding.

benjoshyo-create:
Thanks for the correction, and for catching this properly.

You're right that the parser had the wire format wrong — it was
built against a named type field rather than the two-digit hex
byte, so it would have rejected canonical ZAP1:09:<hash> and
NSM1:04:<hash> memos while accepting a format that doesn't exist
on chain.

Two things in your fix worth calling out beyond the grammar: the
payloadHash rename is right, since type 09 carries a root rather
than a leaf hash, and "Verification hash" / "ZAP1 attestation"
implied the wallet had verified something it hadn't. "Commitment
hash" and the explicit "not verified" labels are more honest.

Merged benjoshyo-create#1 into this branch.

Agreed on leaving the Verify URL out — recognising a memo
shouldn't imply anything about its event claim or anchor binding.

Still relying on CI for full Android validation; no local SDK
here either.

benjoshyo-create:
@Zk-nd3r resolved the CHANGELOG conflict with main — the branch
is clean now. Whenever you have a moment, could you re-review?
The workflows are also awaiting maintainer approval, so CI hasn't
run on this yet.

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 
 ### Added:
 
+ - feature/zap1-attestation-memo-cards
+- ZAP1 and legacy NSM1 transaction memos display event labels and commitment hashes. These cards
+  classify memo text without verifying the attestation (#2172).
+- Bitcoin, Litecoin, EIP-681 and Solana payment URIs are parsed through the SDK and routed into asset selection,
+  Pay and Swap; unsupported schemes, Solana transaction requests and unknown assets are rejected (MOB-1751).
```
Zk-nd3r: Non-blocking changelog cleanup: please move the ZAP1 note into the existing `Unreleased` section. It currently sits under the already-released 3.14.0 heading. The merge resolution also left the `feature/zap1-attestation-memo-cards` branch label and reintroduced an unrelated MOB-1751 payment-URI entry; neither belongs in this PR's changelog diff.

CHANGELOG.md
```diff
@@ -47,6 +53,7 @@ and this application adheres to [Semantic Versioning](https://semver.org/spec/v2
 - Retrying a Coinholder Polling submission that failed while sending the encrypted vote shares now
   resends the shares that never reached a helper server, instead of skipping every question whose
   vote was already on chain and failing the same way again.
+- main
```
Zk-nd3r: This is another branch-name leftover from the merge resolution; it can be removed.
```suggestion
```

Zk-nd3r (approved):
Thanks for incorporating the correction and resolving the merge conflict. Re-reviewed at `018eb3e2a33a3e41ab90ae42f76ff09457d3052b`.

The wire-format and verification-label blockers in my previous review are resolved. The corrected parser, memo-card code, copy-action wiring and parser tests are unchanged from the correction you merged. I reran all 13 native Kotlin/JUnit parser and regression tests on this head; they pass. Targeted ktlint, XML parsing and the PR diff whitespace check also pass.

Rendering remains local, unknown event codes stay explicit, malformed memos keep their existing fallback, and the card does not claim proof verification or introduce the old Verify URL.

Two small CHANGELOG cleanup notes are inline. Full Android/Compose validation is still outstanding: this head has no check runs, and the local Android SDK is absent. A ZODL maintainer still needs to authorize CI before the app-level checks can run.

benjoshyo-create:
Thanks — both were conflict-resolution artifacts. Cleaned up: the
ZAP1 note is now under Unreleased, the branch labels are gone, and
the MOB-1751 entry is back to matching main. The changelog diff is
just the two ZAP1 lines now.

CHANGELOG.md
Zk-nd3r: Thanks. Please restore [the changelog header](https://github.com/zodl-inc/zodl-android/blob/1690b28e0731a7beb08a633019c4c9a60d1cf61e/CHANGELOG.md#L1-L6) too. It got removed in the cleanup. Approval still stands.

benjoshyo-create:
Good catch — the header got dropped in the edit. Restored the title,
intro and format/versioning links, so the diff against main is now
just the new Unreleased entry.

PR #2527, opened 2026-09-17
fix(ci): retry CF cache purge up to 3 times on transient errors

y4ssi:
## Summary

- The v3.13 F-Droid release job failed at the Cloudflare cache purge step with exit code 254 — a transient CF API error. The actual F-Droid publish to GitLab Pages had already succeeded.
- Add a retry loop (3 attempts, 10s between each) before failing the job.
- On each attempt, log the full CF response so future failures are diagnosable.

## Test plan

- [ ] Merge and verify next release's `publish_to_fdroid` job completes without retries under normal conditions
- [ ] If CF is flaky, confirm the retry loop recovers and the job still passes


zcash-harry:
## Review

### 1. The cause is not Cloudflare. The cause is AWS. (Blocker)
- This step failed on the last 5 releases: 3.10.0, 3.10.1, 3.10.2, 3.11.0 and 3.13.0.
- This step has never passed. A temporary error does not fail 5 times in a row.
- Exit code 254 comes from the AWS CLI. It tells you that AWS sent back an error.
- The logs do not show the message "Cloudflare cache purge failed". Thus, the step stopped before it sent the request to Cloudflare.
- The step stops at the first command that fails, because the shell uses `bash -e`.
- The first commands in the step read 2 secrets from AWS Secrets Manager. One of these commands fails.
- **Possible causes:**
  - The release role (`AWS_RELEASE_ROLE_ARN`) cannot read the secrets.
  - A secret does not exist in `us-east-1`.
- **Result:** the retry loop does not run. The next release will fail again at the same step.
- **Limit:** GitHub does not give the full logs now (HTTP 410). I could not see the AWS error text. Look in AWS CloudTrail for the denied `GetSecretValue` calls to find it.

### 2. The retry does not catch most temporary errors
- The step uses `bash -e`. Thus, some errors stop the step on the first try:
  - **Network error:** `curl` exits with an error code. The step stops. I tested this: a DNS error gives exit code 6.
  - **HTML error page from Cloudflare (502/503):** `jq` cannot read the page. The step stops. I tested this: `jq` gives exit code 5.
- The loop tries again only when Cloudflare sends JSON with `success: false`. This error is usually not temporary. Examples are a bad token or a wrong zone.
- **Fix:**
  - Add `|| true` after each command in `$(...)`.
  - Or use `jq -r '.success // false' 2>/dev/null || echo false`.

### 3. Small points
- **Timeout:** `curl` has no time limit. Add `-m 30`. The GitLab step already uses `-m 30`.
- **Readability:** the 4 URLs are now on one line. This line is hard to read. Put them on separate lines again, for example with a heredoc.
- **Log output:** the log shows the full Cloudflare response. This is safe, because the token and zone ID are masked.
- **End of loop:** the line `[ "${attempt}" -lt 3 ] && sleep 10` is correct. I tested it. The loop ends correctly after try 3.

### 4. CI on this PR
- `test_android_modules_emulator (app)` failed.
- The e2e jobs had not finished.
- These jobs do not use `release.yaml`. Thus, this PR almost certainly did not cause these results.

## Recommendation
1. **Fix the AWS access first. This is the real fix.**
   - Give the release role `secretsmanager:GetSecretValue` for the 2 secrets.
   - Or create the missing secret.
2. **Keep the retry only after you fix it** (point 2). Without that fix, the retry does not help much.
3. **Optional:** add `continue-on-error: true` to the purge step. The F-Droid publish is already complete at this point. A failed cache purge must not fail the full release.

y4ssi:
Thanks Harry — you're right on all counts.

**Root cause fixed (IAM):** The `zodl-android-release-github-actions` role was missing `secretsmanager:GetSecretValue` for `/infra/cloudflare/api_token` and `/infra/cloudflare/foss-fdroid-zone-id`. That's why it failed on every release since 3.10.0 — the step never reached Cloudflare. Fixed in infra `f1d4614` and hotfixed live already.

**Workflow updated (latest commit `f69c18df`) addressing your points:**
- `continue-on-error: true` — publish is already complete at this step; a missed purge is non-blocking
- `curl -sf -m 30` — fail on HTTP error, 30s timeout
- curl failure falls back to `{"success":false,"error":"curl_failed"}` so `jq` never sees non-JSON
- `jq -r '.success // false' 2>/dev/null || echo "false"` — safe against parse errors
- `PURGE_SUCCESS="false"` initialised before the loop
- URLs back on separate lines for readability

.github/workflows/release.yaml
```diff
@@ -288,20 +289,26 @@ jobs:
...
-                "https://foss.zodl.com/fdroid/repo/app-zcashmainnet-foss-release-full-externallibs.apk",
-                "https://foss.zodl.com/fdroid/repo/app-zcashmainnet-foss-release.apk",
-                "https://foss.zodl.com/fdroid/repo/index-v2.json",
-                "https://foss.zodl.com/fdroid/repo/entry.jar"
-              ]
-            }' | jq -r '.success')
-          echo "CF purge result: ${PURGE_SUCCESS}"
+          PURGE_SUCCESS="false"
+          for attempt in 1 2 3; do
+            PURGE_RESULT=$(curl -sf -m 30 -X POST \
```
noop-sk: Adding `-f` here makes curl discard the response body on any HTTP error status (4xx/5xx). That defeats the goal of this PR (and Harry's own ask above — "the log shows the full Cloudflare response, this is safe") to log the full CF response for diagnosability, precisely for the error class (auth/permission failures, rate limiting, 5xx) most in need of diagnosis.

Verified empirically: `curl -sf` against a server returning HTTP 403 with a JSON body prints nothing and exits 22, while `curl -s` (no `-f`) prints the full body with exit 0. So on the exact class of error that caused the original incident (invalid/expired token), the fallback `{"success":false,"error":"curl_failed"}` is all that reaches the log — the real CF error code/message never does.

Suggest dropping `-f` and checking status separately, e.g. `curl -s -w '\n%{http_code}'` and branch on the trailing status code, or `curl --fail-with-body` (curl ≥7.76) which fails the same way but still emits the body.

.github/workflows/release.yaml
```diff
@@ -288,20 +289,26 @@ jobs:
...
-              "files": [
-                "https://foss.zodl.com/fdroid/repo/app-zcashmainnet-foss-release-full-externallibs.apk",
-                "https://foss.zodl.com/fdroid/repo/app-zcashmainnet-foss-release.apk",
-                "https://foss.zodl.com/fdroid/repo/index-v2.json",
-                "https://foss.zodl.com/fdroid/repo/entry.jar"
-              ]
-            }' | jq -r '.success')
-          echo "CF purge result: ${PURGE_SUCCESS}"
+          PURGE_SUCCESS="false"
+          for attempt in 1 2 3; do
```
noop-sk, line 293: Minor: the retry loop retries identical requests regardless of whether the failure is transient (5xx/timeout/network) or permanent (4xx auth/permission). With `continue-on-error: true` now in place this no longer fails the release, but a bad/expired token still burns ~90s (3 attempts × 10s sleep) on every release run for a failure that can never succeed on retry. Not blocking — just flagging in case it's worth short-circuiting on a 4xx response code.

noop-sk (commented):
Automated review follow-up (see thread above) — two remaining issues from the latest commit.

y4ssi:
Latest commit `e2ac6cd4` addresses noop-sk's points:

**`-f` removed:** using `curl -s -m 30` (no `-f`). The `|| echo '...'` fallback only fires on network-level failures (DNS, timeout) — HTTP 4xx/5xx return exit 0 with the full body intact, so jq and the log both see the real CF error. The `.success // false` default handles non-JSON bodies as a safety net.

**4xx short-circuit:** after each failed attempt, checks `errors[0].code` from the CF response. If it's a non-zero code < 500 (permanent auth/permission error), breaks the loop immediately instead of sleeping and retrying twice more. Transient 5xx/network failures still retry.

.github/workflows/release.yaml
```diff
@@ -288,20 +289,34 @@ jobs:
...
+                  "https://foss.zodl.com/fdroid/repo/index-v2.json",
+                  "https://foss.zodl.com/fdroid/repo/entry.jar"
+                ]
+              }' || echo '{"success":false,"error":"network_failure"}')
+            PURGE_SUCCESS=$(echo "${PURGE_RESULT}" | jq -r '.success // false' 2>/dev/null || echo "false")
+            echo "CF purge attempt ${attempt}: success=${PURGE_SUCCESS} response=${PURGE_RESULT}"
+            [ "${PURGE_SUCCESS}" = "true" ] && break
+            # Do not retry permanent 4xx failures (auth/permission — will never succeed).
+            HTTP_CODE=$(echo "${PURGE_RESULT}" | jq -r '.errors[0].code // 0' 2>/dev/null || echo "0")
+            if [ "${HTTP_CODE}" != "0" ] && [ "${HTTP_CODE}" -lt 500 ] 2>/dev/null; then
```
noop-sk: `-f` removal — confirmed fixed, response body is now preserved on HTTP errors. Thanks.

But this short-circuit doesn't work for the case it's meant to catch. `${HTTP_CODE}` here is actually Cloudflare's own `errors[0].code`, not an HTTP status — despite the variable name. Cloudflare's auth/permission error codes are in the thousands (e.g. `9109` "Invalid access token", `10000` "Authentication error", `6003` "Invalid API Token"), so `-lt 500` is false for essentially every real permanent error and the loop falls through to `sleep 10` anyway.

Verified empirically:
```
PURGE_RESULT='{"success":false,"errors":[{"code":9109,"message":"Invalid access token"}]}'
HTTP_CODE=$(echo "${PURGE_RESULT}" | jq -r '.errors[0].code // 0')  # -> 9109
[ "${HTTP_CODE}" -lt 500 ]  # -> false, short-circuit does NOT fire
```
So on the exact scenario that caused the original incident (invalid token), this still burns the full ~90s of retries — the commit message's "breaks the loop immediately" claim doesn't hold. If the goal is to skip retries on permanent errors, this needs to key off the actual HTTP status (e.g. `curl -s -w '\n%{http_code}' ...` and split it out) rather than CF's internal error-code namespace.

noop-sk (commented):
Re-checked the fix commit against the diff (not just the claim) — the `-f` removal is correctly fixed, but the 4xx short-circuit doesn't do what the commit message says.

.github/workflows/release.yaml
y4ssi: Confirmed, you're right — that variable was CF's `errors[0].code`, not an HTTP status, so `-lt 500` never fired on the real codes (9109/10000/6003). Fixed in `49e2b5e`:

- `curl -s -m 30 -w '\n%{http_code}'` appends the real status; split off with `tail -n1` / `sed '$d'`. Body still logged verbatim.
- `case` on the HTTP status: `4xx` (except `429`) breaks immediately; `429`, `5xx` and `000` (network fallback) retry.

Ran the loop under `bash -e -o pipefail` against a local mock:

| Scenario | Attempts | http | Notes |
|---|---|---|---|
| 403 + `{"code":9109}` | 1 | 403 | short-circuits, body logged |
| 429 | 3 | 429 | retries (transient 4xx) |
| 502 HTML body | 3 | 502 | retries, HTML logged, jq fallback → false |
| 404 empty body | 1 | 404 | short-circuits, empty response handled |
| connection refused | 3 | 000 | fallback JSON, retries |
| 200 success | 1 | 200 | success=true |

noop-sk (approved):
Reviewed the current head (49e2b5e6) — the HTTP-status short-circuit fix from the last round looks correct (verified against the diff, not just the commit message).

A few additional non-blocking findings from a deeper pass, safe to address later or as a quick follow-up:
- `success:false` with HTTP 200 isn't classified as permanent, so it burns 2 extra retries before giving up (still ends in a warning either way — cosmetic).
- `continue-on-error: true` is step-scoped, so it also swallows failures from the `aws secretsmanager get-secret-value` calls above the retry loop, not just the CF purge itself — worth narrowing the scope at some point so an IAM/secret misconfig doesn't look like a benign transient CF hiccup.
- Related: since purge failure no longer `exit 1`s, `continue-on-error` mostly only matters for that secrets-retrieval path now — slightly misleading for a future reader.

None of these block the release path (this is the last, already non-blocking step of the last job). Approving.

PR #2529, opened 2026-09-17
Merge release/v3.14.0 into main

noop-sk:
Back-merges the shipped app 3.14.0 release (candidate/v3.14.0 via #2525: version bump to 3.14.0, CHANGELOG/WHATS_NEW/fastlane, MOB-1751 revert) into `main`, per the cascade convention. `3.14.0` starts a new minor line with no `maint/v3.14.x` yet, so this merges directly into `main` (own maint line would be `maint/v3.14.x`, which does not exist).

Draft on purpose — this repo's auto-merge.yml arms squash auto-merge on every non-draft PR, which would break cascade ancestry. Mark ready and merge (--merge, not squash) immediately once approved.

2 commits, clean merge (verified via `git merge-tree`, no conflicts).


nesence-m (changes requested):
Release metadata looks right: `ZCASH_VERSION_NAME=3.14.0`, `ZCASH_SDK_VERSION=3.3.0-SNAPSHOT`, `SDK_COMMIT_PIN` cleared (and `build` passes against the published snapshot), `## [3.14.0 (2738)] - 2026-09-16` with an empty `[Unreleased]` above it, What's-new EN/ES and fastlane `2738.txt` consistent. Compare is ahead 2 / behind 0, so nothing on main risks being mis-parented.

The three red checks are not caused by this branch: `static_analysis_detekt` died in the Checkout step (`timed out after 1 minutes`, no detekt finding — it passed on the identical tree in #2525's run), and `Keystone` (`home.syncComplete` never visible) plus `e2e-android (phase 2)` (`crosspay-transaction`, `swap-transaction`, `swap-receive-usdc-to-zec` on `"Confirm" is visible`) failed the same way on #2522/#2517. None is a required check.

**One thing to settle before this lands:** the "Release preparation" commit (deaae5cd) also carries the full MOB-1751 revert — `CrossPayRequest.kt` and 14 related files, plus the `[Unreleased]` CHANGELOG entry is deleted rather than re-parented — while its message lists only the five changelog files. Because `candidate/v3.14.0` was cut from main, this merge applies that revert to main. #2522 (bad6942c) restored MOB-1751 on main one day ago with "Main keeps the feature for a later release". Is dropping it from main now intended?

- If yes: a line in the PR body saying so is enough and I'll approve.
- If main should keep it: this needs the #2522 shape instead — a `merge/release-v3.14.0-into-main` branch with the merge plus a "Restore MOB-1751 on main" commit (the revert is folded into the prep commit, so it has to be a path restore of those 15 files from 9b1914fe plus the CHANGELOG entry, not a `git revert`). SDK-wise nothing forces the revert: `PaymentUriParser` has shipped since SDK 3.2.0 and is in 3.3.0.

Landing note: `auto-merge.yml` is gone from main since 8b0ef0363, so nothing can squash this anymore; but the ruleset still allows "Rebase and merge", which would drop the merge commit 1b81f19e and break the cascade ancestry the same way. Please land with `gh pr merge --merge` only, ideally after tagging `3.14.0-2738` on 1b81f19e.

nesence-m (approved):

Issue #15, opened 2021-10-05
Add unstyled onboarding screen


ccjernigan:
The unstyled onboarding screen has multiple instructional pages, each describing something about the wallet app or zcash.

The designs currently have an initial screen with "skip" button in the corner, along with a next button to proceed.  The following screens also have a back button, to return to the prior step.  The final screen presents buttons for creating a wallet or importing an existing wallet.

We should implement this to be flexible, in terms of adding or removing pages during the middle.

PR #2530, opened 2026-09-17
Add cascade-merge automation for release -> maint -> main

noop-sk:
## Summary
- Adds `.github/workflows/cascade-merge.yml` + `.github/scripts/cascade-merge.sh`, automating the forward-merge cascade documented in the `zodl_an_cascade_merge` skill: shipped `release/vX.Y.Z` → its own `maint/vX.Y.x` → next newer maint line → … → `main`.
- The cascade is manual today — nothing triggers it on a maint landing, so a skipped link silently regresses a fix for anyone upgrading across the gap until a user reports the old bug on a newer version (this is a known, previously flagged gap).
- For every pending link the workflow opens (or reuses) a PR and enables GitHub's **native auto-merge**, `--merge` method only (never squash/rebase — a squashed cascade link poisons every cascade after it, per skill §3–4; this has already happened twice by hand).
- Runs on push to `maint/**`/`release/**` (fast path) and on a 30-min schedule (a merge performed by this workflow's own `GITHUB_TOKEN` doesn't itself trigger a new push-based run, so the schedule is what carries the cascade on to its next link once a human approves one).

## Why it can't be fully unattended
The `zodl_an_cascade_merge` skill (last verified 2026-08-14) describes `maint/*` as accepting direct pushes with no PR required. That's no longer true: the **"Protect main and maint branches" ruleset was added 2026-08-20** and now covers `main` *and* `maint/*`, requiring 1 approving review with `bypass_actors: []` / `current_user_can_bypass: "never"`. So this workflow gets as close to automatic as GitHub's security model allows — it does everything except click Approve. (Side benefit of that same ruleset: `allowed_merge_methods` on main/maint is `["merge", "rebase"]` — squash is disabled at the GitHub level there, so the app's old auto-squash-on-every-PR hazard the skill warns about doesn't apply on this repo; it has no `auto-merge.yml` at all.)

## Test plan
- [x] `bash -n` syntax check on `cascade-merge.sh`
- [x] Dry-ran the branch-discovery + ancestor-check logic (read-only, no `gh` mutation calls) against the real repo state — correctly found all 5 `maint/vX.Y.x` lines in order, correctly identified all 10 shipped `release/vX.Y.Z` branches as already back-merged, and correctly reported every maint→maint and maint→main link as `in-sync` (matches actual current state — nothing pending).
- [ ] First live run on `workflow_dispatch` (safe: on today's clean state it should open zero PRs) before relying on the schedule/push triggers.
- [ ] A real pending link (e.g. next time a bugfix lands on a maint line) to confirm PR creation + auto-merge-on-approval end to end.


nesence-m (changes requested):
Thanks for automating this — the chain, the numeric maint ordering (3.9 vs 3.10 verified against the live branch list) and the `merge-base --is-ancestor` predicate all match the documented cascade model, and the workflow hygiene (least-privilege job permissions, sha-pinned checkout, concurrency group, no `${{ }}` in `run:`) is right. The ruleset claims in the description check out (ruleset 21090408: main + maint/*, 1 approval, no bypass, merge+rebase; no `auto-merge.yml` on main). One hard blocker, then a few design points.

**Blocking**

- `.github/scripts/cascade-merge.sh:86` — `gh pr create` has no `--json` flag (checked gh 2.94.0 locally, gh 2.100.0 on ubuntu-latest, and `pkg/cmd/pr/create/create.go` at v2.100.0). With `2>/dev/null … || { log "[warn] …"; return 0; }` the `unknown flag: --json` error is swallowed, so the script never opens a PR and the run stays green — every pending link just logs "[warn] could not open PR … skipping". Use the URL gh prints on stdout (`pr_url=$(gh pr create …); pr_number="${pr_url##*/}"`) or re-query with `gh pr list --head/--base`, and drop the `2>/dev/null`. Note that all links are in sync right now, so the planned first `workflow_dispatch` would not have surfaced this.

**Non-blocking, worth deciding before relying on it**

- `.github/workflows/cascade-merge.yml:56` — PRs created with `GITHUB_TOKEN` do not trigger `pull_request` workflows, so `pull-request.yml`/`e2e-smoke.yml` never run on cascade PRs; the ruleset has no required checks, so auto-merge lands e.g. maint→main on the first approval with zero CI on the merge result. Either mint an App token / PAT for the create step, or state in the generated PR body that CI has not run.
- `cascade-merge.sh:100-108` — auto-merge is re-armed every 30 min; a human cannot park a link with `--disable-auto`. A `cascade-hold` label check (on reuse and on create) would cover that and the next point.
- `cascade-merge.sh:112-137` — release→maint back-merges need the CHANGELOG re-parenting fix *inside* the merge commit (skill §5; missed in #2499 and SDK #15). Native auto-merge produces a plain merge commit, so the approver can only eyeball CHANGELOG.md, and closing the automated PR to redo it on a topic branch makes line 62 open a new one every 30 min (GitHub only blocks duplicates of open PRs). Suggest not arming auto-merge for the release→maint step, or guarding on the CHANGELOG diff.
- `cascade-merge.sh:93-98` — `mergeable` is `UNKNOWN` right after creation; the script logs it as `[conflict]` and waits for the next tick. Treat `UNKNOWN` as pending (short retry or separate log line).
- `cascade-merge.sh:126-128` — a tagged release without a maint line (release/v3.14.0 once tagged) is skipped with no log line; conflicts/arming failures only reach the run log and the bot-authored PR has no reviewer, so nobody is notified. A `--reviewer` and a `[unmatched-release]` log line would help.
- Repo config, not this diff: ruleset 21090408 still allows `rebase`, which breaks ancestry like squash; consider merge-only now that automation depends on it.

**Nits**

- `cascade-merge.sh:72-85` — the heredoc inside `$( )` with the apostrophe on line 80 does not parse on bash 3.2 (macOS stock bash; `mapfile` needs bash ≥4 anyway). Fine on the runner; either document Homebrew bash for local dry-runs or use `--body-file`.
- `cascade-merge.yml:24-27` — add a `tags:` trigger (`[0-9]+.[0-9]+.[0-9]+-[0-9]+`, as in create-release.yml): the candidate→release push arrives before the tag, so the release→maint link currently waits for the next cron tick.
- `cascade-merge.sh:108-109` — on a draft PR the warning blames "missing required checks"; check `isDraft` and say so.

Sequencing once fixed: merge → `gh workflow run cascade-merge.yml` (expect five `[skip]` lines) → the create path first runs at the next maint landing; whoever approves that first automated PR should look at CHANGELOG.md by hand.

noop-sk:
Thanks for the thorough review — went through each point:

**Blocking — fixed.** Confirmed independently (`gh pr create --help` on 2.88.1 locally has no `--json`/`-q`). Now parses the PR number from the URL `gh pr create` prints on stdout, and dropped the `2>/dev/null` so a real failure surfaces in the run log instead of being swallowed.

**Fixed / addressed this round:**
- `cascade-hold` label check before (re-)arming auto-merge.
- `mergeable == UNKNOWN` is now treated as pending-retry (logged, not enabled, not called a conflict) rather than a conflict.
- Unmatched-release and draft-PR cases now get their own log line.
- Added the `tags:` trigger (same pattern as `create-release.yml`) so a release→maint check doesn't sit until the next cron tick.
- Release→maint links are **never** auto-merged now, full stop — went with your first option rather than trying to detect the CHANGELOG re-parenting generically. The generated PR body explains why and points at #2499 as the worked example.
- `--reviewer` is now wired up via an (optional) `CASCADE_REVIEWER` env var, unset by default — didn't want to hardcode a specific person into the script itself.

**Left open, not decided here:**
- CI not running on bot-opened PRs (no required checks on the ruleset either, so auto-merge currently lands maint→main on approval alone, with zero CI on the merge result itself). The generated PR body now says this explicitly so the approver knows to check. Minting an App token/PAT for the create step is the real fix — want to loop in whoever owns repo secrets before doing that as a separate follow-up rather than bundling it into this PR.
- Ruleset still allows `rebase` on main/maint — agreed that's the same ancestry hazard as squash. Flagging to whoever owns ruleset 21090408 as a follow-up outside this diff.

Re-ran the script for real against the live repo state (read-only path only, since everything's still in-sync — no `gh` mutation calls were reached) to confirm the refactor didn't change behavior: same 5×`[skip]` plus the new `[unmatched-release]` line for `release/v3.14.0` (no `maint/v3.14.x` yet, which is correct).

nesence-m (approved):
Re-checked at 724ec609 — the blocker is fixed (no `--json` on `gh pr create`, stderr no longer swallowed, `${pr_url##*/}` on gh's URL line, a failure logs and skips that link without aborting the sweep), release→maint is structurally kept away from `--auto` (the `[manual]` return at `cascade-merge.sh:137-140` sits before the single `gh pr merge --auto` at `:173`), `UNKNOWN` is a retry, and the label / draft / unmatched-release paths are correct. shellcheck clean. Three non-blocking notes, all one-liners:

1. `cascade-merge.sh:142-147` — the `[hold]` branch returns before the `autoMergeRequest` check and never calls `gh pr merge --disable-auto`, so a label added *after* a tick already armed the PR does not pause anything; the next approval still merges. Either disarm there when auto-merge is already on, or reword the body line "Add the `cascade-hold` label to pause that" to say it only stops the workflow from (re-)arming. Also: the `cascade-hold` label does not exist in the repo yet.
2. `cascade-merge.yml:62` — `CASCADE_REVIEWER` is read by the script but never set by the workflow; `CASCADE_REVIEWER: ${{ vars.CASCADE_REVIEWER }}` would make it a repo variable (empty = today's behaviour). Note a team slug will fail `gh pr create` under `GITHUB_TOKEN`; a user handle is fine.
3. `cascade-merge.sh:131` — a `[ -n "$pr_number" ] || return 0` guard after the URL parse would stop an empty result from becoming `gh pr view ""` and aborting the run under `set -e`.

FYI: `bash -n` under macOS bash 3.2 fails on the apostrophe in `workflow's` inside the `$( cat <<'BODY' … )` (the header's "needs bash >= 4" note covers it in practice); the tag trigger, like create-release.yml, only fires once the workflow file exists on the tagged release line — the cron covers the gap. With 10 merged release links and 5 maint links the first run should print 15 `[skip]` lines plus `[unmatched-release]` for release/v3.14.0, not 5. The red `test_android_modules_emulator (app)` is the emulator host dying (`Expected 12 tests, received 4`, port 5554 refused), pre-existing.

Issue #26, opened 2021-10-12
Android Gradle Plugin 7.0.3


ccjernigan:
A minor bugfix update to the Android Gradle Plugin is available.  To implement this, update the version of the Android Gradle Plugin to 7.0.3 in the root `gradle.properties`.

PR #2531, opened 2026-09-18
MOB-1616: Swap the anchor-error string match for the typed SDK exception

noop-sk:
Follow-up to zodl-inc/zodl-android-wallet-sdk#26 (https://github.com/zodl-inc/zodl-android-wallet-sdk/pull/26), per nesence-m's review note there.

`TransactionProgressVM.isAnchorError()` detected the anchor-not-found failure by matching `"Unable to compute anchor"` on the `rootCause` message of the generic `TransactionEncoderException.TransactionNotCreatedException` — the only signal the SDK gave before (confirmed this is exactly `ProposalError::AnchorNotFound`'s `Display` text in `zcash_client_backend`). The companion SDK PR adds `TransactionEncoderException.AnchorNotFoundException` specifically so callers don't have to string-match. This swaps the check to the typed exception, walking the same cause chain (the new exception is thrown directly from `createProposedTransactions`, and nested as the `cause` of `PcztException.CreatePcztFromProposalException` from `createPcztFromProposal`).

**Draft, blocked on the SDK side**: `SDK_COMMIT_PIN` is set to the SDK PR's current head commit so CI builds against the real typed exception — this is **not mergeable as-is**. Once `zodl-inc/zodl-android-wallet-sdk#26` merges into `main` and a new snapshot publishes, repoint `SDK_COMMIT_PIN` to that merged sha, verify, then clear it before this leaves draft, per the app's own `SDK_COMMIT_PIN` convention. Tracked under MOB-1616 (https://linear.app/zodl/issue/MOB-1616/anchorintervalmismatch-anchornotfound-have-no-typed-kotlin-surface).

Verification: compiled `:ui-lib:compileZcashmainnetStoreDebugKotlin` against the SDK PR branch via `SDK_INCLUDED_BUILD_PATH` (local only, not committed) to confirm the typed exception resolves and the swap type-checks. `ktlint` / `detektAll` pass.


nesence-m (approved):
Verified the swap end to end against the pinned SDK (`cb964d61e4f19e1e83c257ce6577f2b525f653ae` = the #26 merge commit, current fork `main` head): for a regular send the `AnchorNotFoundException` arrives unwrapped as `SubmitResult.Error.cause` (`ZashiProposalRepository.submit` → `SdkBroadcaster` → `PendingSubmitPlanStore` → `OutboundTransactionManagerImpl` → `TransactionEncoderImpl`, none of which re-wrap), so the walk finds it at depth 0, and dropping the string match is correct because the pinned encoder never produces `TransactionNotCreatedException` for an anchor failure any more. No other call site in the app matched the string. The pin line is the only `gradle.properties` change (full 40-char sha, `check_properties` green); CHANGELOG correctly untouched. Keeping the pin set is right: `cb964d61e` is not reachable from `release/v3.3.0`, so `3.3.0-SNAPSHOT` does not contain the typed exception.

Non-blocking notes:

1. `TransactionProgressVM.kt:465-468` — the comment's PCZT justification does not hold in the app: `KeystoneProposalRepository.createPCZTFromProposal()` runs at proposal-creation time (`CreateProposalUseCase`, `ShieldFundsUseCase`, `RequestSwapQuoteUseCase`, `MigrationReviewVM`, …), never through `KeystoneProposalRepository.submit()`, so a `CreatePcztFromProposalException(cause = AnchorNotFoundException)` never becomes a `SubmitResult.Error` and Keystone users still get the generic error, exactly as before (pre-existing reach, same as the old string match). Worth rewording the comment to "defensive walk", and, separately, a follow-up to type-check `e.cause` in those use cases if we want the anchor copy on Keystone too.
2. `isAnchorError` is private and untested (it never was). Making it `internal` like `pendingDescription()` and adding a direct / wrapped / `TransactionNotCreatedException`-with-old-message / `Failure` test would pin the contract on the app side as the SDK does with `TransactionEncoderImplAnchorNotFoundTest`. (`CreatePcztFromProposalException` has an internal constructor, so a plain wrapper exception exercises the same walk.)
3. Nit: the walk at `TransactionProgressVM.kt:471-476` has no cycle guard (pre-existing since MOB-385; `generateSequence(cause) { it.cause }.take(n).any { it is AnchorNotFoundException }` would be both shorter and bounded), and the new comment is a `//` block where the repo wants KDoc.
4. The PR description still says "Draft, blocked / not mergeable as-is / clear before leaving draft"; please update it to say the pin stays until a published SDK build contains the typed exception and gets cleared at release-cut time.

CI: `test_kotlin_modules` is red only because Maven Central answered 429 on the `kotlinx-coroutines-test` pom (the build died in 38 s before any test ran) — a rerun should clear it. e2e / Keystone were still pending at review time.

Otr437 (approved):

